PR #47, opened 2024-03-30
Fix typo in the configuration docs

cedmax:
As per the title


MohamedBassem:
Good catch, thanks for taking the time and sending the PR!

PR #59, opened 2024-03-31
Bump chrome container version to 102 to enable linux/arm64 support on it

lbrame:
Necessary context: https://github.com/MohamedBassem/hoarder-app/issues/42

Changes the chrome container version in docker-compose from 100 to 102, because version 102 has a `linux/arm64` artifact and it can thus run on ARM architecture.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -24,7 +24,7 @@ services:
     volumes:
       - redis:/data
   chrome:
-    image: gcr.io/zenika-hub/alpine-chrome:100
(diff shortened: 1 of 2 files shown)


MohamedBassem:
Thanks a lot for opening a PR to fix it! Actually, let's bump it all the way to `123`. I tested it locally and seems to be working fine. We probably shouldn't be running a one year-old chrome version :)

lbrame:
Great idea! It's probably safer this way. I just pushed the change to the branch, should appear any second

MohamedBassem:
Thanks a lot!

lbrame:
You're welcome :)

PR #88, opened 2024-04-08
feature: Add PDF support

AhmadMuj:
- Allowed PDF upload in both mobile and web
- Dynamic type is being sent to the server based on the mimetype from both the mobile and web
- Supported PDF view as iframe in both mobile and web
- Added pdf-parse library to parse PDF files
- Added content, info and metadata for the bookmarkAssets table to store the pdf parsing results so they can be indexed in the future. This would also be used in getting a description about the image and storing it in the content field.
- Added inferTagsFromPDF that would parse the PDF, update the bookmarkAssets row and then inferTags from the text that includes the PDF content.


apps/web/app/api/assets/route.ts
```diff
@@ -26,7 +27,7 @@ export async function POST(request: Request) {
     });
   }
   const formData = await request.formData();
-  const data = formData.get("image");
+  const data = formData.get("file");
```
MohamedBassem: For backward compatibility reasons, especially with phones which don't get release immediately, we'll need to keep accepting the `image` field for a while.
```suggestion
  // TODO: Deprecate the `image` field in a couple of releases
  const data = formData.get("file") ?? formData.get("image");
```

apps/workers/openaiWorker.ts
```diff
@@ -88,6 +94,14 @@ You must respond in JSON with the key "tags" and the value is list of tags.
 CONTENT START HERE:
 `;
 
+const PDF_PROMPT_BASE = `
```
MohamedBassem: The difference between this and the text prompt is not that big. I think we can just continue using the text prompt even for pdfs, what do you think?

apps/workers/openaiWorker.ts
```diff
@@ -166,6 +180,47 @@ async function inferTagsFromImage(
...
+  }
+
+  await db
+    .update(bookmarkAssets)
+    .set({
+      content: pdfParse.text,
+      info: pdfParse.info ? JSON.stringify(pdfParse.info) : null,
+      metadata: pdfParse.metadata ? JSON.stringify(pdfParse.metadata) : null,
+    })
+    .where(eq(bookmarkAssets.id, bookmark.id));
```
MohamedBassem: I think this logic fits more into being in the `crawler`. It's the worker type that enriches the `bookmark` in prep for the inference. We should probably rename it as well now that it's no longer about "crawling" :D 

I don't mind keeping this here for the sake of this PR and I can move it after it lands :)

apps/workers/openaiWorker.ts
```diff
@@ -182,11 +237,19 @@ async function inferTags(
   if (bookmark.link || bookmark.text) {
     response = await inferTagsFromText(bookmark, inferenceClient);
   } else if (bookmark.asset) {
-    response = await inferTagsFromImage(jobId, bookmark, inferenceClient);
+    if (bookmark.asset.assetType === "image") {
```
MohamedBassem: nit: This is an enum, let's do a switch maybe?

apps/workers/package.json
```diff
@@ -26,6 +26,7 @@
     "metascraper-url": "^5.43.4",
     "ollama": "^0.5.0",
     "openai": "^4.29.0",
+    "pdf-parse": "^1.1.1",
```
MohamedBassem: I'm not a big fan of this package. It seems to be abandoned with the last version being from 5 years ago (https://www.npmjs.com/package/pdf-parse?activeTab=versions). 
Should we maybe use https://github.com/mozilla/pdf.js instead? It even comes with a viewer that we can probably use instead of the `iframe`?

apps/workers/searchWorker.ts
```diff
@@ -72,6 +73,13 @@ async function runIndex(
             content: bookmark.link.content,
           }
         : undefined),
+      ...(bookmark.asset
+        ? {
+            content: bookmark.asset.content,
+            info: bookmark.asset.info,
+            metadata: bookmark.asset.metadata,
+          }
+        : undefined),
```
MohamedBassem: I'm a bit surprised that typescript didn't complain about the fields used here as they are not defined in the search schema in `packages/shared/search.ts`

packages/db/schema.ts
```diff
@@ -157,8 +157,11 @@ export const bookmarkAssets = sqliteTable("bookmarkAssets", {
     .primaryKey()
     .$defaultFn(() => createId())
     .references(() => bookmarks.id, { onDelete: "cascade" }),
-  assetType: text("assetType", { enum: ["image"] }).notNull(),
+  assetType: text("assetType", { enum: ["image", "pdf"] }).notNull(),
   assetId: text("assetId").notNull(),
+  content: text("content"),
+  metadata: text("metadata"),
+  info: text("info"),
```
MohamedBassem: What exactly gets added in the `info` and `metadata`? Are they really worth storing and indexing?

apps/workers/openaiWorker.ts
```diff
@@ -166,6 +180,47 @@ async function inferTagsFromImage(
...
+    .where(eq(bookmarkAssets.id, bookmark.id));
+
+  let prompt = PDF_PROMPT_BASE;
+  if (pdfParse.info) {
+    prompt += "Info: " + JSON.stringify(pdfParse.info) + "\n";
+  }
+  if (pdfParse.metadata) {
+    prompt += "Metadata: " + JSON.stringify(pdfParse.metadata) + "\n";
+  }
+  prompt += "Content: \n" + pdfParse.text;
```
MohamedBassem: Models have a context window that we need to respect. You'll need to truncate the content to fit in that window especially if someone uploads a large pdf.

There's a function called `truncateContent` that's used in the text prompt builder, maybe you can reuse it?

packages/db/schema.ts
```diff
@@ -157,8 +157,11 @@ export const bookmarkAssets = sqliteTable("bookmarkAssets", {
     .primaryKey()
     .$defaultFn(() => createId())
     .references(() => bookmarks.id, { onDelete: "cascade" }),
-  assetType: text("assetType", { enum: ["image"] }).notNull(),
+  assetType: text("assetType", { enum: ["image", "pdf"] }).notNull(),
   assetId: text("assetId").notNull(),
+  content: text("content"),
```
MohamedBassem, line 162: let's maybe also store the file name?

MohamedBassem (changes requested):
Oh wow. I didn't see this coming!
First of all, thanks a lot for taking the time and contributing with this PR. I'm impressed with how you managed to navigate the different parts of the codebase without any help! As the first contributor to this project, please let me know if there's anything I can improve in the development documentation :)

The PR itself looks good to me, I requested some minor changes. Don't feel obliged to tackle the bigger ones, I'm happy to take over what you won't have time to fix!

I think the main missing thing is mobile rendering for the pdf type asset which lives in `apps/mobile/components/bookmarks/BookmarkCard.tsx`. I think for now, we can start with a static image saying that this is of type pdf and add a download button, later we can add the preview there.

apps/web/app/api/assets/route.ts
ahmadmucom: You're correct. I'll add this

apps/workers/openaiWorker.ts
ahmadmucom: You are correct, I'm a big fan of generalization when possible. 
I'll combine them.

apps/workers/openaiWorker.ts
ahmadmucom: I would prefer to move it in the next PR to make this one stable.

packages/db/schema.ts
ahmadmucom: In case of a PDF, the metadata will have any PDF meta that are added to the PDF file like the author and so on. 
For the info, it's any relevant information about the PDF version. 

Honestly I kept them for the sake of applying the same for the images as well.
Basically, we are sending the image to openai, why not ask openai to provide both the tags and a description about the image ?
The cost is the same and it will enhance the images search by a huge amount with no extra cost.

apps/workers/openaiWorker.ts
ahmadmucom: I had in mind the 4MB limit for the upload. But you are correct I'll add the truncate content.

packages/db/schema.ts
ahmadmucom: Well, it's both a negative and a positive thing. 
In case of a "paste" image case, there would be no file name. 
Also many people will upload files with random names. 
I have mixed feelings about this.

packages/db/schema.ts
MohamedBassem: My main concern about them is the fact that they are not structured data and I'm not a big fan of dumping schema-less jsons into the database. Basically, they'll never be useful beyond getting indexed for search. But if you think they provide useful data, then why not. I think we can keep `metadata` and drop `info`? Or merge info into `metadata`?

> why not ask openai to provide both the tags and a description about the image ?

That I 100% agree with and this can be in the `content` field in the database. My comment is in particular about `info` and `metadata`.

packages/db/schema.ts
MohamedBassem: It's fine for the `paste`s to not have a filename. My plan is to use the filename of assets to be their `title` in the bookmark grid and also give the users the ability to modify it if they want. I checked and this is the default behavior in raindrop for example. Files without a `filename` would just show no title (like they are now). I think the filename in the pdf case would be a nice way to disambiguate the card without having to look into the content itself.

packages/db/schema.ts
ahmadmucom: I agree.
I'll dump the info and keep the metadata.

apps/workers/package.json
ahmadmucom: I think mozilla/pdf.js is not the right tool for us as it's a PDF viewer for HTML 5.
What do you think of this ?
https://github.com/modesty/pdf2json

MohamedBassem: I don't mind pdf2json. It has a weird api but sounds like it can work 👍

ahmadmucom: It has a weird API indeed. I'll work with it for now until we find another good fit.

AhmadMuj:
@MohamedBassem Kindly check now and let me know. 
I'll add the images paste feature later after we close this. 
I changed the prompt slightly to add a description for all content types. It's nice to have a summarization of the content. I didn't support indexing nor handling the content for now as I only worked on the prompt side for now. 
The rest is handled

apps/workers/openaiWorker.ts
```diff
@@ -71,34 +77,36 @@ export class OpenAiWorker {
...
+    pdf: "PDF file",
+    image: "Image",
+  };
+  return `I'm building a read-it-later app and I need your help with automatic tagging.
+${
+  type === "web" || type === "pdf" || type === "text"
+    ? `You are currently analyzing the content of a ${typeContent[type]}, please analyze the content after the sentence "CONTENT START HERE:"`
+    : `Please analyze the attached image`
+}
+Suggest relevant tags that describe its key themes, topics, and main ideas.
```
MohamedBassem: Do you mind if we move the prompt changes to another PR? I'm trying to be very careful with prompt changes as they can have unintended behaviors and they require some testing. For example, here we changed everything to `5-8 tags` and some people were complaining that the `3-5` that I had before were already too much.

I think for the sake of this PR let's just use the TEXT prompt without any changes. It'll probably be good enough :)

apps/workers/searchWorker.ts
```diff
@@ -72,6 +73,13 @@ async function runIndex(
             content: bookmark.link.content,
           }
         : undefined),
+      ...(bookmark.asset
+        ? {
+            content: bookmark.asset.content,
+            info: bookmark.asset.info,
```
MohamedBassem: There shouldn't be an `info` here? not sure how TS didn't complain as well

apps/workers/utils.ts
```diff
@@ -14,3 +16,35 @@ export function withTimeout<T, Ret>(
...
+  metadata: Record<string, string | Record<string, string>>;
+}> {
+  return new Promise((resolve, reject) => {
+    // Need raw text flag represents as number (1), reference : https://github.com/modesty/pdf2json/issues/76#issuecomment-236569265
+    const pdfParser = new PDFParser(null, 1);
+    pdfParser.on("pdfParser_dataError", reject);
+    pdfParser.on("pdfParser_dataReady", (pdfData) => {
+      console.log(pdfParser);
+      // eslint-disable-next-line
+      console.log((pdfParser as any).getRawTextContent());
```
MohamedBassem: We probably should remove those `console.log`s

apps/workers/utils.ts
```diff
@@ -14,3 +16,35 @@ export function withTimeout<T, Ret>(
     ]);
   };
 }
+
+export async function readPDFText(buffer: Buffer): Promise<{
+  text: string;
+  metadata: Record<string, string | Record<string, string>>;
```
MohamedBassem: It seems that `Meta` is `Record<string, any>` according to https://github.com/modesty/pdf2json/blob/master/pdfparser.d.ts#L26 ?

MohamedBassem (changes requested):
Thanks! I think this is good to go after addressing the prompt change comment! I'll take care of adding the mobile renderer in a separate PR.

It also seems that CI is failing, you'll need to run in the root of the repo:

```
pnpm typecheck && pnpm format:fix && pnpm lint:fix
```

to ensure that CI will pass when you push a new version.

packages/shared/assetdb.ts
```diff
@@ -12,6 +12,7 @@ function getAssetDir(userId: string, assetId: string) {
 
 export const zAssetMetadataSchema = z.object({
   contentType: z.string(),
+  fileName: z.string(),
```
MohamedBassem: This needs to be nullable for backward compatibility. Otherwise hoarder throws errors when we attempt to read the metadata for assets uploaded before this commit.

```suggestion
  fileName: z.string().nullish(),
```

MohamedBassem (commented):
The PR LGTM! We can land after fixing the asset schema and after you restore the preview fix that was lost in the merge commit!

PR #89, opened 2024-04-08
feature: Support pasting an image in the text field

AhmadMuj:
- Added onPaste listener on the textfield to support pasting images from the clipboard so they can be uploaded directly.


apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -33,6 +35,10 @@ function useFocusOnKeyPress(inputRef: React.RefObject<HTMLTextAreaElement>) {
 export default function EditorCard({ className }: { className?: string }) {
   const inputRef = useRef<HTMLTextAreaElement>(null);
 
+  const uploadAsset = useUploadAsset({
+    onComplete: () => console.log("file Uploaded"),
+  });
```
MohamedBassem, line 40: We probably should just remove the `onComplete` callback here (make it optional in `useUploadAsset`).

apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -80,6 +87,16 @@ export default function EditorCard({ className }: { className?: string }) {
     });
   };
 
+  const handlePaste = (event: React.ClipboardEvent<HTMLTextAreaElement>) => {
+    const item = event.clipboardData.items[0];
```
MohamedBassem, line 91: Is the `[0]` here because there could be multiple files in the clipboard? If yes, should we just upload all of them in that case using the `useUploadAssets` hook?

MohamedBassem (approved):
Thanks, I didn't even know that this is possible! Left a couple of non-blocking comments.

MohamedBassem:
If you want to merge this PR before the `pdf` one, please rebase the `onPaste` commit over main to remove the `pdf` commit from that PR.

PR #95, opened 2024-04-11
feature: Recrawl failed links

AhmadMuj:
- Added a backoff for the Links Crawler Queue
- Added a new field in ( crawStatus ) in the bookmarkLinks table
- Added a new button the web to recrawl the failed URLs based on the crawStatus

### DIFF
--- a/apps/web/app/dashboard/admin/page.tsx
+++ b/apps/web/app/dashboard/admin/page.tsx
@@ -81,6 +81,21 @@ function ActionsSection() {
       },
     });
 
+  const { mutate: recrawlFailedLinks, isPending: isRecrawlFailedPending } =
+    api.admin.recrawlFailedLinks.useMutation({
+      onSuccess: () => {
+        toast({
+          description: "Recrawl Failed Links enqueued",
+        });
+      },
+      onError: (e) => {
+        toast({
+          variant: "destructive",
+          description: e.message,
+        });
+      },
+    });
+
   const { mutate: reindexBookmarks, isPending: isReindexPending } =
     api.admin.reindexAllBookmarks.useMutation({
       onSuccess: () => {
@@ -102,7 +117,15 @@ function ActionsSection() {
       <ActionButton
         className="lg:w-1/2"
         variant="destructive"
-        loading={isRecrawlPending}
+        loading={isRecrawlPending || isRecrawlFailedPending}
+        onClick={() => recrawlFailedLinks()}
+      >
+        Recrawl Failed Links Only
+      </ActionButton>
+      <ActionButton
+        className="lg:w-1/2"
+        variant="destructive"
+        loading={isRecrawlPending || isRecrawlFailedPending}
         onClick={() => recrawlLinks()}
       >
         Recrawl All Links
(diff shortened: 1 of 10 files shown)


apps/web/app/dashboard/admin/page.tsx
```diff
@@ -158,6 +181,10 @@ function ServerStatsSection() {
             <TableCell className="lg:w-2/3">Pending Crawling Jobs</TableCell>
             <TableCell>{serverStats.pendingCrawls}</TableCell>
           </TableRow>
+          <TableRow>
+            <TableCell className="lg:w-2/3">Failed Crawling Jobs</TableCell>
+            <TableCell>{serverStats.failedCrawls}</TableCell>
+          </TableRow>
```
MohamedBassem: How about we change the table to be a bit more 2d?

As in:

| | Pending | Failed |
-----|------|-----|
|Crawling jobs | 0 | 0 |
|Inference jobs | 0 | 0 |
|Search jobs | 0 | - |

packages/trpc/routers/admin.ts
```diff
@@ -59,7 +69,21 @@ export const adminAppRouter = router({
       ),
     );
   }),
-
+  recrawlFailedLinks: adminProcedure.mutation(async ({ ctx }) => {
```
MohamedBassem: It seems that we can easily add a `onlyFailures` in the existing `recrawlAllLinks` instead of creating a new endpoint?

packages/db/drizzle/0017_slippery_senator_kelly.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE bookmarkLinks ADD `crawlStatus` text DEFAULT 'pending';--> statement-breakpoint
+UPDATE bookmarkLinks SET crawlStatus = 'failure' where htmlContent is null;--> statement-breakpoint
+UPDATE bookmarkLinks SET crawlStatus = 'success' where htmlContent is not null;
```
MohamedBassem, line 3: I love that!

packages/trpc/routers/admin.ts
```diff
@@ -25,13 +26,21 @@ export const adminAppRouter = router({
...
         pendingIndexing,
         pendingOpenai,
       ] = await Promise.all([
         ctx.db.select({ value: count() }).from(users),
         ctx.db.select({ value: count() }).from(bookmarks),
-        LinkCrawlerQueue.getWaitingCount(),
+        ctx.db
+          .select({ value: count() })
+          .from(bookmarkLinks)
+          .where(eq(bookmarkLinks.crawlStatus, "pending")),
```
MohamedBassem, line 43: If we're going to do the 2d table, I think we can differentiate between the pending in database and pending in redis. I'm afraid of some jobs getting lost before getting enqueued to redis for example (or redis data getting wiped) which might cause confusion. So I think for both crawling and inference we can do:

```
crawlingStats: {
  waitingInRedis: number,
  pending: number,
  failed: number,
},
inference: {
 ... // Same
}
```

What do you think?

MohamedBassem (changes requested):
That was fast! It looks good to me, I requested only some minor changes :)

EDIT, can you also maybe rebase your branch to get the PDF commits out of the PR? They are not that big of a problem though :)

packages/trpc/routers/admin.ts
ahmadmucom: I would leave this into an upcoming commits honestly. 
I have a few ideas regarding the job queue in general I might discuss with you tomorrow.
For now I think it's enough, what do you think ?

MohamedBassem: Yeah, let's leave it for later then, I don't mind 👍

apps/web/app/dashboard/admin/page.tsx
ahmadmucom: For now the indexing and openai are going to be based off the queue status.

MohamedBassem:
Thanks again, I'm pretty sure a lot of people will find this to be super useful!

PR #117, opened 2024-04-22
Disable meilisearch analytics

vivekmiyani:


MohamedBassem:
oh I didn't know that meilisearch had analytics! Thanks for reporting that! To keep the env file simple, mind adding it to the docker compose file instead?

MohamedBassem:
Thank you!

Issue #144, opened 2024-05-09
Improve merging of tags by simple drag and drop


kamtschatka:
Currently merging of tags requires a lot of clicking, which can get tiresome, when you have to merge a lot (i guess the most common reason is, because inference creates a lot of tags)

I think a good way to allow merging in a fast way, would be drag and drop on the "Tags" page.
You see a list of all the tags and if you want to merge one with the other, you simple drag the one you want to get rid of onto the one that should stay and it gets merged right away.

MohamedBassem:
I like the idea :) I'll guess it's time for me to learn how make things draggable :D

kamtschatka:
I have spent quite some time looking at [react-beautiful-dnd](https://github.com/atlassian/react-beautiful-dnd) and [hello-pangea](https://github.com/hello-pangea/dnd), but seems like they only allows dragging in horizontal or vertical lists, so that does not work for the tags as they are in both directions (gave me some weird hover effects)
I also had a look at dndkit, but that seems to not support combining. I am checking out react-draggable soon, unless someone else knows a library that can be used for dragging/dropping tags onto each other to combine them.

kamtschatka:
OK, i got it to work with react-draggable (merged a tag just before taking the screenshot, that is why the toast is open):
![image](https://github.com/MohamedBassem/hoarder-app/assets/22147234/76d835a1-1d54-4efb-898a-777a82c5ab8d)

Basically I added the "Allow Dragging" checkbox to enable dragging. The reason is that having a link in the TagPill does not really work with the drag and drop and additionally I want to prevent people from merging tags accidentally, so I added the checkbox.

So a few questions:
* Is the use of react-draggable OK?
* Any suggestions for the feature/how the UI should look like?

Currently the code is quite messy, since it involved a lot of trial and error, so I have to first clean it up, but if you give some feedback, I'd clean it up and create a PR.

MohamedBassem:
@kamtschatka implemented it in https://github.com/MohamedBassem/hoarder-app/pull/154!

PR #154, opened 2024-05-15
Improve merging of tags by simple drag and drop #144

kamtschatka:
Added drag&drop functionality
Allowing sorting the tags by name, as this is more intuitive

Just a crude change to get some feedback. As I am no UI dev, I am thankful for any improvement ideas.

### DIFF
--- a/apps/web/components/dashboard/tags/AllTagsView.tsx
+++ b/apps/web/components/dashboard/tags/AllTagsView.tsx
@@ -1,6 +1,6 @@
 "use client";
 
-import Link from "next/link";
+import React from "react";
 import { ActionButton } from "@/components/ui/action-button";
 import ActionConfirmingDialog from "@/components/ui/action-confirming-dialog";
 import { Button } from "@/components/ui/button";
@@ -13,12 +13,15 @@ import InfoTooltip from "@/components/ui/info-tooltip";
 import { Separator } from "@/components/ui/separator";
 import { toast } from "@/components/ui/use-toast";
 import { api } from "@/lib/trpc";
-import { X } from "lucide-react";
+import Draggable, { DraggableData, DraggableEvent } from "react-draggable";
 
 import type { ZGetTagResponse } from "@hoarder/shared/types/tags";
-import { useDeleteUnusedTags } from "@hoarder/shared-react/hooks/tags";
+import {
+  useDeleteUnusedTags,
+  useMergeTag,
+} from "@hoarder/shared-react/hooks/tags";
 
-import DeleteTagConfirmationDialog from "./DeleteTagConfirmationDialog";
+import { TagPill } from "./TagPill";
 
 function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
   const { mutate, isPending } = useDeleteUnusedTags({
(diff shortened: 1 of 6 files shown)


MohamedBassem:
Hi, thanks a lot for working on this! I tried it locally and it's really really cool! I'll review the code now :)

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -54,48 +57,128 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
+  y: number;
 }
 
+const initialState: DragState = {
+  dragId: "",
+  dragOverId: "",
+  x: 0,
+  y: 0,
+};
+let currentState: DragState = initialState;
```
MohamedBassem: I'm not a react expert but I think global mutable state are frowned upon? Is it not possible to do this with a `React.useState` instead?

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -104,22 +187,73 @@ export default function AllTagsView({
...
+      </label>
+      <br />
+      <label>
+        <input
+          type="checkbox"
+          checked={sortByName}
+          onChange={handleSortByNameChange}
+        />
+        Sort by Name
+      </label>
```
MohamedBassem: I think those can probably be `shadcn` toggles (https://ui.shadcn.com/docs/components/toggle) and can probably be aligned right in the page.
If UI design is not your expertise, I'm happy to implement it myself after merging your PR.

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -54,48 +57,128 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
+
 export default function AllTagsView({
   initialData,
 }: {
   initialData: ZGetTagResponse[];
 }) {
+  const [draggingEnabled, setDraggingEnabled] = React.useState(false);
+  const [sortByName, setSortByName] = React.useState(false);
+
+  function handleSortByNameChange(): void {
```
MohamedBassem, line 93: minor comment: I think `toggleSortByName` is more descriptive here.

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -54,48 +57,128 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
+    setSortByName(!sortByName);
+  }
+
+  function handleDraggableChange(): void {
```
MohamedBassem, line 97: Same comment as using `toggleXX`.

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -54,48 +57,128 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
+  dragOverId: "",
+  x: 0,
+  y: 0,
+};
+let currentState: DragState = initialState;
+
+const byUsageSorter = (a: ZGetTagResponse, b: ZGetTagResponse) =>
+  b.count - a.count;
+const byNameSorter = (a: ZGetTagResponse, b: ZGetTagResponse) =>
+  a.name.localeCompare(b.name, undefined, { sensitivity: "base" });
```
MohamedBassem, line 71: Today I learned about `sensitivity: "base"`!

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -54,48 +57,128 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
+      }, 0);
+    }
+
+    currentState = initialState;
+  }
+
+  const { mutate: mergeTag } = useMergeTag({
+    onSuccess: () => {
+      toast({
+        description: "Tag has been updated!",
```
MohamedBassem: Small suggestion:

```suggestion
        description: "Tags has been merged!",
```

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -104,22 +187,73 @@ export default function AllTagsView({
   const tagsToPill = (tags: typeof allTags) => {
     let tagPill;
     if (tags.length) {
-      tagPill = tags.map((t) => (
-        <TagPill key={t.id} id={t.id} name={t.name} count={t.count} />
-      ));
+      tagPill = (
+        <div className="flex flex-wrap gap-3">
```
MohamedBassem, line 145: It seems that this `div` is redundant?

Screenshot 2024-05-16 at 10 31 42 PM

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -104,22 +187,73 @@ export default function AllTagsView({
...
+                  : undefined
+              }
+            >
+              <div
+                className="group relative flex cursor-grab"
+                data-id={t.id}
+                onDragOver={handleDrag}
+              >
+                <TagPill
+                  key={t.id}
```
MohamedBassem: Key is probably not need here anymore.

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -104,22 +187,73 @@ export default function AllTagsView({
...
+                !currentState.dragId
+                  ? { x: currentState.x ?? 0, y: currentState.y ?? 0 }
+                  : undefined
+              }
+            >
+              <div
+                className="group relative flex cursor-grab"
+                data-id={t.id}
+                onDragOver={handleDrag}
+              >
```
MohamedBassem: How about we pass `onDragOver` to `TagPill` and remove that div?

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -54,48 +57,128 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
-          <X className="size-3" />
-        </Button>
-      </DeleteTagConfirmationDialog>
-    </div>
-  );
+interface DragState {
+  dragId: string | null;
+  dragOverId: string | null;
+  x: number;
+  y: number;
```
MohamedBassem: Let's add a comment here and give them a bit more descriptive name?

```suggestion
// The position of the elements being dragged such that on drag over, we can revert the position.
  initialX: number;
  initialY: number;
```

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -54,48 +57,128 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
 }
 
+const initialState: DragState = {
+  dragId: "",
+  dragOverId: "",
```
MohamedBassem: minor comment: Maybe default to null instead?

```suggestion
  dragId: null,
  dragOverId: null,
```

MohamedBassem (changes requested):
Overall looks good to me! Thank you! Left a bunch of hopefully small comments.

apps/web/components/dashboard/tags/AllTagsView.tsx
kamtschatka: turns out the onDragOver is not actually needed, so I removed it.
If I remove the div though, it is suddenly not possible to drag it anymore. I am not sure why though...

apps/web/components/dashboard/tags/AllTagsView.tsx
kamtschatka: If I understand it correctly, it is needed at this place, because you can only return 1 root node, I am not a React expert though.
I have however removed it below, where the div was created 3 times.

apps/web/components/dashboard/tags/AllTagsView.tsx
kamtschatka: I am not an expert either^^
I have however extracted it out now, since that functionality is pretty much the same for the drag&drop functionality for assigning bookmarks to lists (next PR ;-) ), so it makes sense to reuse it.
It is now more encapsulated, so I guess that concern should be gone, but let me know.

apps/web/components/dashboard/tags/AllTagsView.tsx
kamtschatka: image
looks like this now

apps/web/components/dashboard/tags/AllTagsView.tsx
kamtschatka: So, after learning a lot, i found this:
useState needs to be used for the change detection to work, but the drag&drop handlers are quite slow to execute and react-draggable is not implemented that well. Due to the processing overhead, the <div> will drift away from the mouse over time (lagging behind) and it looks stupid. 
My followup PR for also implementing drag&drop for assigning bookmarks to lists has therefore some changes:
* The components themselves keep their state instead of the parent --> more fine grained rerender decisions possible
* most of the state is kept outside of useState and instead in a simple closure, to make sure everything is still fast enough
* the isDragging state is kept with useState, which makes a rerender possible, when the user starts to drag.

apps/web/lib/drag-and-drop.ts
```diff
@@ -0,0 +1,89 @@
...
+  initialX: number;
+  initialY: number;
+}
+
+export interface DragAndDropFunctions {
+  handleDragStart: (e: DraggableEvent, data: DraggableData) => void;
+  handleDrag: (e: DraggableEvent) => void;
+  handleDragEnd: () => void;
+  dragState: DragState;
+}
```
MohamedBassem, line 18: This interface seems to only be used as the return type of the `useDragAndDrop` hook. I think you can drop it completely and typescript will be able to infer it.

apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -55,47 +60,79 @@ function DeleteAllUnusedTags({ numUnusedTags }: { numUnusedTags: number }) {
...
+    useDragAndDrop(
+      "data-id",
+      "data-id",
+      (dragSourceId: string, dragTargetId: string) => {
+        mergeTag({
+          fromTagIds: [dragSourceId],
+          intoTagId: dragTargetId,
+        });
+      },
+    );
```
MohamedBassem, line 91: This is very very cool!

MohamedBassem (approved):
Thanks a lot of addressing the comments! I think there are some minor changes (e.g. adding some icons and margins to the buttons) but I'll take care of those!

Thanks again for working on this and also giving me a tool to be able to add support for drag and drops in other places. Very cool!

MohamedBassem:
@kamtschatka You can see some of the simplifications I made here (https://github.com/MohamedBassem/hoarder-app/commit/e5fd9eeca0d34658c9eed4f1a7d6ec25d4918488). Mainly:

1. Managed to get dragging of links to work by adding `data-id` on the link itself on the parent div.
2. Found a way that we won't need the `initialX` and `initialY`.
3. `handleDrag` turned out to not be needed as you only care when the drag is over.
4. Added some icons and help messages on the buttons.

Again thanks a lot for the idea and implementation!

MohamedBassem:
This is what it looks like now:

Screenshot 2024-05-18 at 12 19 43 PM

PR #162, opened 2024-05-17
Improve ollama tagging

kamtschatka:
Move the instructions to create tags and provide them as JSON to the end of the prompt to improve the result.


apps/workers/openaiWorker.ts
```diff
@@ -97,12 +97,17 @@ with the key "tags" and the value is list of tags. Don't wrap the response in a
 
 const TEXT_PROMPT_BASE = `
```
redoracle, line 98: I wish was possible to customise TEXT_PROMPT_BASE and TEXT_PROMPT_INSTRUCTIONS  directly within the settings menu in all apps.

Great idea a nice job so far! I look forward for the ollama integrated version!

apps/workers/openaiWorker.ts
```diff
@@ -97,12 +97,17 @@ with the key "tags" and the value is list of tags. Don't wrap the response in a
...
 I'm building a read-it-later app and I need your help with automatic tagging.
-Please analyze the text after the sentence "CONTENT START HERE:" and suggest relevant tags that describe its key themes, topics, and main ideas.
+Please analyze the text after the sentence "CONTENT START HERE" and "CONTENT END HERE" and suggest relevant tags that describe its key themes, topics, and main ideas.
```
MohamedBassem:
```suggestion
Please analyze the text between the sentences "CONTENT START HERE" and "CONTENT END HERE" and suggest relevant tags that describe its key themes, topics, and main ideas.
```

MohamedBassem (approved):
Tried it locally and seems to be working well! Thanks! Let's merge after that one small comment is fixed.

apps/workers/openaiWorker.ts
MohamedBassem: I don't want to give the people the freedom to add whatever instructions they want to be honest because it limits my ability to change or improve the prompt in the future. What I can do instead, is maybe allow the users to add some extra instructions at the end.

kamtschatka:
yeah I tried it with my exported bookmarks(~250) from chrome and only had a few untagged bookmarks, but that was due to no content being retrieved. Definitely better than before

MohamedBassem:
> yeah I tried it with my exported bookmarks(~250) from chrome and only had a few untagged bookmarks

That's amazing. Merged!

Issue #140, opened 2024-05-05
[Bug] Items created in the same second may not be shown in the bookmark grid due to pagination


MohamedBassem:
When we're paginating using `createdAt` only, items created in the same second can result in duplicates or missing items.

kamtschatka:
I ran into this issue also just now. Of course this happens a lot if you import a lot of bookmarks using the cli (or in my case the PR to allow importing bookmarks via the UI)

I noticed that it is an easy fix in schema.ts by changing `timestamp` to `timestamp_ms` in the createdAtField.
I tried running the database migrations (https://docs.hoarder.app/Development/database), but it only adds 
`CREATE INDEX `bookmarkLinks_url_idx` ON `bookmarkLinks` (`url`);`

I was hoping that would also create a migration of the existing timestamps, but seems like it doesn't.

MohamedBassem:
@kamtschatka Changing the mode of the column is a software layer thing. You'll have to create an empty migration with `pnpm drizzle-kit generate:sqlite --custom` and write some SQL to multiple all the `createdAt` columns by 1000. However, using milliseconds makes the chances of the problem happening lower but it doesn't really solve the problem. The proper fix is to paginate both on `createdAt` and the `id`. But we'll need to be careful if we're to do that in a backward compatible way.

As for what you got when you generated a migration, that's actually my fault. I apparently forgot to create a migration when I implemented the URL dedups :) Will get that fixed.

MohamedBassem:
Pushed https://github.com/MohamedBassem/hoarder-app/commit/c9dc23f43e0d87453b74105a8ffc81ce9f4bd98a to fix the dangling migration problem.

MohamedBassem:
I think fixing the pagination shouldn't be hard. I'll prioritize it specially in the context of your PR to support importing from the UI.

PR #163, opened 2024-05-17
Bookmark upload

kamtschatka:
Allowing customers to drag and drop bookmark files with the Netscape HTML format into the UI to add all the bookmarks
Updated the documentation


kamtschatka:
Should not be merged before #140 is fixed

MohamedBassem:
Thanks @kamtschatka!

Issue #123, opened 2024-04-25
[Feature request] Drag and Drop Items to Lists


MikeKMiller:
It would be helpful for organization to allow drag and drop from the main bookmark view to lists on the left menu. This would allow quickly moving items into the lists, versus the current (Click the 3 dots, Click Add to List, Click to select a List, Click Save)

MohamedBassem:
@kamtschatka you seem to have became an expert in drag and dropping! I like the feature! I think one thing that would make it look nicer is that when you're dragging the bookmark instead of changing the opacity, we can just replace the bookmark card with something simpler (e.g. a title only card for example).

kamtschatka:
I have pushed all my changes to a branch, that makes this in theory possible (i am able to change the dragged element, when it is dragged), but I am not skilled enough to get that to work^^
When you start dragging, it changes appearance, but that means:
* The new (smaller) div will be at the same top/left coordinates the previous div was --> the mouse cursor is not even on the element anymore. No idea how to fix that ^^
* Funnily the next bookmark will move into the new space. That will cause an image to be directly below the mouse, which will then start dragging the image, which will then cause the UI to show the "Drop your image" overlay, to upload images.... Also no idea how to fix that^^

kamtschatka:
OK, got something working, looks like this now:
![image](https://github.com/MohamedBassem/hoarder-app/assets/22147234/46f4a2f9-b0ff-4554-bb35-efe8299e6651)

MohamedBassem:
@kamtschatka looks amazing <3 Will review the PR hopefully either today or tomorrow. Thank you!

teunvg:
Any updates on this? I just started using Hoarder (love it!) and have a whole bunch of links I want to import, having this feature would be of great help in organizing them!

kamtschatka:
The feature is under discussion in discord. The question is, if this feature is obselete with the bulk edit feature (merged very soon).
If you can add some arguments for this feature or confirm that it is obsolete, once the bulk edit is live, that would be great.

teunvg:
@kamtschatka Thanks for the update! My take: bulk edit looks super nice, but to me it wouldn't replace drag and drop. 

To clarify: can you bulk add items to lists? The preview on the GitHub issue just shows archiving and favouriting. Either way, though, I think drag-and-drop is more intuitive (especially for single items), so I would be in favour of keeping it. If anything, I'd say drag and drop in bulk edit mode should move all selected items.

[...]

PR #164, opened 2024-05-18
[Feature request] Drag and Drop Items to Lists #123

kamtschatka:
reworked the drag and drop mechanism to still have change detection, but not so much that it has a huge overhead
Changed the layout a bit to allow proper drag/drop of elements from the main section to the sidebar
Added the possibility to drag/drop bookmarks onto lists

With this, it is now possible to drag a bookmark and it will not drag the whole thing, but a small placeholder (can still be changed, i just added a placeholder for now)


kamtschatka:
as discussed, removed the actual drag and drop functionality for bookmarks, so we can finally merge this.
Followup PR for reference: https://github.com/hoarder-app/hoarder/pull/264

Issue #158, opened 2024-05-16
Allow adding multiple URLs at once


kamtschatka:
When you have a lot of URLs at hand, it would be great if you can put them all into hoarder at once.
A simple solution would be, if you drop a list of URLs in the "new item" and press "Save", it checks if this is just a list of URLs separated by new-line characters. If it is, then ask if it should be a note, or if the URLs should be imported as separate Bookmarks.

MohamedBassem:
hmmm, I haven't so far implemented any mass importing functionality in the app itself because you can do it with the CLI (https://docs.hoarder.app/import). Though I understand that probably not everyone would be familiar with how to use CLIs.

MohamedBassem:
Implemented by @kamtschatka in #167!

PR #167, opened 2024-05-19
Allow adding multiple URLs at once #158

kamtschatka:
Added a reusable dialog
opening a dialog that allows you to decide if you want to import multiple URLs at once if you provide only that

### DIFF
--- a/apps/web/components/dashboard/bookmarks/EditorCard.tsx
+++ b/apps/web/components/dashboard/bookmarks/EditorCard.tsx
@@ -1,9 +1,10 @@
 import type { SubmitErrorHandler, SubmitHandler } from "react-hook-form";
-import { useEffect, useImperativeHandle, useRef } from "react";
+import React, { useEffect, useImperativeHandle, useRef } from "react";
 import Link from "next/link";
 import { ActionButton } from "@/components/ui/action-button";
 import { Form, FormControl, FormItem } from "@/components/ui/form";
 import InfoTooltip from "@/components/ui/info-tooltip";
+import MultipleChoiceDialog from "@/components/ui/multiple-choice-dialog";
 import { Separator } from "@/components/ui/separator";
 import { Textarea } from "@/components/ui/textarea";
 import { toast } from "@/components/ui/use-toast";
@@ -34,8 +35,16 @@ function useFocusOnKeyPress(inputRef: React.RefObject<HTMLTextAreaElement>) {
   }, [inputRef]);
 }
 
+interface MultiUrlImportState {
+  urls: URL[];
+  text: string;
+}
+
 export default function EditorCard({ className }: { className?: string }) {
   const inputRef = useRef<HTMLTextAreaElement>(null);
+
+  const [multiUrlImportState, setMultiUrlImportState] =
+    React.useState<MultiUrlImportState | null>(null);
 
   const demoMode = !!useClientConfig().demoMode;
   const formSchema = z.object({
@@ -76,14 +85,31 @@ export default function EditorCard({ className }: { className?: string }) {
     },
   });
 
-  const onSubmit: SubmitHandler<z.infer<typeof formSchema>> = (data) => {
-    const text = data.text.trim();
-    try {
-      const url = new URL(text);
+  function tryToImportUrls(text: string): void {
+    const lines = text.split("\n");
+    const urls: URL[] = [];
+    for (const line of lines) {
+      // parsing can also throw an exception, but will be caught outside
+      const url = new URL(line);
       if (url.protocol != "http:" && url.protocol != "https:") {
         throw new Error("Invalid URL");
       }
+      urls.push(url);
+    }
+
+    if (urls.length === 1) {
+      // Only 1 url in the textfield --> simply import it
       mutate({ type: "link", url: text });
+      return;
+    }
+    // multiple urls found --> ask the user if it should be imported as multiple URLs or as a text bookmark
+    setMultiUrlImportState({ urls, text });
+  }
+
+  const onSubmit: SubmitHandler<z.infer<typeof formSchema>> = (data) => {
+    const text = data.text.trim();
+    try {
+      tryToImportUrls(text);
     } catch (e) {
       // Not a URL
       mutate({ type: "text", text });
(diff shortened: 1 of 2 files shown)


apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -150,6 +177,50 @@ export default function EditorCard({ className }: { className?: string }) {
...
+                  setOpen(false);
+                  setImportData(initialImportData);
+                }}
+              >
+                Import as separate Bookmarks
+              </ActionButton>
+            ),
+          ]}
+        >
+          <div></div>
```
MohamedBassem: Triggers are optional in dialogs. So if you modify `MultipleChoiceDialog` to have an optional `children` and only create the trigger node if `children` is set, you won't need this.

apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -34,9 +35,17 @@ function useFocusOnKeyPress(inputRef: React.RefObject<HTMLTextAreaElement>) {
...
+const initialImportData = {
+  urls: [] as URL[],
+  text: "",
+};
+
 export default function EditorCard({ className }: { className?: string }) {
   const inputRef = useRef<HTMLTextAreaElement>(null);
 
+  const [open, setOpen] = React.useState(false);
+  const [importData, setImportData] = React.useState(initialImportData);
```
MohamedBassem: There's some room for simplification here:

1. Have only one state which is `multiURLImportState`. This state can be either set when we detect a multi-url input, or null if we didn't (with null being the default)`.
2. Remove the `open` control state for the dialog and instead only render the dialog if `multiURLImportState` is not null. In react, you typically do that with `(multiURLImportState && <MultiInputDialog> ...)`. The dialog will need to set `defaultOpen=true` so that it's open when it's rendered.
3. When the user takes an action or `onOpenChange`, just set `multiURLImportState` to null which will remove the dialog.

What do you think?

MohamedBassem (approved):
Overall, looks good to me and I really like the feature and I like that we're giving the user the choice! Great job.

I've suggested some simplifications ideas :)

apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -150,6 +176,43 @@ export default function EditorCard({ className }: { className?: string }) {
               : "Press ⌘ + Enter to Save"
             : "Save"}
         </ActionButton>
+
+        {multiUrlImportState && (
+          <MultipleChoiceDialog
```
MohamedBassem, line 181: The user can still dismiss the dialog manually (by clicking outside the dialog box for example), and in that case, you won't be able to open the dialog again.

You'll still need to subscribe to `onOpenChange` and reset the state if it gets closed.
```suggestion
          <MultipleChoiceDialog
            onOpenChange={(open) => {
              if (!open) {
                setMultiUrlImportState(null);
              }
            }}
          
```

apps/web/components/ui/multiple-choice-dialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+  children,
+}: {
+  open?: boolean;
+  title: React.ReactNode;
+  description: React.ReactNode;
+  actionButtons: (() => React.ReactNode)[];
+  children?: React.ReactNode;
+}) {
+  return (
+    <Dialog defaultOpen={true}>
```
MohamedBassem: Given that this is a generic dialog (in the `ui` dir), you'll want to make this as a prop.

apps/web/components/ui/multiple-choice-dialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+  DialogTrigger,
+} from "@/components/ui/dialog";
+
+export default function MultipleChoiceDialog({
+  title,
+  description,
+  actionButtons,
+  children,
+}: {
+  open?: boolean;
```
MohamedBassem, line 20: This doesn't seem to be used anymore?

apps/web/components/ui/multiple-choice-dialog.tsx
```diff
@@ -0,0 +1,36 @@
...
+  title: React.ReactNode;
+  description: React.ReactNode;
+  actionButtons: (() => React.ReactNode)[];
+  children?: React.ReactNode;
+}) {
+  return (
+    <Dialog defaultOpen={true}>
+      {children ? <DialogTrigger asChild>{children}</DialogTrigger> : ""}
```
MohamedBassem:
```suggestion
      {children && <DialogTrigger asChild>{children}</DialogTrigger>}
```

MohamedBassem (changes requested):
After the changes, there's small new bug when the user clicks outside of the dialog. Repro:

1. Input multiple URLs & click submit
2. Click outside the dialog, the dialog will get closed.
3. Try clicking submit again, nothing will happen.

apps/web/components/ui/multiple-choice-dialog.tsx
kamtschatka: ok now I am confused. If the defaultOpen will no longer be true, then I will have to bring back the open function, otherwise this control will never show?
I was under the impression the goal was to make this always show, otherwise don't you have 2 ways of opening and closing it? (one with calling the open function and 1 by configuring it with defaultOpen=true)

MohamedBassem: @kamtschatka Your new component is a generic dialog. So you'll need to pass params similar to that of the dialog to allow the different usecases to customize it. In that case, you'll want to expose `open?`, `onOpenChange?` and `defaultOpen?` (and forward them to `Dialog`). Then our usecase (which is the confirmation of multi URL), is the one that won't pass `open`, but will set `defaultOpen` to `true`, does that make sense?

kamtschatka: sure, i was just confused that above you said to remove the `open` control state and below you said `open` is no longer used.
So basically it was good as it was before, the idea was more to add an additional way to allow showing the dialog without calling `setOpen`.

MohamedBassem: `open` should continue being there, but you'll need to make it optional and pass it to the dialog. Currently, it's there, but not being used.

The whole suggestion for simplification was yes to not have a `controlled` open property for the dialog. We just render it when needed it, and un-render it when it's no longer needed. All of that while keeping the option to use `MultiInputDialog` in a controlled fashion in a different place if we want to (aka keeping it generic)

kamtschatka: ok, but then I have to get rid of the defaultOpen again, because then I would have the `open` parameter and the `defaultOpen` parameter, where 1 could be true and 1 could be false, which would be weird. 
I'll just pass true to open, without having a state to keep it, because that is implied by the MultiUrlImportState.

MohamedBassem: @kamtschatka you need both. For example this the API of the dialog component (https://www.radix-ui.com/primitives/docs/components/dialog#root) we're using. As you can see, it has both.

`open` is a controlled param. If you pass open to `true` (instead of `defaultOpen`) people won't be able to close the dialog at all. `defaultOpen` just initializes the initial state of the dialog.

So in our case, we need `defaultOpen=true` but we need to tell the dialog component that it's not a controlled dialog by setting `open=undefined` in other words, not passing `open`.

kamtschatka: but we are controlling the rendering state through `{multiUrlImportState && (....`, so if we don't want the dialog to show, we are not even rendering it.
And as for reusability: if you want to control the open state, then you pass in a proper state?

I will push again, for you to review, I am very convinced that the defaultOpen AND open state are not required at the same time.

MohamedBassem: I think functionality wise, this looks good. But with the current MultiInputDialog implementation, users won't be able to do an uncontrolled dialog with a trigger. I'll followup with a small PR to show you what I had in mind.

MohamedBassem (approved):
I think functionality wise, this PR looks good to me. Let me merge it and send a followup minor modifications to explain my point I guess.

Issue #150, opened 2024-05-11
`bookmark list` output is not a valid JSON


jegwastaken:
With the CLI, `hoarder bookmarks list > bookmarks.json` kinda works, but it outputs malformed/invalid JSON.

MohamedBassem:
hmmm, the output of `bookmark list` is supposed to be a well-formed JSON. Just checked and I think I saw the bug. Dates are not quoted. I'll fix it :) Thanks for the report!

jegwastaken:
I think both keys and string values should be in double quotes. Other JSON issues I noticed: 

- Some descriptions contain multiple quoted parts (and sometimes a plus sign) for some reason. 
- `htmlContent: '<div id="r... <CROPPED>'` seems problematic. Maybe a simpler `"htmlContent": "<CROPPED>"` would be better?

Example output:

```
[
  {
    id: 'xxxxxxxxx',
    createdAt: 2024-05-11T16:51:13.000Z,
    title: null,
    archived: false,
    favourited: false,
    taggingStatus: 'success',
    note: null,
    tags: [
      'Literature',
      'Fiction Writing',
      'Ernest Hemingway',
      'Writing Advice',
      'Author Techniques'
    ],
    content: {
      type: 'link',
      url: 'http://www.openculture.com/2013/02/seven_tips_from_ernest_hemingway_on_how_to_write_fiction.html',
      title: 'Seven Tips From Ernest Hemingway on How to Write Fiction',
      description: 'Image by Lloyd Arnold via Wikimedia Commons\n' +
        'Before he was a big game hunter, before he was a deep-sea fisherman, Ernest Hemingway was a craftsman who would rise very early in the morning and write.',
      imageUrl: 'https://cdn8.openculture.com/wp-content/uploads/2013/02/EH-354-e1361297347123.jpeg',
      imageAssetId: '43768e6a-bcc7-4655-9dfe-f69935d847ac',
      screenshotAssetId: '156d4d7a-c5a8-47a3-b067-2b32f7d03a92',
      favicon: 'https://cdn8.openculture.com/2017/06/17162746/OC-favicon-300x300.png',
      htmlContent: '<div id="r... <CROPPED>',
      crawledAt: 2024-05-11T18:08:22.000Z
    }
  },
  ...
]
```

kamtschatka:
I had a look at the issue and the problem is, that console.dir does not perform a proper encoding, so I switched it to JSON.stringify with 4 space indentation to have the same formatting.

[...]

PR #168, opened 2024-05-19
bookmark list output is not a valid JSON #150

kamtschatka:
fixed the encoding issue by using json.stringify to get parsable JSON

### DIFF
--- a/apps/cli/src/commands/bookmarks.ts
+++ b/apps/cli/src/commands/bookmarks.ts
@@ -136,7 +136,7 @@ bookmarkCmd
       results = [...results, ...resp.bookmarks];
     }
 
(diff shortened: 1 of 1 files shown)


MohamedBassem (changes requested):
My issue with this is that we lose the coloring for everyone.

My plan for that particular issue was what you mentioned in the issue itself. Let's create a global `--json` flag, and then have a reusable `printObject` utility that uses `console.dir` if the flag is not there (the default) and json serializes the response if it's there.

MohamedBassem:
On a related note, @kamtschatka I really appreciate the work you've been doing for hoarder! Would be interested in joining the `#development` channel on discord so that we can coordinate ideas before implementing them? :)

kamtschatka:
sure, i was wondering where you coordinate ;-)

PR #176, opened 2024-05-24
Switched to cursorV2 to make use of the fix for #140

kamtschatka:
realized that the cli is not using cursorV2 and is causing an infite loop/going out of memory if bookmarks created at the same second exist


MohamedBassem:
Ah, good catch!

Issue #195, opened 2024-06-05
Any plans to support copy to clipboard (markdown code) for notes?


ItsNoted:
Love this but I’m curious if there will ever be an option to past code into notes we can easily copy to clipboard. I like to save snippets and compose stacks.

kamtschatka:
are you talking about the whole note, or about separate code blocks you have added using markdown? (with 3 `)
I like both ideas actually.

ItsNoted:
Just the code blocks with copy to clipboard. 😊

kamtschatka:
I managed to do this:
![image](https://github.com/hoarder-app/hoarder/assets/22147234/1e8432c8-8925-4272-8c55-4c80795bdac8)
![image](https://github.com/hoarder-app/hoarder/assets/22147234/c8f7f9e6-2bc4-4339-819c-fcef24398d00)

When you copy it, you get a short indicator: 
![image](https://github.com/hoarder-app/hoarder/assets/22147234/f329e563-b7fd-4b71-847e-ea0c7a8538f4)

The black border around the syntax highlighter is not perfect, but tbh, i have no idea how to get rid of it^^.
Maybe @MohamedBassem or someone else knows.

I will create a PR so people can have a look.

ItsNoted:
Looking good!

ItsNoted:
What version can I grab to try it out for docker or do I need to build it myself?

kamtschatka:
I have only opened the PullRequest so far. Right now you would need to pull from that branch and build it yourself to try it out.
To have it in the docker image, the PR needs to be merged and a new version needs to be released.

ItsNoted:
This really could replace the amazing project Snippet-Box which has been abandoned for quite some time now. https://github.com/pawelmalak/snippet-box. Take a look at that project and you can see how Hoarder has some similar attributes. This is why I love this project so much because it reminds me a lot of Snippet-Box. I'm excited to see where it goes :)

ItsNoted:
Thanks for adding this! It didn't work locally but when I tried it on my cloudflare tunnel it worked. Not sure why but maybe something to do with my update process.

MohamedBassem:
@ItsNoted chrome doesn't allow accessing the clipboard except when using https or localhost. So it won't work on http unfortunately.

kamtschatka:
there should be a tooltip though, that explains this?

PR #197, opened 2024-06-05
Any plans to support copy to clipboard (markdown code) for notes? #195

kamtschatka:
added a button to copy the markdown and added code highlighting

1 component seems unused
I am not sure how to test the mobile version/if that should also use the new component


apps/mobile/components/bookmarks/BookmarkCard.tsx
```diff
@@ -227,6 +227,7 @@ function TextCard({ bookmark }: { bookmark: ZBookmark }) {
     throw new Error("Wrong content type rendered");
   }
   const content = bookmark.content.text;
+  // TODO: replace this markdown also? I am not really sure how to test if everything still works
```
MohamedBassem: Don't worry about mobile markdown for now, we still don't show full note contents there anyways.

apps/web/components/dashboard/bookmarks/TextCard.tsx
```diff
@@ -45,18 +45,15 @@ export default function TextCard({
 
   return (
     <>
+      {/* TODO: Where is that used? Is it simply unused?*/}
```
MohamedBassem: This is the bookmark card shown in the BookmarkGrid.

Screenshot 2024-06-07 at 11 03 14 AM

apps/web/components/ui/code-copy-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+    "props" in child &&
+    "children" in child.props
+  ) {
+    const element = child as React.ReactElement<{ children: string }>;
+    return element.props.children;
+  }
+  return "";
+}
+
+export default function CodeCopyBtn({
```
MohamedBassem: This looks a lot like the button in `apps/web/components/dashboard/settings/AddApiKey.tsx` that you use to copy the newly created API key. Can we maybe make it a generic `CopyButton` and re-use it in those places?

And the `getStringToCopy` can be a callback passed to this component.

An the

apps/web/components/ui/code-copy-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+  const handleClick = async () => {
+    await navigator.clipboard.writeText(getStringToCopy(child));
+    setCopyOk(true);
+    setTimeout(() => {
+      setCopyOk(false);
+    }, 500);
+  };
+
+  return (
+    <div className="absolute right-2 top-2 transform cursor-pointer text-xl text-white transition-all duration-300 ease-in-out hover:scale-110 hover:opacity-90">
```
MohamedBassem: If you're going to make it generic, some of those classes will need to passed from the parent instead of being inlined in here.

apps/web/components/ui/code-copy-button.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+export default function CodeCopyBtn({
+  children: child,
+}: {
+  children: React.ReactNode;
+}) {
+  const [copyOk, setCopyOk] = React.useState(false);
+
+  const handleClick = async () => {
+    await navigator.clipboard.writeText(getStringToCopy(child));
```
MohamedBassem: I found out that if you're not using `localhost` or `https`, `navigator.clipboard` becomes undefined ([source](https://stackoverflow.com/questions/51805395/navigator-clipboard-is-undefined)). This is a bug even in my implementation in the api key button. Let's maybe make this button disabled if `navigator.clipboard` is undefined.

Screenshot 2024-06-07 at 10 55 05 AM

apps/web/components/ui/markdown-component.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { dracula } from "react-syntax-highlighter/dist/cjs/styles/prism";
+
+export function MarkdownComponent({
+  children: markdown,
+}: {
+  children: string;
+}) {
+  const Pre = ({ children }: { children?: React.ReactNode }) => (
+    <pre className="relative mx-auto mb-12 shadow-lg">
+      <CodeCopyBtn>{children}</CodeCopyBtn>
```
MohamedBassem: It doesn't sound right that the `CopyBtn` wraps the children. I think the proper way of doing it is creating a reference with `React.useRef`, setting it as the reference of the inner component with `ref={ref}` and then on the on then on the button `onCopy` or `onClick`, you do `navigator.writeClipboard(ref.innerHTML)`.

apps/web/components/ui/markdown-component.tsx
```diff
@@ -0,0 +1,46 @@
...
+  );
+  return (
+    <Markdown
+      className="prose mx-auto dark:prose-invert"
+      components={{
+        pre: Pre,
+        code({ className, children, ...props }) {
+          const match = /language-(\w+)/.exec(className ?? "");
+          return match ? (
+            // @ts-expect-error i have absolutely no idea what it complains about and passing refs to it also does not solve it
```
MohamedBassem: What is the error you're getting?

MohamedBassem (commented):
Nice! Left a bunch of small comments about reusing some of the exiting code.

apps/web/components/dashboard/bookmarks/TextCard.tsx
kamtschatka: the thing is, if I remove the BookmarkedTextViewer, nothing changes, so I am not quite sure what it does?
Hovering the element in the React Extension does not highlight it either and previewModalOpen is only bound to this component, nothing else, so I don't think that value can actually change?

apps/web/components/ui/markdown-component.tsx
kamtschatka: with the getStringToCopy function, this isn't actually needed. That was just to retrieve the innerHTML, but you can do that differently, which is nice.

apps/web/components/ui/markdown-component.tsx
kamtschatka:
```
Types of property ref are incompatible.
Type LegacyRef<HTMLElement> | undefined is not assignable to type LegacyRef<SyntaxHighlighter> | undefined
```

```
Types of property ref are incompatible.
Type LegacyRef<HTMLElement> | undefined is not assignable to type LegacyRef<SyntaxHighlighter> | undefined
Type (instance: HTMLElement | null) => void is not assignable to type LegacyRef<SyntaxHighlighter> | undefined
Type (instance: HTMLElement | null) => void is not assignable to type (instance: SyntaxHighlighter | null) => void
Types of parameters instance and instance are incompatible.
Type SyntaxHighlighter | null is not assignable to type HTMLElement | null
Type Component<SyntaxHighlighterProps, {}, any> is missing the following properties from type HTMLElement:
accessKey, accessKeyLabel, autocapitalize, dir
, and 290 more.
```

Issue #188, opened 2024-06-01
Feature request: Browser extension "Open your Hoarder saves"


Deathproof76:
I have one QoL/feature request for the browser extension: Would it be possible to integrate a link to ones own hoarder page? Similar to the pocket extension.

Like, right click on the addon in the toolbar and the second selection is something like "Open your Hoarder saves" -> click it -> and a new tab opens the hoarder instance. The url link used should be the same the addon already knows.

MohamedBassem:
Sounds like something we can easily support indeed :)

PR #206, opened 2024-06-06
Feature request: Browser extension "Open your Hoarder saves" #188

kamtschatka:
Adds a context menu entry when a hoarder instance is configured and removes it again, if it is not configured anymore

### DIFF
--- a/apps/browser-extension/src/background/background.ts
+++ b/apps/browser-extension/src/background/background.ts
@@ -1,10 +1,14 @@
-import { Settings } from "../utils/settings.ts";
+import {
+  getPluginSettings,
+  Settings,
+  subscribeToSettingsChanges,
+} from "../utils/settings.ts";
 
 const OPEN_HOARDER_ID = "open-hoarder";
 
 function checkSettingsState(settings: Settings) {
   if (settings?.address) {
-    registerContextMenu(settings.address);
+    registerContextMenu();
   } else {
     chrome.contextMenus.remove(OPEN_HOARDER_ID);
   }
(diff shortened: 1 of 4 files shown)


apps/browser-extension/src/background/background.ts
```diff
@@ -0,0 +1,45 @@
...
+    const { menuItemId } = info;
+    if (menuItemId === OPEN_HOARDER_ID) {
+      chrome.tabs.create({ url: address, active: true });
+    }
+  };
+}
+
+chrome.storage.sync.get("settings", (items) =>
+  checkSettingsState(items.settings as Settings),
+);
```
MohamedBassem: This is leaking the way settings are stored. We already have a way to fetch the settings in `settings.ts`, so Replace that with:
```suggestion
getPluginSettings().then((settings: settings) =>
  checkSettingsState(settings),
);

```

apps/browser-extension/src/background/background.ts
```diff
@@ -0,0 +1,45 @@
...
+chrome.storage.sync.onChanged.addListener((changes) => {
+  checkSettingsState(changes.settings.newValue as Settings);
+});
```
MohamedBassem: Same, I'd move this to `settings.ts` and call it:

```
subscribeToSettingsChanges(callback: (settings: Settings) => void)`
```

apps/browser-extension/src/background/background.ts
```diff
@@ -0,0 +1,45 @@
...
+ * @param address The address of the configured hoarder instance
+ */
+function registerContextMenu(address: string) {
+  chrome.contextMenus.create({
+    id: OPEN_HOARDER_ID,
+    title: "Open Hoarder",
+    contexts: ["action"],
+  });
+  chrome.contextMenus.onClicked.addListener(createContextClickHandler(address));
+}
```
MohamedBassem: I don't think we should be adding a listener every time the setting changes. We should just always register the listener, and inside the listener (at top level), and inside the listener, we grab the setting object and open the address if it's set.

MohamedBassem (commented):
As always, thanks for working on this!

Issue #139, opened 2024-05-05
hoarder --version reports version 0.0.0


skirmess:
The npm CLI reports

```
$ hoarder --version
0.0.0
```

but installed version is 0.13.2:

```
$ more package.json
{
  "dependencies": {
    "@hoarderapp/cli": "^0.13.2"
  }
}
```

kamtschatka:
So the reason is that the version is taken from `process.env.npm_package_version` and if that does not exist, it takes 0.0.0.
I guess a replacement step during the build that bakes this version in would make sense? @MohamedBassem?

PR #212, opened 2024-06-08
hoarder --version reports version 0.0.0 #139

kamtschatka:
added a build step after vite, to replace the fallback 0.0.0 value with the version from the package.json

### DIFF
--- a/apps/cli/.gitignore
+++ b/apps/cli/.gitignore
@@ -1,1 +1,2 @@
 dist
+/buildscripts/*.js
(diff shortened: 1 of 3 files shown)


MohamedBassem:
I didn't like the "find and replace" approach here to be honest. I found a way to let vite inline some variables during build, so used that instead here: https://github.com/hoarder-app/hoarder/commit/f7a77533240ec435c8a7b103b6f6be409bf995d8

kamtschatka:
TIL

Issue #28, opened 2024-03-27
feature request: pdf support


asg0451:
I'd like to be able to capture a pdf url, eg https://gavinadair.files.wordpress.com/2017/03/baker-changes-of-mind.pdf

currently, it is captured but no tags or added nor is text extracted
![image](https://github.com/MohamedBassem/hoarder-app/assets/4358545/3f31cbf2-4b34-4f1f-9698-3833e5685a91)

logs:

```
hoarder-workers 2024-03-27T16:53:27.624Z info: [Crawler][9] Will crawl "https://gavinadair.files.wordpress.com/2017/03/baker-changes-of-mind.pdf" for link with id "h03n4dihn2gp0kn8giwiyir7"                                                                                           hoarder-workers 2024-03-27T16:53:27.813Z info: [search][30] Completed successfully                                                                                                                                                                                                      hoarder-workers 2024-03-27T16:53:27.822Z error: [Crawler][9] Crawling job failed: {}                               
```

MohamedBassem:
Yeah, only html `Content-Type` currently works. PDF support is a reasonable feature request though. Will add it to the backlog. Thanks!

MarkLuk:
Exactly my use-case! I research & bookmark a lot of PDF files. Would like to have support to view their content in the preview.

PR #214, opened 2024-06-09
feature request: pdf support #28

kamtschatka:
extended the database to allow storing pdf assets alongside links added downloading of pdfs
added aiinference for pdfs
updated the UI to display the same as for asset bookmarks

I am not sure if there are other locations where pdf support would be needed.
The preview and hte bookmarks overview look the same as the assetId.
The content of the pdf is stored in the db and used for reindexing


MohamedBassem (changes requested):
As we discussed on discord, we'll instead change the bookmark type from links to assets if we detect that it's one of the supported assets.

PR #216, opened 2024-06-10
feature request: pdf support #28

kamtschatka:
Added a new sourceUrl column to the asset bookmarks
Added transforming a link bookmark pointing at a pdf to an asset bookmark
made sure the "View Original" link is also shown for asset bookmarks that have a sourceURL
updated gitignore for IDEA

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,7 @@ data
 
 # Turbo
 .turbo
+
+# Idea
+.idea
+*.iml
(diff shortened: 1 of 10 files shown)


MohamedBassem:
PDF parsing as of today, happens in the openAI worker. Not optimal (for a lot of reasons), but for consistency, I'll leave there. We can address that later.

MohamedBassem:
This is pretty cool!

Issue #215, opened 2024-06-09
Allow downloading more content from a webpage and index it


kamtschatka:
I regularly bookmark youtube videos, Instagram videos, other videos.
It is not assured, that those videos stay online forever, so I prefer to download important videos (yes I am a real hoarder).

Would be great if you can enable downloading videos and serving them from hoarder for later viewing (Filesize does not matter to me, but I guess for some it matters).

Would be also great if the subtitles would be downloaded and indexed, so searching is possible also in the video content.
In the long run it would also be cool if it is possible to transcribe the video contents and make it searchable that way.

OliverLippertVw:
>It is not assured, that those videos stay online forever

True. I was planning a long (1 month) trip for over half a year and even in this "short" period of time we had the one or other reel that doesn't exist any longer.

Not to start telling about my various collection of receipts (10+ years) -.-

So linking but also downloading + indexing is a huge benefit, I love to see as well.

huyz:
For media downloading, I currently use ArchiveBox.

khronimo:
I would love to see it being able to index the transcript in Youtube links. I often bookmark interesting videos, interviews and am always frustrated when i try to find that particular resource again later.

ItsNoted:
Are we able to use Hoarder to archive Youtube transcripts? I wasn't sure if this was added or not.

dimitrieh:
> Would be also great if the subtitles would be downloaded and indexed, so searching is possible also in the video content.
In the long run it would also be cool if it is possible to transcribe the video contents and make it searchable that way.

> Are we able to use Hoarder to archive Youtube transcripts? I wasn't sure if this was added or not.

When using yt-dlp args: Files are downloaded but not indexed - Karakeep downloads .srt and/or metadata .json files but doesn't parse them into the bookmark's searchable text

Follow up issues:
- https://github.com/karakeep-app/karakeep/issues/1629
- https://github.com/karakeep-app/karakeep/issues/1442

PR #220, opened 2024-06-12
Changes to have more assets #215

kamtschatka:
The first PR to remove the assetids from the bookmarks table and move them to a new table
updated the UI to use the new format


apps/workers/crawlerWorker.ts
```diff
@@ -457,20 +471,42 @@ async function runCrawler(job: Job<ZCrawlLinkRequest, void>) {
...
+        imageUrl: meta.image,
+        favicon: meta.logo,
+        content: readableContent?.textContent,
+        htmlContent: readableContent?.content,
+        crawledAt: new Date(),
+      })
+      .where(eq(bookmarkLinks.id, bookmarkId));
+
+    if (screenshotAssetId) {
+      if (oldScreenshotAssetId) {
```
kamtschatka, line 544: these sections are 3 times the same, probably can be extracted out?

PR #223, opened 2024-06-14
add Arch Linux installation

devome:
add Arch Linux installation

### DIFF
--- a/docs/docs/02-Installation/03-archlinux.md
+++ b/docs/docs/02-Installation/03-archlinux.md
@@ -2,6 +2,8 @@
 
 ## Installation
 
+> [Hoarder on AUR](https://aur.archlinux.org/packages/hoarder) is not maintained by the hoarder official.
+
 1. Install hoarder
 
     ```shell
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Oh wow, that I didn't see coming :) Thanks a lot for submitting this to arch's registry. Pardon my ignorance with the arch model, but I'm wondering, what's the maintenance model here? For example, with version updates, is it going to be on you to notice that there's a version upgrade an update the arch package? Same with new package dependencies, etc? Does the arch registry accept PRs (or whatever they are called there) from people other than the maintainer?

As for the PR itself, I'll have a look on the PKGBUILD file and merge afterwards :)

devome:
If Hoarder has an update, the package on AUR will also be updated automatically by my [nvcheck](https://github.com/lilydjwg/nvchecker) scripts. If there is a change in dependency, I usually notice it, or other users will remind me to add it when they notice it. 

AUR can add co-maintainers and maintain packages together. If you are willing to participate in the maintenance work of `hoarder` on AUR, I can add you as a co-maintainer.

MohamedBassem:
Thanks a lot for your effort adding Hoarder to arch. I'm not an arch user myself, but if you need any help maintaining the package, please let me know and I'd be happy to help :)

devome:
@MohamedBassem That's OK, thank you.

Issue #225, opened 2024-06-14
Empty bookmark


arobase-che:
Hello,

What about empty bookmark?
I have the feeling that I should not be able to « Save » an empty bookmark.
Maybe the Save button and ||Ctrl||  + ||Enter||  should be disabled when the « New item » textbox is empty?

Or maybe en empty bookmark should not be processed and early discarded ?

Thank you

MohamedBassem:
This is good call. We shouldn't allow empty bookmarks :)

PR #227, opened 2024-06-15
Empty bookmark #225

kamtschatka:
preventing the creation of empty bookmarks in the UI and the rest api
updated gitignore

### DIFF
--- a/packages/trpc/routers/bookmarks.ts
+++ b/packages/trpc/routers/bookmarks.ts
@@ -204,6 +204,13 @@ export const bookmarksAppRouter = router({
         if (alreadyExists) {
           return { ...alreadyExists, alreadyExists: true };
         }
+      } else if (input.type === "text") {
+        if (!input.text) {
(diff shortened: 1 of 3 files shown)


packages/trpc/routers/bookmarks.ts
```diff
@@ -204,6 +204,13 @@ export const bookmarksAppRouter = router({
         if (alreadyExists) {
           return { ...alreadyExists, alreadyExists: true };
         }
+      } else if (input.type === "text") {
+        if (!input.text) {
+          throw new TRPCError({
+            message: "Creating empty bookmarks is not allowed",
+            code: "BAD_REQUEST",
+          });
+        }
```
MohamedBassem, line 213: We don't typically do input validation on the code. This kind of validation should live in the zod schema. However, I don't really think this is something worth enforcing on the server side anyways.

MohamedBassem (changes requested):

PR #228, opened 2024-06-17
feature: support pasting image in the textfield

AhmadMuj:
I've added an onPaste event to the Textfield to allow pasting images, instead of just dragging and dropping them. This is to help those who often take screenshots without saving them.


MohamedBassem:
Thanks @AhmadMuj and welcome back!

PR #232, opened 2024-06-18
fix: text paste bug introduced with image paste

mdsaban:
Fixes: Text paste not working due to the image paste feature introduced in #228


MohamedBassem (approved):
Thanks!

PR #235, opened 2024-06-18
Hide archived bookmarks on favourite List

mdsaban:
## What?
- Updated the favorite list functionality to exclude archived bookmarks.

## Why?
- Improved UX by ensuring that the favorite list only displays active, non-archived bookmarks.
- Maintained consistency with the intended behavior of archiving, where archived assets should only be visible on the dedicated archived page and not in other sections of the application.

### DIFF
--- a/apps/mobile/app/dashboard/favourites.tsx
+++ b/apps/mobile/app/dashboard/favourites.tsx
@@ -8,6 +8,7 @@ export default function Favourites() {
       <UpdatingBookmarkList
         query={{
           favourited: true,
+          archived: false 
         }}
         header={<PageTitle title="⭐️ Favourites" />}
       />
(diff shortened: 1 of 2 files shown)


MohamedBassem:
We have a different definition for archive I guess. Archive used to work like that before, but we changed the meaning in the `0.12` release (https://github.com/hoarder-app/hoarder/releases/tag/v0.12.0). Archive now means "remove from homepage", but it'll continue appearing elsewhere. This is aligned with other services, like "pocket" for example were favourited content still shows up in the favourites tab even if you archive it.

MohamedBassem:
The old version of "archive" wasn't very different from just deleting the bookmark, that's why we changed it.

PR #236, opened 2024-06-18
Changes to have more assets #215

kamtschatka:
The same as https://github.com/hoarder-app/hoarder/pull/220 but with extracted functions and automatic mapping from the database to the asset properties


kamtschatka:
closing this branch and will follow up only with the refactoring to reuse code better and make it easier to extend, but keep the way it currently works the same.

PR #240, opened 2024-06-20
Do not allow adding bookmark in archive list

mdsaban:
- There's no need to allow users to add bookmarks in the Archived list

image


MohamedBassem:
Similar explanation to the one in https://github.com/hoarder-app/hoarder/pull/235. I sometimes add stuff to archive list, if I've already read it and I don't want it to appear on my homepage.

PR #242, opened 2024-06-20
[UX improvement] Remove separate search page

mdsaban:
### What
- Removal of the separate search page, thus improving user experience (UX)
- UX alignment with competitor products

### Why
- Increase friction for the user by redirecting them to the `/search` page after searching from the homepage. When they clear the search query, they remain on the `search` page, preventing them from adding a new item without first navigating back to the `home` page.

## Open ended questions
- Discuss usecase for pagination

## Screenshot
image


apps/web/components/dashboard/search/SearchInput.tsx
```diff
@@ -9,11 +9,15 @@ const SearchInput = React.forwardRef<
   React.HTMLAttributes<HTMLInputElement> & { loading?: boolean }
 >(({ className, ...props }, ref) => {
   const { debounceSearch, searchQuery } = useDoBookmarkSearch();
+  const isSearchEnabled = process.env.MEILISEARCH_ENABLED;
```
MohamedBassem, line 12: Accessing `process.env` from a client component won't work. If you want it there, you need to add this setting to the client config in `shared/config.ts`, and access it via `useClientConfig()`.

apps/web/next.config.mjs
```diff
@@ -49,6 +49,9 @@ const nextConfig = withPWA({
       },
     ];
   },
+  env: {
+    MEILISEARCH_ENABLED: !!process.env.MEILI_ADDR,
+  },
```
MohamedBassem, line 54: This is a build time env variable so not exactly what we want. Also next now advices against using this in their docs (https://nextjs.org/docs/pages/api-reference/next-config-js/env). Let's use our custom config mechanism in `shared/config.ts` instead for consistency.

apps/web/app/dashboard/bookmarks/page.tsx
```diff
@@ -1,20 +1,40 @@
...
+
+import { Suspense, useRef } from "react";
+import BookmarksGrid from "@/components/dashboard/bookmarks/BookmarksGrid";
+import { SearchInput } from "@/components/dashboard/search/SearchInput";
+import { FullPageSpinner } from "@/components/ui/full-page-spinner";
 import { Separator } from "@/components/ui/separator";
+import { useBookmarkSearch } from "@/lib/hooks/bookmark-search";
+
+function BookmarksComp() {
+  const { data } = useBookmarkSearch();
```
MohamedBassem, line 11: Relying on the data returned from the search endpoint in the homepage is not a great idea. The search endpoint depends on the data being fetched from meillisearch. Also, if mellisearch is disabled, this endpoint won't return any data.

If you really want to get rid of the redirection to the search page, I'd make the distinction between empty query string (in which case you use the `Bookmarks` component) and a non-empty query string, in which case you use the search endpoint and render the `BookmarkGrid`.

apps/web/app/dashboard/search/page.tsx
```diff
@@ -1,35 +0,0 @@
-"use client";
```
MohamedBassem, line 1: The search bar doesn't exist in all pages (e.g. in favourites, etc). I'd still keep the search page tab for quick search access.

MohamedBassem (changes requested):

mdsaban:
@MohamedBassem I'm closing this PR for now. Will revisit this later.

PR #244, opened 2024-06-20
fix: Remove the bookmark update text limit

AhmadMuj:
I so no use of having a limit in the update only while the create have no such limit.


MohamedBassem:
Ah good catch!

PR #247, opened 2024-06-21
feature: Enhance the bookmark textarea in the list layout

AhmadMuj:
This PR contains 2 fixes: 
1- Allow resizing the textfield in the list layout only, it does not break the UI and it gives a better experince when writing notes because the area for writing in the list mode is so small
2- Auto resizing the textfield to fit the content to a max of half the screen height

### DIFF
--- a/apps/web/components/dashboard/bookmarks/EditorCard.tsx
+++ b/apps/web/components/dashboard/bookmarks/EditorCard.tsx
@@ -74,6 +74,9 @@ export default function EditorCard({ className }: { className?: string }) {
         });
       }
       form.reset();
+      if (inputRef?.current?.style) {
+        inputRef.current.style.height = "auto";
+      }
     },
     onError: (e) => {
       toast({ description: e.message, variant: "destructive" });
@@ -125,10 +128,6 @@ export default function EditorCard({ className }: { className?: string }) {
     } catch (e) {
       // Not a URL
       mutate({ type: "text", text });
-    } finally {
-      if (inputRef?.current?.style) {
-        inputRef.current.style.height = "auto";
-      }
     }
   };
 
--- a/apps/web/components/dashboard/bookmarks/EditorCard.tsx
+++ b/apps/web/components/dashboard/bookmarks/EditorCard.tsx
@@ -103,6 +103,21 @@ export default function EditorCard({ className }: { className?: string }) {
     setMultiUrlImportState({ urls, text });
   }
 
+  const onInput = (e: React.FormEvent<HTMLTextAreaElement>) => {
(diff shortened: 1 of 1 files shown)


apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -181,6 +188,20 @@ export default function EditorCard({ className }: { className?: string }) {
...
+                  const maxHeight = window.innerHeight * 0.5;
+                  target.style.height = "auto";
+
+                  if (target.scrollHeight <= maxHeight) {
+                    target.style.height = `${target.scrollHeight}px`;
+                  } else {
+                    target.style.height = `${maxHeight}px`;
+                  }
+                }
+              }}
```
mdsaban: Can be extracted as a separate method `handleInput`, instead of keeping it in the template

mdsaban (approved):

apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -163,7 +182,10 @@ export default function EditorCard({ className }: { className?: string }) {
             <Textarea
               ref={inputRef}
               disabled={isPending}
-              className="h-full w-full resize-none border-none text-lg focus-visible:ring-0"
+              className={cn(
+                "h-full w-full border-none text-lg focus-visible:ring-0",
+                { "resize-none": bookmarkLayout !== "list" },
```
MohamedBassem, line 191: I didn't know that you can conditionally add classes this way, TIL.

apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -99,6 +103,21 @@ export default function EditorCard({ className }: { className?: string }) {
...
+    if (bookmarkLayout === "list") {
+      const target = e.target as HTMLTextAreaElement;
+      const maxHeight = window.innerHeight * 0.5;
+      target.style.height = "auto";
+
+      if (target.scrollHeight <= maxHeight) {
+        target.style.height = `${target.scrollHeight}px`;
+      } else {
+        target.style.height = `${maxHeight}px`;
+      }
```
MohamedBassem, line 120: That's a neat trick

MohamedBassem (commented):
Overall, this looks good to me. I think the only caveat is that we need to reset the size after the form is submitted somehow. Not sure what's the best way to do this though.

AhmadMuj:
I think now it's good to go @MohamedBassem

MohamedBassem:
Looks amazing, thank you!

PR #252, opened 2024-06-23
Move layout control to main screen

mdsaban:
- Move layout control on top of bookmarks grid. 
- Currently, it’s hidden in the settings menu at the bottom left, and people sometimes don’t notice it

https://github.com/hoarder-app/hoarder/assets/45597394/9fa6a7da-d55a-4caa-b87e-b4543e6affaa


MohamedBassem (approved):
Thank you for quickly working on this. It's much better than what we had before!

PR #253, opened 2024-06-23
refactoring asset types

kamtschatka:
Extracted out functions to silently delete assets and to update them after crawling
Generalized the mapping of assets to bookmark fields to make extending them easier

### DIFF
--- a/apps/workers/crawlerWorker.ts
+++ b/apps/workers/crawlerWorker.ts
@@ -5,9 +5,11 @@ import type { Job } from "bullmq";
 import type { Browser } from "puppeteer";
 import { Readability } from "@mozilla/readability";
 import { Mutex } from "async-mutex";
+import Database from "better-sqlite3";
 import { Worker } from "bullmq";
 import DOMPurify from "dompurify";
-import { eq } from "drizzle-orm";
+import { eq, ExtractTablesWithRelations } from "drizzle-orm";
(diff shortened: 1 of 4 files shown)


packages/shared/assetdb.ts
```diff
@@ -117,6 +117,20 @@ export async function readAsset({
...
 }
 
+/**
+ * Deletes the passed in asset if it exists and ignores any errors
+ * @param userId the id of the user the asset belongs to
+ * @param assetId the id of the asset to delete
+ */
+export async function silentDeleteAsset(
+  userId: string,
+  assetId: string | undefined,
```
MohamedBassem: Tbh, not convinced that an API in `assetdb` that's called `silentDeleteAsset` accepts an optional assetId. If we really care about this, I'd move it as a helper function in crawlerWorker. Not sure it's worth it.

apps/workers/crawlerWorker.ts
```diff
@@ -5,9 +5,11 @@ import type { Job } from "bullmq";
 import type { Browser } from "puppeteer";
 import { Readability } from "@mozilla/readability";
 import { Mutex } from "async-mutex";
+import Database from "better-sqlite3";
```
MohamedBassem: I don't think we should leak `better-sqlite3` here. I'll re-export the type that we need from `packages/db`.

PR #259, opened 2024-06-26
[Feat] Add bulk edit option for bookmarks

mdsaban:
Added bulk edit option. I have added 3 options for now, will extend this to tags and lists in a separate PR
- Bulk Favourite/Un-Favourite
- Bulk Archive/Un-Archive
- Bulk delete

https://github.com/hoarder-app/hoarder/assets/45597394/d8625dfe-dcb6-4b1f-91a6-8dbe289cebdb

### DIFF
--- a/apps/web/components/dashboard/BulkBookmarksAction.tsx
+++ b/apps/web/components/dashboard/BulkBookmarksAction.tsx
@@ -6,25 +6,24 @@ import {
   ActionButtonWithTooltip,
 } from "@/components/ui/action-button";
 import ActionConfirmingDialog from "@/components/ui/action-confirming-dialog";
-import { ButtonWithTooltip } from "@/components/ui/button";
 import { useToast } from "@/components/ui/use-toast";
-import bulkActions from "@/store/useBulkBookmarksAction";
-import { Archive, ArchiveRestore, Pencil, Star, Trash2, X } from "lucide-react";
-import { useTheme } from "next-themes";
+import useBulkActionsStore from "@/lib/bulkActions";
+import { Pencil, Trash2, X } from "lucide-react";
 
 import {
   useDeleteBookmark,
   useUpdateBookmark,
 } from "@hoarder/shared-react/hooks/bookmarks";
 
+import { ArchivedActionIcon, FavouritedActionIcon } from "./bookmarks/icons";
+
 export default function BulkBookmarksAction() {
-  const { theme } = useTheme();
-  const { selectedBookmarks, isBulkEditEnabled } = bulkActions();
-  const setIsBulkEditEnabled = bulkActions(
+  const { selectedBookmarks, isBulkEditEnabled } = useBulkActionsStore();
+  const setIsBulkEditEnabled = useBulkActionsStore(
     (state) => state.setIsBulkEditEnabled,
   );
   const { toast } = useToast();
-  const [isButtonLoading, setIsButtonLoading] = useState(false);
+  const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
 
   useEffect(() => {
     setIsBulkEditEnabled(false); // turn off toggle + clear selected bookmarks on mount
@@ -52,65 +51,38 @@ export default function BulkBookmarksAction() {
     onError,
   });
 
-  const DeleteButton = () => {
-    return (
-      <ActionConfirmingDialog
-        title={"Delete Bookmarks"}
-        description={<p>Are you sure you want to delete these bookmarks?</p>}
-        actionButton={() => (
-          <ActionButton
-            type="button"
-            variant="destructive"
-            loading={isButtonLoading}
-            onClick={() => deleteBookmarks()}
-          >
-            Delete
-          </ActionButton>
-        )}
-      >
-        <ButtonWithTooltip
-          tooltip="Delete"
-          disabled={!selectedBookmarks.length}
-          variant="ghost"
-        >
-          <Trash2 size={18} color="red" />
-        </ButtonWithTooltip>
-      </ActionConfirmingDialog>
-    );
-  };
-
   interface UpdateBookmarkProps {
     favourited?: boolean;
     archived?: boolean;
   }
 
-  const updateBookmarks = ({
+  const updateBookmarks = async ({
     favourited,
     archived,
-  }: UpdateBookmarkProps): void => {
-    setIsButtonLoading(true);
-    selectedBookmarks.map((item) => {
-      updateBookmarkMutator.mutate({
-        bookmarkId: item.id,
-        favourited: favourited ?? item.favourited,
-        archived: archived ?? item.archived,
-      });
-    });
+  }: UpdateBookmarkProps) => {
+    await Promise.all(
+      selectedBookmarks.map((item) =>
+        updateBookmarkMutator.mutateAsync({
+          bookmarkId: item.id,
+          favourited,
+          archived,
+        }),
+      ),
+    );
     toast({
       description: `${selectedBookmarks.length} bookmarks have been updated!`,
     });
-    setIsButtonLoading(false);
   };
 
-  const deleteBookmarks = () => {
-    setIsButtonLoading(true);
-    selectedBookmarks.map((item) => {
-      deleteBookmarkMutator.mutate({ bookmarkId: item.id });
-    });
+  const deleteBookmarks = async () => {
+    await Promise.all(
+      selectedBookmarks.map((item) =>
+        deleteBookmarkMutator.mutateAsync({ bookmarkId: item.id }),
+      ),
+    );
     toast({
       description: `${selectedBookmarks.length} bookmarks have been deleted!`,
     });
-    setIsButtonLoading(false);
   };
 
   const alreadyFavourited =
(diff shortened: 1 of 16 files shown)


apps/web/components/dashboard/BulkBookmarksAction.tsx
```diff
@@ -0,0 +1,204 @@
...
+  const [showDeleteModal, setShowDeleteModal] = useState(false);
+
+  useEffect(() => {
+    handleBulkEdit(false); // turn off toggle + clear selected bookmarks on mount
+  }, []);
+
+  const onError = () => {
+    toast({
+      variant: "destructive",
+      title: "Something went wrong",
```
kamtschatka, line 35: this has been remarked before: People want to know more information. Can we provide information about the affected bookmark somehow?

apps/web/components/dashboard/GlobalActions.tsx
```diff
@@ -0,0 +1,22 @@
...
+import bulkActions from "@/store/bulkBookmarksAction";
+
+export default function GlobalActions() {
+  const { isBulkEditEnabled } = bulkActions();
+
+  return (
+    <div
+      className={cn(
+        "flex rounded-md border bg-background transition-all",
+        isBulkEditEnabled ? "w-[251px]" : "w-[102px]",
```
MohamedBassem: Why are we explicitly setting the width here? Why can't it flex depending on the components inside the global actions?

apps/web/components/dashboard/BulkBookmarksAction.tsx
```diff
@@ -0,0 +1,204 @@
...
+      {showDeleteModal && (
+        <DeleteModal
+          open={showDeleteModal}
+          setOpen={setShowDeleteModal}
+          deleteBookmarks={deleteBookmarks}
+        />
+      )}
+      <div className="flex">
+        {actionList.map(({ name, icon: Icon, action }) => (
+          <ButtonWithTooltip
```
MohamedBassem: We should probably disable the buttons if nothing is selected.

apps/web/components/dashboard/bookmarks/BookmarkLayoutAdaptingCard.tsx
```diff
@@ -45,6 +47,41 @@ function BottomRow({
...
+          "bg-opacity-10": isSelected,
+        },
+      )}
+      onClick={() => toggleBookmark(bookmark)}
+    >
+      <button className="absolute right-2 top-2 z-50 opacity-100">
+        <div
+          className={cn(
+            "flex h-4 w-4 items-center justify-center rounded-full border border-dashed border-gray-600",
+            isSelected ? "bg-black" : "bg-white bg-opacity-40",
```
MohamedBassem: This effect is not as obvious in dark mode

apps/web/components/dashboard/BulkBookmarksAction.tsx
```diff
@@ -0,0 +1,204 @@
...
+      selectedBookmarks.every((item) => item.archived === true)) ||
+    pathname.includes("archive");
+
+  const actionList = [
+    {
+      name: alreadyFavourited ? "Unfavourite" : "Favourite",
+      icon: (
+        <Star
+          color={alreadyFavourited ? "#ebb434" : "#000"}
+          fill={alreadyFavourited ? "#ebb434" : "transparent"}
```
MohamedBassem: The star doesn't appear at all in dark mode.

apps/web/store/bulkBookmarksAction.tsx
```diff
@@ -0,0 +1,44 @@
+// reference article https://refine.dev/blog/zustand-react-state/#build-a-to-do-app-using-zustand
+import { create } from "zustand";
+
+import type { ZBookmark } from "@hoarder/shared/types/bookmarks";
+
+interface BookmarkState {
+  selectedBookmarks: ZBookmark[];
+  isBulkEditEnabled: boolean;
+  handleBulkEdit: (isEnabled: boolean) => void;
```
MohamedBassem: This is a small nitpicking, but it wasn't clear to me what `handle` here means. Maybe `setIsBulkEditEnabled`?

apps/web/store/bulkBookmarksAction.tsx
```diff
@@ -0,0 +1,44 @@
...
+  toggleBookmark: (bookmark: ZBookmark) => void;
+  getSelectedBookmarks: () => ZBookmark[];
```
MohamedBassem: What's the point of this getter? Why can't we just use `selectedBookmarks` directly?

apps/web/components/dashboard/BulkBookmarksAction.tsx
```diff
@@ -0,0 +1,204 @@
...
+  open,
+  setOpen,
+  deleteBookmarks,
+}: {
+  open: boolean;
+  setOpen: (open: boolean) => void;
+  deleteBookmarks: () => void;
+}) {
+  return (
+    <Dialog open={open} onOpenChange={setOpen}>
```
MohamedBassem: Sounds like you can use the `ActionConfirmingDialog` here.

apps/web/components/dashboard/BulkBookmarksAction.tsx
MohamedBassem: Also, you probably want to differentiate between buttons that does actions and have loading states, and buttons that just close bulk edit. For the ones that do actions, you probably want to use `ActionButtonWithTooltip` so that they are disabled in demo mode.

apps/web/store/bulkBookmarksAction.tsx
```diff
@@ -0,0 +1,44 @@
...
+}
+
+const selectedBookmarksStore = create<BookmarkState>((set, get) => ({
```
MohamedBassem: Looking at zustand's documentations, those stores are meant to be `hooks`. So they should start with the `use` prefix, and similarly when you import them the import name should have `use` as a prefix.

apps/web/components/dashboard/BulkBookmarksAction.tsx
```diff
@@ -0,0 +1,204 @@
...
+    });
+    toast({
+      description: `${selectedBookmarks.length} bookmarks have been deleted!`,
+    });
+  };
+
+  const alreadyFavourited =
+    (selectedBookmarks.length &&
+      selectedBookmarks.every((item) => item.favourited === true)) ||
+    pathname.includes("favourites");
```
MohamedBassem: Is this check really needed? Isn't it kinda redundant? I don't like the path matching here tbh, kinda fragile.

apps/web/components/dashboard/BulkBookmarksAction.tsx
```diff
@@ -0,0 +1,204 @@
...
+
+  const alreadyArchived =
+    (selectedBookmarks.length &&
+      selectedBookmarks.every((item) => item.archived === true)) ||
+    pathname.includes("archive");
```
MohamedBassem: Same here about the path matching.

MohamedBassem (changes requested):
Played with it a bit, and it's amazing! Great job really. Left a bunch of small comments.

apps/web/components/dashboard/GlobalActions.tsx
mdsaban: Handled it dynamically in a bettter way. Basically to animate the width we need initial and final width to do transition.

apps/web/components/dashboard/BulkBookmarksAction.tsx
mdsaban: - Replace with `ActionButtonWithTooltip `
- Added `disable` when nothing is selected

apps/web/components/dashboard/bookmarks/BookmarkLayoutAdaptingCard.tsx
mdsaban: Thanks for testing in dark mode, fixed it.

apps/web/components/dashboard/BulkBookmarksAction.tsx
mdsaban: Fixed

apps/web/store/bulkBookmarksAction.tsx
mdsaban: Removed `getSelectedBookmarks ` as we were not using it

apps/web/store/bulkBookmarksAction.tsx
mdsaban: Fixed, Renamed store name

apps/web/components/dashboard/BulkBookmarksAction.tsx
mdsaban: - I added this check if you're on a list and you select multiple items that are already favourite/archived, you can un-favourite/un-archive at once. Good to have. 
- Agree with path checking. What do you suggest? How should we handle the case where we want to show 'un-favorite' and 'un-archive' icons on the Favorites and Archive pages?

apps/web/store/bulkBookmarksAction.tsx
mdsaban: Renamed

apps/web/components/dashboard/BulkBookmarksAction.tsx
mdsaban: Resolved

apps/web/components/dashboard/BulkBookmarksAction.tsx
MohamedBassem: If the selected bookmark are in favorites wouldn't that match the condition before the path check anyways?

apps/web/components/dashboard/BulkBookmarksAction.tsx
mdsaban: Let's pick this in separate PR
cc - @MohamedBassem

PR #265, opened 2024-06-30
Allow downloading more content from a webpage and index it #215

kamtschatka:
Added a worker that allows downloading videos depending on the environment variables
refactored the code a bit
added new video asset
updated documentation

Depends on https://github.com/hoarder-app/hoarder/pull/256


kamtschatka:
declining for now, to add a state to the database and allow recrawling from the admin page

PR #269, opened 2024-07-01
fixed the editorcard textarea

kamtschatka:
fixed an issue that would half the size of the textarea in grid/masonry layout after submitting a bookmark
additionally it would resize the bookmark input field after each input on the textarea

### DIFF
--- a/apps/web/components/dashboard/bookmarks/EditorCard.tsx
+++ b/apps/web/components/dashboard/bookmarks/EditorCard.tsx
@@ -33,6 +33,7 @@ function useFocusOnKeyPress(inputRef: React.RefObject<HTMLTextAreaElement>) {
         inputRef.current.focus();
       }
     }
+
     document.addEventListener("keydown", handleKeyPress);
     return () => {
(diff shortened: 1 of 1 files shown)


apps/web/components/dashboard/bookmarks/EditorCard.tsx
```diff
@@ -107,21 +108,6 @@ export default function EditorCard({ className }: { className?: string }) {
     setMultiUrlImportState({ urls, text });
   }
 
-  const onInput = (e: React.FormEvent<HTMLTextAreaElement>) => {
```
MohamedBassem: Seems like you're basically reverting (https://github.com/hoarder-app/hoarder/pull/247)? Is that intentional?

MohamedBassem (changes requested):

apps/web/components/dashboard/bookmarks/EditorCard.tsx
kamtschatka: ok, misunderstood what resize was meant. reverted that part

PR #272, opened 2024-07-04
fix(web): Add sans-serif fallback for Inter font in CJK character sets

kou029w:
Thank you for creating such awesome software.

This PR addresses the issue of Inter fonts not loading for CJK (Chinese, Japanese, and Korean) character sets, resulting in an inconsistent appearance. To improve consistency of appearance, a sans-serif fallback is specified. This ensures that if the Inter font fails to load, the generic sans-serif font will be used to maintain a consistent look across different character sets.

Original:
Screenshot 2024-07-04 at 11-08-18 Hoarder

This PR:
Screenshot 2024-07-04 at 11-08-44 Hoarder

Input text: "日本語表示確認用サンプル文"

### DIFF
--- a/apps/web/app/layout.tsx
+++ b/apps/web/app/layout.tsx
@@ -18,7 +18,10 @@ import { ReactQueryDevtools } from "@tanstack/react-query-devtools";
 
 import { clientConfig } from "@hoarder/shared/config";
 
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thanks for the PR!

Issue #155, opened 2024-05-16
Add context menu item in the browser extension


lucius346346:
Are there any chances of implementing this? It would be helpful to quickly add links without visiting them.

kamtschatka:
You mean so that you simply rightclick on a link, press "add to hoarder" and that's it and tag and list configuration will then be done later?

MohamedBassem:
@lucius346346 Yeah, I agree with you. I think this is a needed feature indeed. Though it's a bit non-trivial to implement because chrome doesn't allow you to trigger the "plugin" popup window when the context menu is clicked. However, I agree with you, I probably need to spend the effort and just do it :) Other services (e.g. pocket) avoid that by not using a popup window at all and instead injecting the plugin dialog via javascript inside the page itself.

MohamedBassem:
@kamtschatka The idea here is that when you right click and press "add to hoarder" the plugin popup would still appear and allow you to edit the tags and lists.

kamtschatka:
I have opened a PR for it. With the current stable chrome it is not possible to open the UI programmatically, but with Chrome 127 (rollout start is around 17th of July), this will be possible.

PR #278, opened 2024-07-06
Add context menu item in the browser extension #155

kamtschatka:
Added a context menu entry to add links directly to hoarder


kamtschatka:
Note: Does not work with Chrome 126 (current stable release at the moment). 
Will start working with Chrome 127 (planned release is ~17th of July). I ran the tests with the beta channel.

MohamedBassem:
Now that chrome 127 is released, we can merge this! I tried it and it's indeed really cool.

PR #282, opened 2024-07-07
Admin settings config

kamtschatka:
just showing the wireframe


packages/db/config/configUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+      .select()
+      .from(config)
+      .where(eq(config.key, configValue.key));
+    if (rows.length === 0) {
+      return void 0;
+    }
+    return parseValue(configValue, rows[0].value);
+  } catch (e) {
+    return void 0;
+  }
```
MohamedBassem: Returning undefined on sqlite error here is dangerous. I'd rather let the server crash than use incorrect value. So let's remove the try-catch here.

packages/db/config/configUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+}
+
+/**
+ * @returns the value from the environment variable. Undefined if it does not exist.
+ */
+export function getConfigValueFromEnv<T extends ConfigType>(
+  configValue: ConfigValue<T>,
+): ConfigTypeMap[T] | undefined {
+  const environmentValue = process.env[configValue.key];
+  if (!environmentValue) {
```
MohamedBassem: You should check explicitly for undefined here. Empty strings can be a value env variable value.
```suggestion
  if (environmentValue === undefined) {

```

packages/db/config/configUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+      await db
+        .update(config)
+        .set({ value })
+        .where(eq(config.key, configValue.key))
+        .execute();
+    } else {
+      // If the record doesn't exist, insert it
+      await db.insert(config).values({ key: configValue.key, value }).execute();
+    }
+  });
```
MohamedBassem: You don't need a transaction. Sqlite has upserts (https://orm.drizzle.team/docs/insert#on-conflict-do-update).

packages/db/config/configUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+
+function parseValue<T extends ConfigType>(
+  configValue: ConfigValue<T>,
+  value: string,
+): ConfigTypeMap[T] | undefined {
+  const parsed = configValue.validator.safeParse(value);
+  if (parsed.success) {
+    return parsed.data as ConfigTypeMap[T];
+  }
+  return undefined;
```
MohamedBassem: nitpicking: I'd indent the error path.
```suggestion
  if (!parsed.success) {
    return undefined;
  }
  return parsed.data as ConfigTypeMap[T];

```

packages/db/config/config.ts
```diff
@@ -0,0 +1,151 @@
...
+      key: "INFERENCE_PROVIDER",
+      name: "Inference Provider",
+      type: ConfigType.INFERENCE_PROVIDER_ENUM,
+      defaultValue: InferenceProviderEnum.DISABLED,
+      validator: InferenceProviderEnumValidator,
+    }),
+    openApiKey: new ConfigValue({
+      key: "OPENAI_API_KEY",
+      name: "OpenAPI Key",
+      type: ConfigType.STRING,
```
MohamedBassem: We probalby need `ConfigType.PASSWORD` to hide the openai string in the admin UI with a password field.

packages/db/config/configValue.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export const InferenceProviderEnumValidator = z.enum([
+  InferenceProviderEnum.DISABLED,
+  InferenceProviderEnum.OPEN_AI,
+  InferenceProviderEnum.OLLAMA,
+]);
+
+export type InferenceProviderEnumZodType =
+  typeof InferenceProviderEnumValidator;
```
MohamedBassem: This should be:
```suggestion
  z.infer<typeof InferenceProviderEnumValidator>;

```

packages/db/config/configValue.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export interface ConfigProperties<T extends ConfigType> {
+  key: string;
+  name: string;
+  type: T;
+  defaultValue: ConfigTypeMap[T];
+  validator: ConfigValidatorMap[T];
+}
+
+export class ConfigValue<T extends ConfigType> {
```
MohamedBassem, line 41: Why not `ConfigValue extends ConfigProperties<T extends ConfigType>` to remove the duplicate properties?

packages/db/config/configUtils.ts
```diff
@@ -0,0 +1,91 @@
...
+  const dbValue = await getConfigValueFromDB(configValue);
+  if (dbValue) {
+    return dbValue;
+  }
+  const envValue = getConfigValueFromEnv(configValue);
+  if (envValue) {
+    return envValue;
+  }
+  return configValue.defaultValue;
+}
```
MohamedBassem: Can this be a method on `ConfigValue`?

MohamedBassem (changes requested):

packages/db/config/configUtils.ts
kamtschatka: no, as this has a dependency on the database, which will not work with "use client".
That's why I had to move it out.
Unless there is some magic I don't know about?

packages/db/config/configValue.ts
kamtschatka: I don't think so, i am not using infer on any other of the types there either, it is just an alias at this point.

packages/db/config/configValue.ts
kamtschatka: I am not sure I understand. ConfigProperties is an interface, so i can't use extends if I want a class from it.
I can use implements, but that is not shorter:

```
export class ConfigValue<T extends ConfigType> implements ConfigProperties<T> {
  key: string;
  name: string;
  type: T;
  defaultValue: ConfigTypeMap[T];
  validator: ConfigValidatorMap[T];
  value: ConfigTypeMap[T] | undefined;

  constructor(props: ConfigProperties<T>) {
    this.key = props.key;
    this.name = props.name;
    this.type = props.type;
    this.defaultValue = props.defaultValue;
    this.validator = props.validator;
  }

  setValue(value: ConfigTypeMap[T]): void {
    this.value = value;
  }
}
```

Issue #281, opened 2024-07-07
last version bookmarks import error 


nsk:
hi ,
on the latest pull version, still impossible to import a list of favorites in drag and drop mode from chrome, it says that something is going wrong?

![chrome_7uBi5VRQ5r](https://github.com/hoarder-app/hoarder/assets/1135867/70abf0e0-b148-4815-91a5-c6c3b178e0d9)

kamtschatka:
Please post the logs, the screenshot alone does not help.

nsk:
@kamtschatka I have nothing in the logs

kamtschatka:
There is either something in the logs of the "web" container or in the devtools. 
alternatively you can also provide the file you want to import (i am also in discord, if you do not want to share it publicly)

carduswork:
got the same error,here is the log in chrome,maybe helpful
![image](https://github.com/hoarder-app/hoarder/assets/17943894/639fb0ac-c1b4-4e02-b9c4-c375bbe71588)
![image](https://github.com/hoarder-app/hoarder/assets/17943894/10194280-8068-4f48-82c0-bffb06a9c766)

kamtschatka:
I just tried it with my installation and everything works fine, so must have to do with your state/bookmark file.
What happens? Does it not import anything at all, or are some of them skipped?
Can you provide the bookmarks file?

carduswork:
I exported my bookmark file from a platform like raindrop,it works fine after i changed my bookmark file to the file exported from my chrom,thanks!

> I just tried it with my installation and everything works fine, so must have to do with your state/bookmark file. What happens? Does it not import anything at all, or are some of them skipped? Can you provide the bookmarks file?

kamtschatka:
from a platform like raindrop, or raindrop? 
I would like to try that out when I am back home. There is a safeguard against importing completely random bookmark files, but maybe there is a slight difference between the browser exported format and whatever you tried.

nsk:
hi I did the conversion to .txt via the CLI do you think it would be possible to directly import the all-linx.txt directly into the interface without going through the CLI, I tried the other way to do it too but it doesn't work

kamtschatka:
At some point you'll have to provide the file or some instructions for me to reproduce what you are doing, otherwise whis discussion will go nowhere

nsk:
I have a raindrop account I'm trying this way I'll come back and tell you if it works later

nsk:
[...]

PR #291, opened 2024-07-09
last version bookmarks import error #281

kamtschatka:
limiting the length of the URL for batched requests to prevent 431 errors

### DIFF
--- a/apps/cli/src/lib/trpc.ts
+++ b/apps/cli/src/lib/trpc.ts
@@ -10,6 +10,7 @@ export function getAPIClient() {
     links: [
       httpBatchLink({
         url: `${globals.serverAddr}/api/trpc`,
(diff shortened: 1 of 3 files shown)


MohamedBassem:
How did you come up with the number btw? A quick google search suggests that chrome has a max URL length of 2048?

kamtschatka:
trial and error. used postman to make a request until it no longer failed. tried with chrome and then tried the import and it worked fine

kamtschatka:
fyi, i just checked, max url length in chrome is 2mb, so we should be good: https://chromium.googlesource.com/chromium/src/+/master/docs/security/url_display_guidelines/url_display_guidelines.md#:~:text=Chrome%20limits%20URLs%20to%20a,is%20used%20on%20VR%20platforms.

PR #308, opened 2024-07-14
chore: Add a new sqlite based queue package

MohamedBassem:
chore: Add a new sqlite based queue package

### DIFF
--- a/packages/queue/db.ts
+++ b/packages/queue/db.ts
@@ -1,0 +1,19 @@
+import path from "node:path";
+import Database from "better-sqlite3";
+import { BetterSQLite3Database, drizzle } from "drizzle-orm/better-sqlite3";
+import { migrate } from "drizzle-orm/better-sqlite3/migrator";
+
+import * as schema from "./schema";
+
+export function buildDBClient(dbPath: string, runMigrations = false) {
+  const sqlite = new Database(dbPath);
+  const db = drizzle(sqlite, { schema });
+  if (runMigrations) {
+    migrateDB(db);
+  }
+  return db;
+}
+
+export function migrateDB(db: BetterSQLite3Database<typeof schema>) {
+  migrate(db, { migrationsFolder: path.join(__dirname, "drizzle") });
+}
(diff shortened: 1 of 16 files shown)


kamtschatka (commented):
code looks good, I am wondering why it is a separate db though?
I was thinking to add searching capability for all kinds of states, which is harder if we have 2 database connections.
Right now, if you go to the admin panel, you see a lot of numbers regarding bookmarks waiting for worker runs/inference runs, but that is all.
I would love to be able to click on it (or create a filter) to see all the bookmarks that are pending/had issues/whatever, so you can reschedule them.

packages/queue/drizzle/0000_blue_weapon_omega.sql
```diff
@@ -0,0 +1,18 @@
+CREATE TABLE `tasks` (
```
kamtschatka: if we had the bookmark id as a field, we could do queries on them, without having to look into the payload.
To my knowledge, everything done in the worker is currently always related to a bookmark?

MohamedBassem:
> why it is a separate db though?

Mostly for isolation purposes. I didn't want to expose hoarder to too much knowledge about what happens in the queue layer. The queues are mostly separate state that's not related to the data stored in hoarder, so I didn't see a reason to tie them to the main database. They are also kinda ephemeral. I'm planning to eventually extract it into a separate package, and having that package own the database makes things simpler.

> I would love to be able to click on it (or create a filter) to see all the bookmarks that are pending/had issues/whatever, so you can reschedule them.

The state is already reflected in the database (`bookmarks.taggingStatus` and `bookmarkLinks.crawlStatus`), so you can already do this without introspecting the queue status.

packages/queue/drizzle/0000_blue_weapon_omega.sql
MohamedBassem: Currently, yes. But it doesn't have to stay like that. I already have plans to include a worker to periodically reap orphaned assets for example. So I want to keep the queues generic.

kamtschatka: we can have that as a nullable field then?

Issue #295, opened 2024-07-11
[Feature request] Allow Epub generation from multiple bookmarks


winston0410:
A feature I found very useful but missing in hoarder right now is Epub generation from multiple bookmarks, this will allow a seamless reading experience offline, and we can create our own books from multiple websites.

This feature is supported by some other opensource bookmark manager, like https://readeck.org/en/.

kamtschatka:
I have opened a PR to get the discussion rolling.
Currently it only shows a link to download a single bookmark, but the code is already prepared to generate multiple bookmarks into a single file. 
I'll start a discussion in discord to see how we are going to continue.

jalberto:
is this feature still alive?

kijekijikokwe:
Would love to see this feature. It's the one, simple feature that's letting smaller apps like readeck make us look bad.

Igglybuff:
This would be awesome. Readeck and Wallabag can do this.

Being able to send saved articles/blog posts to KOReader from Karakeep is the one feature I'm missing

AlgusDark:
I hope this feature can be done before I implement it on my KOReader plugin as it will make things nicer :)

filviu:
I just ran into Readeck as well (checking out the competition :) ) and was about to open a feature request beacause I missed this one. So I'll write what I was searching for so future searches land here.

What I found very nice was the fact that Readeck allows downloading collections (I'm guessing for us that would be lists) as one epub and also provides an OPDS server - that would be the killer feature for me - to link koreader on my eink android device.

PR #312, opened 2024-07-16
Allow Epub generation from multiple bookmarks #295

kamtschatka:
Added a library to generate epub files
added a button to create epubs for single link bookmarks


apps/web/app/api/epub/route.ts
```diff
@@ -0,0 +1,65 @@
...
+    });
+  }
+
+  const bookmarkInformation = await db
+    .select({
+      title: bookmarkLinks.title,
+      htmlContent: bookmarkLinks.htmlContent,
+    })
+    .from(bookmarkLinks)
+    .where(inArray(bookmarkLinks.id, assetIds));
```
MohamedBassem: This is missing privacy checks and would allow anyone to download bookmarks they don't own

apps/web/app/api/epub/route.ts
```diff
@@ -0,0 +1,65 @@
...
+  const assetIds = searchParams.getAll("assetId");
+
+  if (!assetIds || assetIds.length === 0) {
+    return new Response("", {
+      status: 404,
+    });
+  }
+
+  const bookmarkInformation = await db
+    .select({
```
MohamedBassem, line 26: For simple queries like this, prefer `db.query.bookmarkLinks`

apps/web/app/api/epub/route.ts
```diff
@@ -0,0 +1,65 @@
...
+  const chapters = bookmarkInformation.map((information) => {
+    return {
+      content: information.htmlContent ?? "",
+      title: information.title ?? "",
+    };
+  });
+
+  const title = getTitle(chapters);
+
+  const generatedEpub = await epub(title, chapters);
```
MohamedBassem: This crashes if it fails to download some links, for example in this link (https://www.da.vidbuchanan.co.uk/blog/r1-jailbreak.html)

We might want to pass `ignoreFailedDownloads: true`.

MohamedBassem: I think we should also remove the table of contents when it's only a single bookmark

apps/web/package.json
```diff
@@ -47,6 +47,7 @@
     "clsx": "^2.1.0",
     "dayjs": "^1.11.10",
     "drizzle-orm": "^0.29.4",
+    "epub-gen-memory": "^1.0.10",
```
MohamedBassem: This library seems unmaintained with last version being from 2 years ago. I wouldn't want to add this as a dependency to be honest. Is there a better library to use for the epub generation?

It's also worth noting that this library "fetches" links (e.g the images in a URL) from the `web` container which is something we currently never do. Not sure how I feel about this. It might be better if we do this from the user's browser instead given that we have all the info there anyways.

MohamedBassem (changes requested):

apps/web/package.json
kamtschatka: If we do it from the users browsers, we will have CORS issues. I created a chrome extension for games workshop before, where this was a huge issue and it was basically missing all the images all the time. If we don't include all the images, this will be the first bug report we have, that there are not all images included.

Regarding the library, there is https://github.com/cyrilis/epub-gen, which seems quite popular, but has not been updated for 6 years. There is a fork https://github.com/lesjoursfr/html-to-epub, which is still maintained. If you prefer this one, I can try using it instead.

kamtschatka: There is also a fork from the current repo: https://www.npmjs.com/package/@epubkit/epub-gen-memory

apps/web/app/api/epub/route.ts
kamtschatka: Seems like this is actually a requirement for epubs: https://ebooks.stackexchange.com/questions/7209/epub-without-a-table-of-contents-is-it-possible
I had a look at the library and it is hardcoded to generate it. I then tried an empty template and it does not work.

apps/web/app/api/epub/route.ts
kamtschatka: I guess with the join this is obsolete?

apps/web/package.json
kamtschatka: I have switched to the epubkit generation, as this generates in memory instead of writing to the disk and it is maintained and used in the tool https://epubkit.app/

apps/web/app/api/epub/route.ts
kamtschatka: was able to fix that by updating the template

PR #324, opened 2024-07-24
Add Kubernetes installation with Kustomize

j-fuentes:
This PR adds sample manifests for a Kubernetes installation using Kustomize.

I tried to reduce it to the basics, aiming to make it as similar as possible to the docker-compose one.

It should be pretty easy to extend to specific environments.


docs/docs/02-Installation/04-kubernetes.md
```diff
@@ -0,0 +1,71 @@
+# Kubernetes
+
+### Requirements
+
+- A kubernetes cluster with at least a amd64 node.
```
MohamedBassem: I'm curious, why do you need a single amd64? Hoarder (and all of its dependencies) work on arm as well.

kubernetes/amd64-patch.yaml
```diff
@@ -0,0 +1,50 @@
+apiVersion: apps/v1
```
MohamedBassem: Same question as above. Why do we need to pin all the containers to amd64 nodes?

MohamedBassem (approved):
Wow, thank you so much!

Looks good to me! We can merge after addressing the amd64 discussion.

Note: This can get much simpler after the next hoarder release. We'll get rid of the `redis` dependency, and we're doing an All-in-One container to get rid of the workers container as well.

docs/docs/02-Installation/04-kubernetes.md
j-fuentes: My cluster has a mix of amd64 and arm64 nodes. When I tried it, I was having issues with the chromium container. It could run but there was an error. I did not copy the error but I tried to run it on amd64 and everything worked.

So I have been a bit lazy not getting to the button of the issues.

But if everything is supposed to be arm64 compatible we can remove that restriction. I will update the PR.

kubernetes/amd64-patch.yaml
j-fuentes: answered above.

docs/docs/02-Installation/04-kubernetes.md
j-fuentes: Done in [1b7ebf7](https://github.com/hoarder-app/hoarder/pull/324/commits/1b7ebf70a8272ec7f10838a72e9a1d4d860bbcf4)

MohamedBassem:
Thank you!

PR #374, opened 2024-09-06
Clear disabled signups message

nghduc97:
The current behavior to disable only the button when `DISABLE_SIGNUPS=true` is really easy to misunderstood https://github.com/hoarder-app/hoarder/issues/52

### DIFF
--- a/apps/web/components/signin/CredentialsForm.tsx
+++ b/apps/web/components/signin/CredentialsForm.tsx
@@ -204,10 +204,7 @@ function SignUp() {
               );
             }}
           />
-          <ActionButton
-            type="submit"
-            loading={form.formState.isSubmitting}
-          >
+          <ActionButton type="submit" loading={form.formState.isSubmitting}>
             Sign Up
           </ActionButton>
         </div>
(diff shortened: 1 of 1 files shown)


apps/web/components/signin/CredentialsForm.tsx
```diff
@@ -208,7 +207,6 @@ function SignUp() {
           <ActionButton
             type="submit"
             loading={form.formState.isSubmitting}
-            disabled={clientConfig.auth.disableSignups}
           >
             Sign Up
```
MohamedBassem, line 208: How about we just do the change in the text directly here?:
```suggestion
            {clientConfig.auth.disableSignups ? "Signups are currently disabled" : "Sign Up"}
```

MohamedBassem (changes requested):
Hey, sorry for the delay in review. The change itself looks ok to me, just requested a simpler way to achieve the same thing.

apps/web/components/signin/CredentialsForm.tsx
nghduc97: Hiding the whole signup form or tab would be much clearer, I don't think we need the form shown if it serve no purposes other than confusing users.

kamtschatka: I agree, this also confused me when i had that set for testing purposes. If signup is disabled, I don't expect to see a signup page

MohamedBassem: Ok, in that case I don't mind.

MohamedBassem:
Thanks a lot @nghduc97 and welcome to the contributor lists!

Issue #387, opened 2024-09-11
Removing referrer header when clicking on links


TechClusterHQ:
It would be great if there was an option to remove the "referrer" header when clicking on bookmarks. Currently, you can be traced back to your instance of hoarder using the header. An option to add a prefix url (like https://anonym.es) to all urls (only when viewing them on the homepage) would be a good solution for this problem.

kamtschatka:
how about adding rel="noreferrer" instead? https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel/noreferrer

kamtschatka:
Tried it out and it works fine if you click on the image on the dashboard, but does not work when clicking on the image(i guess that is what you mean with "only when viewing them on the homepage"), when you are showing the details of the bookmark, because this requires the target="_blank" attribute. 
We could either:
* add that attribute and have all the links open in a new tab (which i personally prefer)
* apparently using javascript to navigate also works

I am not really a fan of adding a prefix URL, as it makes you dependent on e.g. "anonym.es" to keep your data private....

PR #388, opened 2024-09-11
Removing referrer header when clicking on links #387

kamtschatka:
added "noreferrer" to the links to prevent sending referrers to bookmarked pages

I have added it to the links on the cards itself for now as that is not a change in behavior. 
We could do it for the details also, but it needs some discussion


MohamedBassem:
Thanks @kamtschatka!

Issue #382, opened 2024-09-09
long delay when selecting tags in UI


cirrusflyer:
I have nearly 4000 bookmarks, so a lot of tags.  When I select the "Tags" on the left UI menu, it takes around 14 seconds to display the list.  Hoarder is running on a server with a  Celeron N5105, 16 GB RAM, and SSD storage.

kamtschatka:
This is most likely not related to the database query/your server, but rather on the UI showing all those tags. 
How many tags do you have? Should you not have less tags than bookmarks, to keep everything grouped together?

cirrusflyer:
I have too many tags to count.  Is there a way to get a count?

cirrusflyer:
Looks like over 6000 tags based on a copy and paste to another file and counting the rows.

kamtschatka:
yeah, i was able to reproduce the issue, with 4000 tags that I added directly in the datbase and rendering took about 5 seconds for me. 
I optimized the code and got it to below 2 seconds. Maybe someone else with more knowledge of React has better ideas, but at least it improved quite a bit already.

cirrusflyer:
That's a big improvement! Will that be pushed to Docker soon?

kamtschatka:
Depends on: 
* When it gets merged. I am a contributor, the maintainer needs to review and merge it
* Which docker container you are using. If you are using stable, then it usually takes more than a month to release new versions, so you'll have to wait. If you are using the latest docker container, it would be made available pretty fast after the merge, but you'd be using a development version of hoarder, which potentially has more bugs.

PR #390, opened 2024-09-11
long delay when selecting tags in UI #382

kamtschatka:
improved performance by not handling hover in css
also rendering the draggable div only if draggable mode is active

### DIFF
--- a/apps/web/components/dashboard/tags/TagPill.tsx
+++ b/apps/web/components/dashboard/tags/TagPill.tsx
@@ -1,4 +1,4 @@
-import React from "react";
+import React, { useState } from "react";
 import Link from "next/link";
 import { Button } from "@/components/ui/button";
 import { Separator } from "@/components/ui/separator";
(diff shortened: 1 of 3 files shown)


apps/web/components/dashboard/tags/TagPill.tsx
```diff
@@ -62,27 +67,25 @@ export function TagPill({
...
+        className={
+          "flex gap-2 rounded-md border border-border bg-background px-2 py-1 text-foreground hover:bg-foreground hover:text-background"
+        }
+        href={`/dashboard/tags/${id}`}
+        data-id={id}
+      >
+        {name} <Separator orientation="vertical" /> {count}
+      </Link>
 
+      {isHovered && (
```
MohamedBassem: This seems to break the deletion dialog. Maybe we can have just one dialog in the page, and trigger it with that button.

MohamedBassem (changes requested):

PR #409, opened 2024-09-23
fix(web): Improve SidebarItem tap target accessibility

Trugamr:
This PR enhances the accessibility and user experience of the SidebarItem component by expanding the clickable area to cover the entire item.

#### Before
chrome_Jgp0L9nYyq

#### After
chrome_JsRJSZFLMs


MohamedBassem:
Ah that's a good catch, thanks @Trugamr! I'll review this over the weekend :)

MohamedBassem:
Thanks a lot @Trugamr for the fix, and welcome to the contributors list!

PR #416, opened 2024-09-24
Mobile - Android accepting local imported CAs

CrypticC3s4r:


MohamedBassem:
I'm really sorry this took that long... Thanks for the fix!

Issue #421, opened 2024-09-25
[Bug Report] Importing bookmarks adds all tags to all bookmarks


phampyk:
I wanted to importa all of my raindrop bookmarks into hoarder, it was around 500 to 600 bookmarks in total... The problem is that it added random tags to the bookmarks, and a lot of them... if not all, I even had some repeating bookmarks already on hoarder that were added onto the "imported bookmarks" list and it added all the extra bookmarks too. 
Now my hoarder is really sluggish and even completely froze my machine.

I imported the bookmarks with an HTML file, I can provide the file if you want, generated from a raindrop backup. 

![image](https://github.com/user-attachments/assets/e975e341-1bc8-49ad-9768-aeee0f1a531c)

Same bookmarks in all of the imported bookmarks.

---

![image](https://github.com/user-attachments/assets/f958ab0c-1970-4e22-a831-5dd65ab5a6f7)

All the imported bookmarks have what's basically all the tags I had on hoarder.

---

![image](https://github.com/user-attachments/assets/6b89b428-1b67-406b-b9a7-88cf84b97e55)

If I already had the bookmark previously, it added all the extra tags to them as well.

kamtschatka:
The last point is expected, i guess. Since it already exists, it will simply be updated.
The rest of course does not make any sense. Can you provide the file somehow? Can you maybe share it with me in discord? Or if you don't mind it being public, then upload it somewhere?

kamtschatka:
OK, i found the issue and created a fix

phampyk:
> OK, i found the issue and created a fix

Is there any hope to recover what I have, or should I delete all the tags and retag everything from 0?

MohamedBassem:
@phampyk I think this only affected newly imported bookmarks. So we can help you craft a query to remove all those newly imported bookmarks. We can do it over discord

kamtschatka:
there is no way to roll back unfortunately, but as @MohamedBassem  said, it "only" affects newly imported bookmarks.

PR #428, opened 2024-09-25
[Bug Report] Importing bookmarks adds all tags to all bookmarks #421

kamtschatka:
fixed an issue that caused all existing tags to be assigned to a new bookmark

### DIFF
--- a/packages/trpc/routers/bookmarks.test.ts
+++ b/packages/trpc/routers/bookmarks.test.ts
@@ -169,7 +169,22 @@ describe("Bookmark Routes", () => {
       "tag2",
       "tag3",
       "tag4",
+      "tag5",
     ]);
+
+    await expect(() =>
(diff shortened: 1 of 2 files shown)


MohamedBassem:
Oh, that's an ugly bug. I think I'll have to cut a hotfix for this one.

MohamedBassem:
I think if the user is passing an object with both `tagName` and `tagId` as empty, then the request is invalid. Let me change the check to be a zod validation instead.

kamtschatka:
but that will show up as a lot of errors then? 
you'll at least have to prevent it on the client side as well, or not?

MohamedBassem:
@kamtschatka Yeah, I found the bug in the client side. Apparently I fixed it for Pocket but not for the netscape importer.

MohamedBassem:
The frontend fix: https://github.com/hoarder-app/hoarder/commit/4db50c234fb0da8747b902a7c1e551cc54451f4c

Issue #424, opened 2024-09-25
inference job failed


hujh818:
2024-09-25T03:12:13.473Z info: [inference][15] Starting an inference job for bookmark with id "z3nf9s85x282vcvf42pp7v9u"
2024-09-25T03:12:14.911Z info: [inference][15] Starting an inference job for bookmark with id "z3nf9s85x282vcvf42pp7v9u"
2024-09-25T03:12:16.418Z info: [inference][15] Starting an inference job for bookmark with id "z3nf9s85x282vcvf42pp7v9u"
2024-09-25T03:12:17.703Z error: [inference][15] inference job failed: TypeError: Cannot read properties of undefined (reading '0')

kamtschatka:
Please explain what you did and what you checked.
Is this a link bookmark or a text bookmark?
is your Meilisearch connected and working? 
Does it work for other bookmarks and only a specific one is failing, or is this happening all the time?

Diew:
for me it happen all the time after update  1-2 day ago (maybe ? im not sure but it like have update on portainer)

then i try to test by start fresh and it still happen on both text and url

everything work fine but when it try to do inference job in log it error like first post guy

test on nearly default docker compose but change

OPENAI_API_KEY
NEXTAUTH_SECRET
MEILI_MASTER_KEY
and add 
OPENAI_BASE_URL
for openrouter

english not my native hope you can understand what im saying

![Screenshot 2024-09-26 at 01-17-08 Portainer docker-web](https://github.com/user-attachments/assets/fbe430fe-5f63-492b-8bf2-df4b09858893)

![Screenshot 2024-09-26 at 01-17-22 Hoarder](https://github.com/user-attachments/assets/d48da2c2-9741-4b08-82a5-227d6e29202b)

kamtschatka:
Unfortunately the logging does not show where the issue happens. I have added better logging, once that is merged, you'll have to change your deployment to use the "latest" docker container and we'll see where the error actually happens.

hujh818:
> Please explain what you did and what you checked. Is this a link bookmark or a text bookmark? is your Meilisearch connected and working? Does it work for other bookmarks and only a specific one is failing, or is this happening all the time?

1. a link bookmark 
2.  Meilisearch connected and working
3.  happening all the time

hujh818:
> for me it happen all the time after update 1-2 day ago (maybe ? im not sure but it like have update on portainer)
> 
> then i try to test by start fresh and it still happen on both text and url
[...]

PR #429, opened 2024-09-25
inference job failed #424

kamtschatka:
extended logging when an exception occurrs, so it is possible to see the stacktrace of a failed execution

### DIFF
--- a/apps/workers/crawlerWorker.ts
+++ b/apps/workers/crawlerWorker.ts
@@ -172,7 +172,9 @@ export class CrawlerWorker {
         },
         onError: async (job) => {
           const jobId = job?.id ?? "unknown";
-          logger.error(`[Crawler][${jobId}] Crawling job failed: ${job.error}`);
+          logger.error(
+            `[Crawler][${jobId}] Crawling job failed: ${job.error}\n${job.error.stack}`,
+          );
           const bookmarkId = job.data?.bookmarkId;
           if (bookmarkId) {
             await changeBookmarkStatus(bookmarkId, "failure");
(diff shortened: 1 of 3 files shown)


MohamedBassem:
I think the queue runner library currently doesn't do any logging (to give control to the user). Mind moving the log line to the error handler of the workers instead?

kamtschatka:
there is already a log, but since it is async, the stacktrace is missing the information that helps

MohamedBassem:
Not sure I understand the async part. How is logging `error.stack` inside the worker, is different than logging `error.stack` inside the onerror callback given that they are the same object?

MohamedBassem:
Something like this:

```diff
diff --git a/apps/workers/openaiWorker.ts b/apps/workers/openaiWorker.ts
index 8bd2cf4..8cf4b5a 100644
--- a/apps/workers/openaiWorker.ts
+++ b/apps/workers/openaiWorker.ts
@@ -73,7 +73,7 @@ export class OpenAiWorker {
         onError: async (job) => {
           const jobId = job?.id ?? "unknown";
           logger.error(
-            `[inference][${jobId}] inference job failed: ${job.error}`,
+            `[inference][${jobId}] inference job failed: ${job.error}.\n${job.error.stack}`,
           );
           await attemptMarkTaggingStatus(job?.data, "failure");
         },
diff --git a/apps/workers/searchWorker.ts b/apps/workers/searchWorker.ts
index 56c2aac..0e8fea0 100644
--- a/apps/workers/searchWorker.ts
+++ b/apps/workers/searchWorker.ts
@@ -25,7 +25,7 @@ export class SearchIndexingWorker {
         },
         onError: (job) => {
           const jobId = job?.id ?? "unknown";
-          logger.error(`[search][${jobId}] search job failed: ${job.error}`);
+          logger.error(`[search][${jobId}] search job failed: ${job.error}.\n${job.error.stack}`);
           return Promise.resolve();
         },
       },
```

Issue #407, opened 2024-09-22
Better handling env NEXTAUTH_URL


ElTopo:
There is an env var in .env file: NEXTAUTH_URL

By default, it's "http://localhost:3000", and the admin is supposed to change it to "https://myhoarder.myname.com".

This is not very friendly to reverse proxies. The app (Hoarder-app) itself should not know or care about by which host or port the clients are accessing its data. If the Hoard-app server has several domain names, the env NEXTAUTH_URL might direct the clients to the wrong sign in page URL when they log out.

For example, if my hoarder server has 2 URLs: 
https://hd.myname1.com
https://hd.myname2.org
There would be always chances for a user to log out their account and found they were at the wrong server sign in page.

How about hoard server does NOT use the env var NEXTAUTH_URL, and always uses whatever the current server?

e.g., logging out from https://hd.myname1.com, go to https://hd.myname1.com/signin; logging out from https://hd.myname2.org, go to https://hd.myname2.org/signin.

kamtschatka:
From what I can tell, this is not really up to us to implement, as this is a nextauth specific thing. From what Copilot said, you can keep it empty and then everything will be relative to the URL, but OAuth will not work. 
I am not sure if Copilot is correct, but you could try it. 
Other than that, I don't see any way how hoarder could fix this

ElTopo:
Well I did try to leave it empty (NEXTAUTH_URL=""), when the user logged out at https://myhd.myname.com, the server returned to http://localhost:3000, which of course is not accessible.

kamtschatka:
So I was able to improve the situation for local logins, but this WILL NOT work for OAUTH. It requires some cookies to be set and you need the NEXTAUTH_URL to be configured to redirect back to your instance after you authenticated at the 3rd party provider.
I personally do not like this solution, but it works.
Ultimately it is up to @MohamedBassem to decide if we want to merge this or if we close this issue as not supported (or if there is even another solution, but information on this is pretty scarce)

ElTopo:
@kamtschatka sounds good to me, thanks.

I was not aware that this was for OAUTH, might be better to have another option for OAUTH: if OAUTH is enabled, require NEXTAUTH_URL as the real auth site; if OAUTH is disabled, then NEXTAUTH_URL could be empty, in that case, just go back to /signin.

BertanAygun:
[...]

PR #437, opened 2024-09-28
Better handling env NEXTAUTH_URL #407

kamtschatka:
improved handling to reuse the existing URL

### DIFF
--- a/apps/web/server/auth.ts
+++ b/apps/web/server/auth.ts
@@ -172,6 +172,9 @@ export const authOptions: NextAuthOptions = {
       }
       return token;
     },
+    async redirect({ url }) {
+      return url;
+    },
     async session({ session, token }) {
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Can you explain a bit how this improves the handling of the URL?

kamtschatka:
when I tried to access hoarder using localhost/my ip address and returned the URL without having NEXTAUTH_URL set, it would just reuse whatever URL you accessed hoarder with. Otherwise it magically just used localhost:3000, no matter if you used localhost or the IP address.
Doesn't work of course if you are using OAuth, as that always requires the URL

MohamedBassem:
I wouldn't worry about people not setting NEXTAUTH_URL for now, it's a required env, so I'd just keep it.

PR #445, opened 2024-09-30
inference job failed #424

kamtschatka:
added handling for the case where the chatCompletion is returned as a string

NOTE:
* I can not reproduce this locally, but I worked with @hujh818 over discord who helped me figure out what is going on
* I have absolutely no idea, why that is a string, but the logs he added, clearly showed that it is a string and after adding a JSON.parse, he confirmed that it is working now...

### DIFF
--- a/apps/workers/inference.ts
+++ b/apps/workers/inference.ts
@@ -1,5 +1,6 @@
 import { Ollama } from "ollama";
 import OpenAI from "openai";
+import { ChatCompletion } from "openai/resources/chat/completions";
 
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Not a big fan of this at all to be honest. The best we can do is update the openai package and hope it resolves the issue. If it doesn't, then they should take it with the API provider. Upgraded openai here https://github.com/hoarder-app/hoarder/commit/b56655cd45843f7206206b62d5daf60fcd85e810

PR #456, opened 2024-10-03
Fixing Docker Dev Environment

CrypticC3s4r:
I had issues running the Docker development environment without first setting up a local development environment.
And put the Environment variables into a seperate .env file

### DIFF
--- a/docker/Dockerfile.dev
+++ b/docker/Dockerfile.dev
@@ -1,6 +1,14 @@
 FROM node:21-alpine
 
-RUN apk add --no-cache libc6-compat make g++ py3-pip linux-headers
-ENV PUPPETEER_SKIP_DOWNLOAD true
-
 WORKDIR /app
+ENV PNPM_HOME="/pnpm"
+ENV PATH="$PNPM_HOME:$PATH"
+
+# Check https://github.com/nodejs/docker-node/tree/b4117f9333da4138b03a546ec926ef50a31506c3#nodealpine to understand why libc6-compat might be needed.
+RUN apk add --no-cache libc6-compat make g++ py3-pip linux-headers git
+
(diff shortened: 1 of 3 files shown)


docker/Dockerfile.dev
```diff
@@ -1,6 +1,14 @@
-FROM node:21-alpine
-
-RUN apk add --no-cache libc6-compat make g++ py3-pip linux-headers
-ENV PUPPETEER_SKIP_DOWNLOAD true
+FROM node:21-alpine AS base
```
MohamedBassem: The "as base" here is not needed.

docker/docker-compose.dev.yml
```diff
@@ -8,9 +8,8 @@ services:
       - ..:/app
     ports:
       - 3000:3000
-    environment:
-      MEILI_ADDR: http://meilisearch:7700
-      DATA_DIR: /data
```
MohamedBassem, line 13: I think those are still needed (similar to those of the workers)

docker/docker-compose.dev.yml
```diff
@@ -8,9 +8,8 @@ services:
...
+    env_file:
+      - .env  # Automatically load all variables from this file
```
MohamedBassem, line 12: If you want to add a `.env`, let's add a `.env.sample` (even if it's empty) in the directory so that people can get it up and running quickly.

MohamedBassem (changes requested):
Thanks for taking the time to fix the docker dev env! Left some small comments.

CrypticC3s4r:
Removed the BASE and introduced an .env.sample for configuration recommendation

MohamedBassem:
Thanks @CrypticC3s4r!

PR #465, opened 2024-10-04
Fix typo in Docker Installation docs

jarednipper:
"greated" should be "greeted".


MohamedBassem:
@jarednipper Thank you!

PR #469, opened 2024-10-04
Introducing nodemon with legacywatch for workers

CrypticC3s4r:
With my Setup the tsx watcher does not run correctly.
Introducing nodemon with its legacy-watch fixes this problem


MohamedBassem:
What kind of issues are you facing with tsx watch?

CrypticC3s4r:
I have removed the tsx watch and changed it to nodemon.

The normal tsx watch does not work with my development setup, where I needed to introduce nodemon with its legacy watch feature. 

As a result I do not need to restart all docker machines to get my changes compiled rather than the nodemon recognizes a change at compiles it "on the go"

MohamedBassem:
Trying to cleanup some of the old PRs, I don't think anyone else reported problems with `tsx`, so closing this PR for now.

PR #481, opened 2024-10-04
https://github.com/hoarder-app/hoarder/issues/459

kamtschatka:
made the search and bulk edit controls sticky
fixed the z index of the tooltips to be above the bulk edit checkmarks updated the search page as well to look and behave the same as the home page


MohamedBassem:
This breaks with the preview modal:

Screenshot 2024-10-05 at 14 56 28

And it looks a bit off when scrolling ue to the lack of continuation between the navbar and the sidebar:

Screenshot 2024-10-05 at 14 58 37

I'd say, let's wait on this until we have a proper sticky navbar which according to the figma designs in discord, we'll be having soon.

PR #482, opened 2024-10-04
fix dead installation link

disappointingsupernova:


disappointingsupernova:
The installation link in the README is dead. A capitalised I in Installation

MohamedBassem:
hmmm, the link in the readme works fine for me

kamtschatka:
yeah works for me as well, but technically, with the uppercase it returns status code 200, with lowercase you get status code 404 and it works itself out somehow. 
So since the PR is already up, I would just merge it, it is definitely correct.

MohamedBassem:
Oh, the 404 is surprising.

PR #483, opened 2024-10-04
Introducing Accepting Cookie Consent lite

CrypticC3s4r:
Tried to create a more generic version to accept common cookie consent banners and trying to hide those in addition.
Since its more of a one size fits all solution, there could be of course i.e. a button named "ok" on a page which might be clicked, if its not a cookie consent button.


MohamedBassem:
We're already using an adblocker that should be doing something similar. In the next release we're also going to migrating to playwright and will introduce more per-site rules, so I think I'll close this for now.

PR #485, opened 2024-10-05
Adding txt file support for Bookmarks import in Settings

CrypticC3s4r:
This is for my own need and I hope someone else needs it too.

Importing a txt file with tags or without is possible with it.

In addition I had issues in not setting the title and letting the crawler read it. 
As a result I changed the code slightly to not send any title if there is no title set.

Example with Tags
```
https://hoarder.app/ [Hoarder, BestApp]
https://....
```

Example without Tags
```
https://hoarder.app/ 
https://....
```

### DIFF
--- a/apps/web/components/dashboard/settings/ImportExport.tsx
+++ b/apps/web/components/dashboard/settings/ImportExport.tsx
@@ -11,6 +11,7 @@ import {
   ParsedBookmark,
   parseNetscapeBookmarkFile,
   parsePocketBookmarkFile,
+  parseTextBookmarkFile,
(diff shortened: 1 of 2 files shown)


MohamedBassem:
@CrypticC3s4r I don't understand the benefit of this format to be honest. If it's just a list of URLs, then pasting them in the editor card will just work. And the one with the tags on the same line is something that people won't typically get exported from anywhere? If it's just a one off, maybe using something like the CLI is better to import them is better. But I don't think it's worth adding it as a builtin format in the app.

Issue #406, opened 2024-09-22
[Feature Request] Allow to disable default password log in after SSO is configured


MFYDev:
Reference: https://github.com/hoarder-app/hoarder/issues/92

Thank you for the wonderful project! And finally I can use SSO to secure my instance.

However after the SSO is set, can we add an env variable like DISABLE_DEFAULT_LOGIN or DISABLE_LOGIN_FORM to disable the default email and password log in form? Only through this way the SSO can properly protect the Hoarder itself.

Thank you so much for your effort again!

IamTaoChen:
Maybe the `DISABLE_SIGNUPS` should be split for the local account and OIDC. OIDC account should be created automatically and the local account should be disable for signup.

PR #502, opened 2024-10-06
[Feature Request] Allow to disable default password log in after SSO is configured #406

kamtschatka:
changed the flag to also disallow logging in via password
The extensions will also no longer be allowed to log in via username/password then

### DIFF
--- a/apps/web/components/signin/CredentialsForm.tsx
+++ b/apps/web/components/signin/CredentialsForm.tsx
@@ -35,6 +35,8 @@ function SignIn() {
   const [signinError, setSigninError] = useState("");
   const router = useRouter();
   const searchParams = useSearchParams();
+  const clientConfig = useClientConfig();
+
   const oAuthError = searchParams.get("error");
   if (oAuthError && !signinError) {
     setSigninError(`${OAUTH_FAILED} ${oAuthError}`);
@@ -43,6 +45,14 @@ function SignIn() {
   const form = useForm<z.infer<typeof signInSchema>>({
     resolver: zodResolver(signInSchema),
   });
+
+  if (clientConfig.auth.disablePasswordAuth) {
+    return (
+      <p className="text-center">
+        Password authentication is currently disabled.
+      </p>
+    );
+  }
 
   return (
     <Form {...form}>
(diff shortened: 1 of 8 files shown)


mikevanes:
Nice! Was looking for this.

packages/trpc/routers/apiKeys.ts
```diff
@@ -74,6 +75,13 @@ export const apiKeysAppRouter = router({
     .output(zApiKeySchema)
     .mutation(async ({ input }) => {
       let user;
+      // Special handling as otherwise the extension would show "username or password is wrong"
+      if (serverConfig.auth.disablePasswordAuth) {
+        throw new TRPCError({
+          message: "Password authentication is currently disabled",
+          code: "FORBIDDEN",
```
MohamedBassem, line 82: I think UNAUTHORIZED is more suitable here.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/401

> A 401 Unauthorized is similar to the [403 Forbidden](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) response, except that a 403 is returned when a request contains valid credentials, but the client does not have permissions to perform a certain action.

In that case, we don't have valid creds. So I think a 401 is more suitable. What do you think?

apps/web/components/signin/CredentialsForm.tsx
```diff
@@ -44,6 +46,14 @@ function SignIn() {
     resolver: zodResolver(signInSchema),
   });
 
+  if (clientConfig.auth.disablePasswordAuth) {
+    return (
```
MohamedBassem, line 50: If we do this, api key signin errors won't get rendered. Maybe you should inline the error handling here as well?

MohamedBassem (changes requested):

packages/trpc/routers/apiKeys.ts
kamtschatka: 1. I put the reason for this in the comment above. If we return 401, the extension will show "Wrong username or password", which is not correct, since they might be correct, they might be wrong. We did not check. Additionally it will make problem solving harder, since it is very misleading to tell the users that their credentials are wrong, when they are possibly correct, but the admin decided to turn off the login via credentials. I have looked at changing the way we show the message in the extension, but that would mean just taking whatever the server sends as a message, which is currently very revealing (e.g. user exists, but password is wrong), so that is not good from a security perspective.
2. I find "forbidden" much better as it is actively forbidden to sign in this way. It does not fulfill the first part of the spec though (valid credentials, as we never check), but the second part is fulfilled. But I also do not find 401 correct, since you might have valid credentials, but we did not check.

PR #505, opened 2024-10-07
feature: Allow the user to configure fully custom LLM prompts

Papierkorb:
Hello!

I just found this great piece of software! I was thrilled to see that I can use my own OpenAI endpoint and point it towards my Llama 3.1 8B endpoint.

# Status Quo

I think the standard prompt as set in `packages/shared/prompts.ts` could require some tweaking.
The default prompt doesn't work at all with my model (Which I think is a pretty commonly used one). This one does work great:

```
You're an expert document tagger. Given an input you generate an JSON object of the following format: { "tags": ["First tag", "Another tag", ...] }

Only respond in this format! Write only the JSON data and nothing else. Do not write an explanation. If you can't find tags you're satisfied with respond with an empty tags array. Write at most five tags. Write your tags in ${lang}.

${content}
```

I've only tested this prompt manually with my model. With this, my 0% success rate turns into a 100% success rate; Which surprises me, usually you should accept if the model gives an introduction like "Your requested JSON document: ...".

# Thus

Long story short: This PR adds the "TEXT_TAG_PROMPT" and "IMAGE_TAG_PROMPT" environment variables. If not set, then the already existing prompt is used. If it is set, it's used instead. The variables are then interpolated `Like {{this}}`, akin to Jinja2 templates (Which the project may adopt in the future?)

<details><summary>My template in this syntax</summary>
<p>

```
You're an expert document tagger. Given an input you generate an JSON object of the following format: { "tags": ["First tag", "Another tag", ...] }

Only respond in this format! Write only the JSON data and nothing else. Do not write an explanation. If you can't find tags you're satisfied with respond with an empty tags array. Write at most five tags. Write your tags in {{lang}}.

{{content}}
```

</p>
</details> 

**Not Great**: As is, this change breaks `AISettings.tsx`, as it would now require the server configuration to work. For my tests I simply removed the calls to the buildPrompt functions - But this of course isn't acceptable for merging. I'd need a little help here on how we should approach this.


MohamedBassem:
Hey, thanks for taking the time, attempting to tweak the prompt and sending the PR. However, I intentionally don't want to lose control over the prompt. I want to keep the ability to change the format of the output in the future to include extra stuff, etc. If I let people replace the prompt completely, their future upgrades might not be backward compatible.

My plan is to give the people the option to "customize" the prompt and add extra rules. But I don't want people to completely replace it.  The next release allows people to add new rules, but I'm also planning on letting the people modify the prebuilt rules.

Now, regarding the base prompt. I'd be happy to accept tweaking for it if you think your prompt achieves better result in llama, but we'd probably also want to test it on gpt models as well.

kamtschatka:
I tried it out with 34 links and it failed on 7 of them(also llama 3).
This happened before, so a while back I did a PR that moved the whole "respond in json and this is how it should look like" to the end of the prompt and that fixed it. (I just tried the unmodified version version as well and all of them worked).
So at this point, we can't merge this, as it makes things worse for existing users.

Papierkorb:
I'm not saying that this should be merged outright. However, a 16% failure rate is already a lot better than the 100% failure rate I saw on my end.

I'm pretty sure that the issue lies in the code assuming that `json_object` mode working. If you're accessing OpenAI that's of course fine - But custom engines don't support it (At least I'm not aware of one). As solution, one usually does:

1. Lenient parsing: Search for the first `{` and last `}` in the response and only parse those parts
2. Retry: If it failed, re-try. Also helps with short-while networking issues

With this I usually see an almost perfect success rate. These changes wouldn't hurt OpenAI as `json_object` mode is still there. They'd help Ollama just as much as custom OpenAI endpoints.

kamtschatka:
It makes things better for you, but makes things worse for everyone else though, so not very promising.
Since you are knowledgeable in this area, how about you come to the discord channel (see readme.md, there is no link, only an image with a link) to discuss this further?

poculeka:
@Papierkorb I had the same issue and I think incorrect responses could be a result of the prompt being truncated incorrectly on Hoarder side. In my Ollama instance I found logs like this:
```
time=2025-02-02T12:06:16.960Z level=WARN source=runner.go:129 msg="truncating input prompt" limit=2048 prompt=2145 keep=5 new=2048
```
even with `INFERENCE_CONTEXT_LENGTH` being lower than 2048. I observed that whenever the prompt was truncated by Ollama, the results couldn't be parsed by Hoarder. I don't know how truncation procedure looks inside Ollama (maybe it's model-specific?), but I guess it might just cut the last part, which, in case of the default prompt, specifies expected JSON response.

Moreover, `INFERENCE_CONTEXT_LENGTH` is passed to Ollama API as a request parameter. This means that even if the prompt would fit Ollama's context size (2048 by default), it still might be truncated by Ollama.

I fixed this in my own Hoarder instance by removing the line `num_ctx: serverConfig.inference.contextLength,` in `OllamaInferenceClient` and setting `INFERENCE_CONTEXT_LENGTH` to 1024. This way the prompt sent to Ollama API seems to always fit Ollama's 2048 token context size.

MohamedBassem:
Prompt truncation kinda got fixed with #1405. Given that we agreed we're not changing the prompt, closing this PR.

PR #509, opened 2024-10-07
Add Source Label to Dockerfile

wuast94:
To get changelogs shown with Renovate a docker container has to add the source label described in the OCI Image Format Specification.

For reference: https://github.com/renovatebot/renovate/blob/main/lib/modules/datasource/docker/readme.md


MohamedBassem:
Thanks!

Issue #517, opened 2024-10-08
[Bug] Degraded performance when rendering large images in the bookmarkgrid


Antebios:
I have over 15k bookmarks that it took a long while to import.  After it finally imported the UI performance was excruciatingly slow.  I waited a few days later to try to use it again since it might have a background job, but it was still slow.  I even restarted the service, but still slow.  I have given up using it.

MohamedBassem:
hey, thanks for creating the issue. However, it would be great if you can give us something to help you with here. Some suggestions:
1. A screen recording or something showing us the slowness you're facing?
2. A description of the hardware you're running hoarder on?
3. You mentioned background jobs, maybe check the admin panel and see if there are more background jobs that are still running?
4. Screenshots of the cpu utilization of the server you're running hoarder on?

Without any of this info, I'm not sure we can help you unfortunately.

kamtschatka:
Yeah this will need more information. I actually added 15k bookmarks to my hoarder dev instance yesterday and I don't see any noticable slowdown. Since pagination is used, I also don't see why it would be affected that much by the 15k bookmarks.
And yes, there are some queries that go through all the data on the server side, but 15k rows should not be that much of an issue for the database. 
So please provide the information above. Are you maybe adding a lot of pdf files to hoarder and the preview of the files takes too long? Or is it only on the "tags" page, because there is an issue open, that this takes very long to load and we will have to improve the performance there.

kissgyorgy:
Hello! Great app, looks very promising!

I had the same feeling but only with 10 bookmarks imported that the UI seems very sluggish.
I just did a Lighhouse test and it seems like the biggest problems are that huge pictures (screenshots of websites) are loaded on every page. I have `CRAWLER_FULL_PAGE_SCREENSHOT=true˙ and `CRAWLER_FULL_PAGE_ARCHIVE=true` and intend to use Hoarder this way.

Here is a Lighthouse HTML report I just did:
[lighthouse-report-huge-images.zip](https://github.com/user-attachments/files/17304418/lighthouse-report-huge-images.zip)

I think it's easy to reproduce just saving some big pages with full-page screenshots.

kamtschatka:
hi, thanks that makes a lot of sense. 
We should probably calculate preview images for those cards and only deliver the full image, when you actually open the cards.

kamtschatka:
@kissgyorgy would you be able to give me a list of the your bookmarks that caused this slowdown?
[...]

PR #526, opened 2024-10-12
Performance #517

kamtschatka:
made sure the image size is based on the bookmark containers width and not on the full page width


packages/shared-react/utils/assetUtils.ts
```diff
@@ -1,3 +1,3 @@
-export function getAssetUrl(assetId: string) {
-  return `/api/assets/${assetId}`;
+export function getAssetUrl(assetId: string, preview?: boolean) {
+  return `/api/assets/${assetId}${preview ? "?preview" : ""}`;
```
kamtschatka, line 2: as for why i am using the querystring "?preview" instead of "?preview=true" (or something similar):
The <Image> element would then URL encode the "=" sign, which breaks the URL for some reason and then route.ts will not be called anymore.

PR #540, opened 2024-10-15
Made the thumbnail image keeping its real ratio instead of object-cover like Pinterest layout

echo-saurav:
Screenshot 2024-10-15 at 9 32 53 PM


echo-saurav:
So i have change the max height to `max-h-screen` , i guess its enough ?

PR #546, opened 2024-10-16
Adding issue templates

Inventhrice:
I found that making issues was a little inconsistent, so I wanted to throw my two cents in and provide some issue templates! I've taken inspiration from both NewPipe and BookStack's issue templates and modified them to fit Hoarder.


MohamedBassem:
This is very cool, thanks a lot!

PR #557, opened 2024-10-17
feat: add cli option to read note from stdin

lippoliv:
Like in "add" command

### DIFF
--- a/apps/cli/src/commands/bookmarks.ts
+++ b/apps/cli/src/commands/bookmarks.ts
@@ -187,6 +187,7 @@ bookmarkCmd
   .description("update a bookmark")
   .option("--title <title>", "if set, the bookmark's title will be updated")
   .option("--note <note>", "if set, the bookmark's note will be updated")
+  .option("--stdin", "if set, the bookmark's note will be updated by stdin (not to be used with --note)")
   .option("--archive", "if set, the bookmark will be archived")
   .option("--no-archive", "if set, the bookmark will be unarchived")
   .option("--favourite", "if set, the bookmark will be favourited")
@@ -194,6 +195,11 @@ bookmarkCmd
   .argument("<id>", "the id of the bookmark to update")
   .action(async (id, opts) => {
     const api = getAPIClient();
+
+    if (opts.stdin) {
+      opts.note = fs.readFileSync(0, "utf-8");
+    }
+
     await api.bookmarks.updateBookmark
       .mutate({
         bookmarkId: id,


apps/cli/src/commands/bookmarks.ts
```diff
@@ -187,13 +187,19 @@ bookmarkCmd
   .description("update a bookmark")
   .option("--title <title>", "if set, the bookmark's title will be updated")
   .option("--note <note>", "if set, the bookmark's note will be updated")
+  .option("--stdin", "if set, the bookmark's note will be updated by stdin (not to be used with --note)")
```
MohamedBassem, line 190: Can we give the flag a better name to indicate that it's updating the notes? Maybe `--note-stdin` or something?

apps/cli/src/commands/bookmarks.ts
```diff
@@ -187,13 +187,19 @@ bookmarkCmd
...
   .option("--archive", "if set, the bookmark will be archived")
   .option("--no-archive", "if set, the bookmark will be unarchived")
   .option("--favourite", "if set, the bookmark will be favourited")
   .option("--no-favourite", "if set, the bookmark will be unfavourited")
   .argument("<id>", "the id of the bookmark to update")
   .action(async (id, opts) => {
     const api = getAPIClient();
+
+    if (opts.stdin) {
```
MohamedBassem, line 199: Can we validate here that the `--note` flag is not set as well and print an error if it's not?

MohamedBassem (changes requested):
Thanks for the PR, I think this makes sense. Left a couple of small comments.

apps/cli/src/commands/bookmarks.ts
lippoliv: Indeed I wondered about `--stdin` for `add` parameter as well. It's unclear what it's for. But I decided to be consistent. I'ld suggest to rename `--stdin` for both commands after the merge? Or rename it in parallel and I'll update + adapbt before merging this PR?

apps/cli/src/commands/bookmarks.ts
lippoliv: Done

MohamedBassem:
I'm really sorry this took this long to get back to the PR. On a second though, I don't think a `--stdin` intuitively will mean updating the node of a bookmark. For example, if it's a text bookmark i'd expect it to update the content of the bookmark itself. I think for now, I'll close the PR and we can re-open if there's a better way to do it (or if there's still demand for it).

PR #562, opened 2024-10-17
Add share to PWA

wiinxt:
Adds PWA in sharing options for images and PDF files, as well as sharing selected text.
I know a native app exists but I prefer the PWA, so I'm adding a button for upload files (in [another pull request](https://github.com/hoarder-app/hoarder/pull/561)) and the share option because it was missing.

Hoarder is the PWA version

---

I'm not sure if this is the best possible implementation as I'm not really familiar with Next, I'm open to your comments

### DIFF
--- a/apps/mobile/lib/upload.ts
+++ b/apps/mobile/lib/upload.ts
@@ -56,10 +56,11 @@ export function useUploadAsset(
       return zUploadResponseSchema.parse(await resp.json());
     },
     onSuccess: (resp) => {
-      const assetId = resp.assetId;
-      const assetType =
-        resp.contentType === "application/pdf" ? "pdf" : "image";
-      createBookmark({ type: BookmarkTypes.ASSET, assetId, assetType });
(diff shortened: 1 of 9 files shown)


Tukks:
Would be great if merged.
I use the pwa too

thiswillbeyourgithub:
Hi @wiinxt , I am interested in this feature. Unless I'm mistaken this is on hold because there are unresolved conflicts on your side? Or is there some other reason this has not been merged?

MohamedBassem:
Honestly, it's not @wiinxt fault, this is a 7 months old PR that I just didn't review in time. Back then we still didn't have a proper RESTful API. I'd recommend starting a new PR @thiswillbeyourgithub  because a bunch of things have changed since then and we now have the singleFile endpoint that would probably be very similar to that one.

It's also never too late to apologize to @wiinxt. I shouldn't have left this PR unreviewed for that long. I'm sorry.

Issue #477, opened 2024-10-04
[Feature Request] Log failed login attempts for fail2ban implementation 


timmish:
Hi there,
thanks for the great project, started using it a few days ago and already loving it.

I'm not using any SSO provider, but exposing behind a firewall with geo-ip blocking and a fail2ban implementation. 
It would be great if Hoarder had a log-file accessible in the data folder in which failed login attempts are logged, including the ip of the attempt. 
This way I could easily implement fail2ban.

Thanks for looking into it!

thiswillbeyourgithub:
Hi @timmish, would you mind sharing your fail2ban filters for karakeep?

PR #569, opened 2024-10-19
[Feature Request] Log failed login attempts for fail2ban implementation #477

kamtschatka:
added more logging for extension related logins
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/hoarder-app/hoarder/pull/569).
* __->__ #569
* #576
* #575


apps/web/server/auth.ts
```diff
@@ -69,6 +70,16 @@ async function isAdmin(email: string): Promise<boolean> {
   return res?.role == "admin";
 }
 
+function logAuthenticationError(
+  user: string,
+  message: string,
+  ip: unknown,
```
MohamedBassem: Why `unknown`?

apps/workers/crawlerWorker.ts
```diff
@@ -436,7 +436,7 @@ async function archiveWebpage(
 
   await execa({
     input: html,
-  })`monolith  - -Ije -t 5 -b ${url} -o ${assetPath}`;
+  })`C:\\Projekte\\hoarder-app\\data\\stuff\\monolith.exe  - -Ije -t 5 -b ${url} -o ${assetPath}`;
```
MohamedBassem: Those are irrelevant changes

apps/workers/inference.ts
```diff
@@ -43,7 +43,7 @@ class OpenAIInferenceClient implements InferenceClient {
 
   async inferFromText(prompt: string): Promise<InferenceResponse> {
     const chatCompletion = await this.openAI.chat.completions.create({
-      messages: [{ role: "system", content: prompt }],
+      messages: [{ role: "user", content: prompt }],
```
MohamedBassem: Same here

packages/shared/logger.ts
```diff
@@ -15,3 +15,22 @@ const logger = winston.createLogger({
...
+  format: winston.format.combine(
+    winston.format.timestamp(),
+    winston.format.printf(
+      (info) => `${info.timestamp} ${info.level}: ${info.message}`,
+    ),
+  ),
+  transports: [
+    new winston.transports.Console(),
+    new winston.transports.File({
+      filename: "failedLogins.log",
```
MohamedBassem:
```suggestion
      filename: "auth_failures.log",
```

packages/shared/logger.ts
```diff
@@ -15,3 +15,22 @@ const logger = winston.createLogger({
 });
 
 export default logger;
+
+export const authLogger = winston.createLogger({
```
MohamedBassem: Given that we're writing to a file called `failedLogins`, I think this should indicate a failure instead of a generic `authLogger`.
```suggestion
export const authFailureLogger = winston.createLogger({
```

apps/web/server/auth.ts
```diff
@@ -77,17 +88,34 @@ const providers: Provider[] = [
...
-        return await validatePassword(
+        const isValidPassword = await validatePassword(
           credentials?.email,
           credentials?.password,
         );
+        logAuthenticationError(
+          credentials?.email,
+          "Password invalid",
+          req.headers?.["x-forwarded-for"],
+        );
```
MohamedBassem: It's not clear to me why is this an authentication error? Also `validatePassword` returns the user or throws an error. So it's return value being `isValidPassword` is confusing.

apps/web/server/auth.ts
```diff
@@ -77,17 +88,34 @@ const providers: Provider[] = [
...
+        return isValidPassword;
       } catch (e) {
+        const error = e as Error;
+        logAuthenticationError(
+          credentials?.email,
+          error.message,
+          req.headers?.["x-forwarded-for"],
```
MohamedBassem: Let's use (https://github.com/pbojinov/request-ip) to get the real IP. It's much more complicated than just the `x-forwarded-for`.

MohamedBassem (changes requested):
We probably should also log in the `APIKey.exachange` endpoint.

kamtschatka:
the NextAuthRequest and the node Request are not really compatible, so I was not able to find a solution that works to find the IP address, but I am open to suggestions

Issue #580, opened 2024-10-22
Updating docker compose documentation


Inventhrice:
### Describe the Bug

This is me documenting my steps as I setup a dev environment for a Windows Machine.

1. `time="2024-10-22T16:20:51-04:00" level=warning msg="\\hoarder-dev\\hoarder\\docker\\docker-compose.dev.yml: `version` is obsolete"
env file \hoarder-dev\hoarder\docker\.env not found: CreateFile \hoarder-dev\hoarder\docker\.env: The system cannot find the file specified.`

The instructions on [the website](https://docs.hoarder.app/Development/setup/) says to `Start by copying the template by cp .env.sample .env.` If we want this to be true, then we should change the docker-compose.dev.yaml to look one directory above.

2. So far I am seeing a lot of `Progress: resolved 0, reused 0, downloaded 442, added 456`. Will keep updated on if this is expected.

If I have understood this correctly, apparently this step is ` Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-9.0.0-alpha.8.tgz`. It looks like it is installing 3121 packages. Just noting here that this is giving my laptop a RUN for it's money, fans are spinning lol.

i was right 😄 

3. I am waiting for some confirmation that I can start developing.

4. Upon navigation to localhost:3000, i got `Error: Cannot open database because the directory does not exist`

### Steps to Reproduce

ref above

### Expected Behaviour

N/A

### Device Details

Windows 11

### Exact Hoarder Version

v0.18.0

Inventhrice:
4. Upon navigation to localhost:3000, i got `Error: Cannot open database because the directory does not exist`

![image](https://github.com/user-attachments/assets/01a41a3c-3387-4235-9e12-d421480762bb)

From `/packages/db/drizzle.config.ts`, we find:
```
const databaseURL = serverConfig.dataDir
  ? `${serverConfig.dataDir}/db.db`
  : "./db.db";
```
serverConfig.dataDir comes from ..\shared\config.ts. dataDir is defined at line 106 `    dataDir: val.DATA_DIR,`, which comes from line 45 `  DATA_DIR: z.string().default(""),` which comes from line 1 `import { z } from "zod";`.

what the heck is "zod".

I have `DATA_DIR=../hoarder-data/` in .env, and that does not seem to have worked.

### Note for Disabling Sign In's as per Issue 556
There seems to be variables called 
```
  DEMO_MODE: stringBool("false"),
[...]

PR #584, opened 2024-10-24
Update Docker Compose Dev Environment (#580)

Inventhrice:
This is tackling issue #580, to update the docker compose development documentation to make it more reliable for people to use and set up.


docker/docker-compose.dev.yml
```diff
@@ -56,13 +56,10 @@ services:
...
+      - ${DOCKER_DATA_DIR}/data:/data
       - ..:/app
     command:
      - /bin/sh
      - -c
      - "pnpm install --frozen-lockfile && pnpm run db:migrate"
 
-volumes:
-  meilisearch:
-  data:
```
MohamedBassem: Why though? Sounds easier to not have to mount a manual data dir?

docs/docs/07-Development/01-setup.md
```diff
@@ -60,4 +60,13 @@ In dev mode, opening and closing the plugin menu should reload the code.
 
 ## Docker Dev Env
 
-If the manual setup is too much hassle for you. You can use a docker based dev environment by running `docker compose -f docker/docker-compose.dev.yml up` in the root of the repo. This setup wasn't super reliable for me though.
+If the manual setup is too much hassle for you, you can use a docker based dev environment instead by following the instructions:
+
+- Clone the repo and navigate to the root of the repo.
+- Run `cp /docker/.env.sample /docker/.env`
+- Make the necessary changes to the .env file (setting the DOCKER_DATA_DIR, NEXTAUTH_SECRET, and MEILI_MASTER_KEY)
```
MohamedBassem, line 67: You don't really need to set nextauth_secret or even meili key during development. So I wouldn't worry too much about them

docs/docs/07-Development/01-setup.md
```diff
@@ -60,4 +60,13 @@ In dev mode, opening and closing the plugin menu should reload the code.
 
 ## Docker Dev Env
 
-If the manual setup is too much hassle for you. You can use a docker based dev environment by running `docker compose -f docker/docker-compose.dev.yml up` in the root of the repo. This setup wasn't super reliable for me though.
+If the manual setup is too much hassle for you, you can use a docker based dev environment instead by following the instructions:
+
+- Clone the repo and navigate to the root of the repo.
+- Run `cp /docker/.env.sample /docker/.env`
```
MohamedBassem, line 66: I don't think there's a need for an env file during development to be honest. Hoarder should work just fine without one.

MohamedBassem (changes requested):
Hey, thanks for taking the time and porting back the changes to the repo. I wonder, if you've just used `--build` from the get go, would it have solved all your problems without the rest of the changes?

docs/docs/07-Development/01-setup.md
Inventhrice: Wouldn't it be a good idea to set it up anyways, just to make sure that any development does not break functionality offered by nextauth or meilisearch?

docker/docker-compose.dev.yml
Inventhrice: This is mostly up to preference, I would much rather prefer having the /data directory to be persistent so that I can modify the sqlite db in case it is needed

docs/docs/07-Development/01-setup.md
Inventhrice: This ties back into the above comment: "Wouldn't it be a good idea to set it up anyways, just to make sure that any development does not break functionality offered by nextauth or meilisearch?"

docs/docs/07-Development/01-setup.md
CrypticC3s4r: I am too prefer having nextauth and meilisearch being ready while developing

MohamedBassem: Just to be clear. They will be ready and usable. The requirements of the variables being set are only when deployed to prod.

Inventhrice: oh gotcha because you want that stuff seeded with a random value

docs/docs/07-Development/01-setup.md
Inventhrice: coming back to this: wouldnt we need it for NEXTAUTH_URL and MEILISEARCH_URL?

Inventhrice:
@MohamedBassem Let me know if theres anything I can do to push this forward!

CRYPTO-GENIK:
The env file interferes with several homelab server distros

docker/docker-compose.dev.yml
iMuhammadMustafa: @MohamedBassem
If data dir is the actual directory which holds the bookmarks, it's important to manually mount it for backup consideration

thiswillbeyourgithub:
Maybe this should be closed now.

MohamedBassem:
Closing this in favor of #2149

PR #590, opened 2024-10-27
Faq

kamtschatka:
Added some initial FAQ regarding password reset and user management


kamtschatka:
Followup for: https://github.com/hoarder-app/hoarder/issues/544

MohamedBassem:
Thank you for starting this!

The way that the password is set, reminds me that we probably should be adding salts to our user passwords :) Also, I'm planning to eventually add an `admin-cli` that's only accessed from within the container that can alter the database and simplify those operations on the admin account.

PR #609, opened 2024-11-01
Download indicator

kamtschatka:
added 2 icons that show if a video and/or a full page archive has been downloaded for this bookmark

### DIFF
--- a/apps/web/components/dashboard/bookmarks/BookmarkActionBar.tsx
+++ b/apps/web/components/dashboard/bookmarks/BookmarkActionBar.tsx
@@ -2,10 +2,10 @@ import Link from "next/link";
 import { Button, buttonVariants } from "@/components/ui/button";
 import { toast } from "@/components/ui/use-toast";
 import { cn } from "@/lib/utils";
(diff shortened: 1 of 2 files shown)


MohamedBassem:
Between this PR and #608 I feel like the action bar is getting very crowded. I think the proper way to do this is to have an overlay on hover, and move all those icons there.

kamtschatka:
i am not sure how your screen resolution shows this, but on my screen there is plenty of space to show the icons.
Hover is definitely not a good solution as it is an accessibility nightmare, not usable on mobile devices and annoying to use on desktop devices.

MohamedBassem:
Screenshot 2024-12-20 at 5 23 00 PM

This is what it looks like on a 13" macbook screen (with arc's sidebar open). As you can see in the youtube bookmark, the URL is almost completely erased.

I'm currently experimenting with the hover idea, and your concern about accessibility is valid. Will check how other websites with similar functionalities handle this on touch screens.

For desktops, it's really useable IMO.

Screenshot 2024-12-20 at 5 26 25 PM

Issue #136, opened 2024-05-04
Need a way to change email address of account


scubanarc:
Either through app or CLI there needs to be a way to change the email address that you use to login.

MohamedBassem:
Fair. If this is urgent, you can do it from the sqlite database directly and things should work just fine.

scubanarc:
I did just that and it worked. Thanks.

hamoudydev:
> Fair. If this is urgent, you can do it from the sqlite database directly and things should work just fine.

how would one go about doing this?

mpowaa:
> > Fair. If this is urgent, you can do it from the sqlite database directly and things should work just fine.
> 
> how would one go about doing this?

By default, the Hoarder app’s db is stored in the data Docker volume, which resides at /var/lib/docker/volumes/hoarder_data/_data inside the Docker VM on macOS.

Lipown:
I need to do that, but have no idea how :/ I changed my email in Oauth provider and now I am not able to log in...

aha1215:
https://docs.karakeep.app/FAQ/

And instead of changing the password do this:

`update user set email='newemail@gmail.com' where email='oldemail@gmail.com';`

aha1215:
> I need to do that, but have no idea how :/ I changed my email in Oauth provider and now I am not able to log in...

https://docs.karakeep.app/FAQ/

And instead of changing the password do this:

`update user set email='newemail@gmail.com' where email='oldemail@gmail.com';`

mpowaa:
> > Fair. If this is urgent, you can do it from the sqlite database directly and things should work just fine.
> 
> how would one go about doing this?

@hamoudydev and @Lipown, were you able to update your email address?  If not, then post more details (like what step you're on, any error messages, or device you're using), the faster we can help you fix the issue.

borncrusader:
For the future, using `sqlite3` to update `email` works magically! Just tried it.

Lipown:
@mpowaa  well I did changed it back to original, than back to new and it is still working without changing anything.
@borncrusader I have to learn that...

kapuett:
> > I need to do that, but have no idea how :/ I changed my email in Oauth provider and now I am not able to log in...
> 
> https://docs.karakeep.app/FAQ/
> 
> And instead of changing the password do this:
> 
> `update user set email='newemail@gmail.com' where email='oldemail@gmail.com';`

~~This works but does not change the mail displayed in the profile, it still is the old one:~~

[...]

PR #611, opened 2024-11-02
Need a way to change email address of account #136

kamtschatka:
added a way to change the email address for local users added another check to not show the change password/email UI to users with OAuth


apps/web/components/settings/ChangeEmailAddress.tsx
```diff
@@ -0,0 +1,122 @@
...
+    },
+  });
+
+  const mutator = api.users.changeEmailAddress.useMutation({
+    onSuccess: () => {
+      toast({
+        description: "Email Address changed successfully. Logging out.",
+      });
+      form.reset();
+      signOut();
```
MohamedBassem, line 37: hmmm, this gets a bit tricky with JWTs. Yes here you'll sign out this session, but all other sessions will continue having a valid JWT which can get problematic. I really hate that next auth is forcing us to use JWTs :(

PR #615, opened 2024-11-05
Docker Installation Minor Language Update

LightningMarshal:
I changed some language in the docker installation documentation that I found confusing when I tried to follow the steps. I am one of vagabonds fleeing Omnivore as it is shutting down. I am not very familiar with docker, but I was able to get it to work. I believe these minor clarifications would improve the documentation.


MohamedBassem:
Thank you and sorry for the delay!

Issue #528, opened 2024-10-13
[Perf] Move away from the APIs that return all tags in one request


MohamedBassem:
People with thousands of tags are already feeling slowness in viewing their tags or searching through them. This issue is to:
1. Change the AllTags page to be paginated (with a search at top).
2. Change the tag editor to do async search requests to the backend instead of inline all the tags.

Arcturuss:
I feel there is something wrong with the page layout, if just a thousands (not even millions) of tags is enough to slow the browser. Same goes for the API request. Maybe it will be better in the long run to fix the underlying cause instead of hiding it with pagination?

petrm:
I just started to notice a huge slowdown with 5000 or so tags. Looking at the performance of the page, the browser actually spends most of the time waiting for /dashboard/tags, in my case over 9 seconds.
It doesn't look to me like a problem in the browser, but in the backend. Is there any flag that could be enabled to spit out the SQL query used to generate the tags on the page? 5000 rows is nothing - this has to return results in milliseconds.

petrm:
Well, I went to the drizzle docs, looks like there would be a way to show the query, but as far as I can see it would need to be explicitely added to the hoarder codebase. Also there is https://github.com/drizzle-team/drizzle-orm/issues/2605.

MohamedBassem:
it's unlikely that the problem is in the query itself. Last time we've seen this problem was because of the rendering (and potentially the server side rendering in this case).

petrm:
It can be that previously it was, but look at this:
![image](https://github.com/user-attachments/assets/f386fc59-3453-4800-bae9-5dd52c1d38fc)
Would you be able to share the queries?

MohamedBassem:
Most likely that's server side rendering taking the time. I can share the queries when I'm in front of the laptop.

kamtschatka:
```
{
  sql: 'select "id", "name", "createdAt", "userId", (select coalesce(json_group_array(json_array("attachedBy")), json_array()) as "data" from "tagsOnBookmarks" "bookmarkTags_tagsOnBookmarks" where "bookmarkTags_tagsOnBookmarks"."tagId" = "bookmarkTags"."id") as "tagsOnBookmarks" from "bookmarkTags" where "bookmarkTags"."userId" = ?',
  params: [ 't9uxin6b6q7fz1v72jm31nr6' ],
  typings: [ 'none' ]
}
```

petrm:
[...]

PR #618, opened 2024-11-05
[Perf] Move away from the APIs that return all tags in one request #528

kamtschatka:
improved the query speed by getting rid of the "with" part of the query and doing that manually


MohamedBassem:
I think I found the problem (https://github.com/hoarder-app/hoarder/blob/main/packages/db/schema.ts#L267). The index on `tagId` is misconfigured :) Let me send a proper fix instead.

PR #630, opened 2024-11-09
Update 10-import.md

bjornmonnens:
Importing through a bash script, necessary in case the errror "xargs: command line cannot be assembled, too long" occurs. Only tested on Mac, but should work on all environments


MohamedBassem:
Thanks!

PR #676, opened 2024-11-19
Add pikapods.com as install option

m3nu:
Many users have asked us to add Hoarder as supported app, which we did about 2 weeks ago. Right now there are about two dozen pods running Hoarder already, which tells me that the setup works for people.

To make this hosting option easier to find, I'm proposing to also it to the official docs, if the spirit of the project permits. I also tried to keep the text more factual and technical.

If there is interest in the future, we also offer a revenue sharing program for our app authors. Always open to talk about this.

Thanks for the consideration!


MohamedBassem:
Hey, happy to include this if it's clearly labeled that it's a paid hosting option. Let me modify the docs a bit and merge it.

m3nu:
Sure thing! Thanks for adding this.

PR #707, opened 2024-12-01
Add Linux install script and documentation

vhsdream:
Hi @MohamedBassem  - thank you so much for your help with getting the Proxmox LXC script going. Hopefully it's ready to be merged soon!

This PR adds an all-in-one bash installer/updater script that is very similar to the ones in their repo. It adds most of the suggestions that you had for that one, such as separating the data and config dirs from the install dir, but also adds support for Ubuntu 24.04 as well as Debian 12. I have tested it extensively on both platforms.

I've also written the documentation for the script because I figured why not?

Please let me know what you think and if this would be a good addition to installation methods for Hoarder.

### DIFF
--- a/docs/docs/02-Installation/06-debuntu.md
+++ b/docs/docs/02-Installation/06-debuntu.md
@@ -1,0 +1,76 @@
+# Debian 12/Ubuntu 24.04
+
+:::warning
+This script is a stripped-down version of those found in the [Proxmox Community Scripts](https://github.com/community-scripts/PromoxVE) repo. It has been adapted to work on baremetal Debian 12 or Ubuntu 24.04 installs **only**. Any other use is not supported and you use this script at your own risk.
+:::
+
+### Requirements
+
+- **Debian 12** (Buster) or
+- **Ubuntu 24.04** (Noble Numbat)
+
+The script will download and install all dependencies (except for Ollama), install Hoarder, do a basic configuration of Hoarder and Meilisearch (the search app used by Hoarder), and create and enable the systemd service files needed to run Hoarder on startup. Hoarder and Meilisearch are run in the context of their low-privilege user environments for more security.
+
+The script functions as an update script in addition to an installer. See **[Updating](#updating)**.
+
+### 1. Download the script from the [Hoarder repository](https://github.com/hoarder-app/hoarder/blob/main/hoarder-linux.sh).
+
+```
+wget https://raw.githubusercontent.com/hoarder-app/hoarder/main/hoarder-linux.sh
+```
+
+### 2. Make the script executable
+
+    ```shell
+    chmod +x ./hoarder-linux.sh
+    ```
+
+### 3. Run the script
+
+> This script must be run as `root`, or as a user with `sudo` privileges.
+
+    If this is a fresh install, then run the installer by using the following command:
+
+    ```shell
+    ./hoarder-linux.sh install
+    ```
+
+### 4. Create an account/sign in!
+
+    Then visit `http://localhost:3000` and you should be greated with the Sign In page.
+
+## Updating
+
+> This script must be run as `root`, or as a user with `sudo` privileges.
+
+    If Hoarder has previously been installed using this script, then run the updater like so:
+
+    ```shell
+    ./hoarder-linux.sh update
+    ```
+
+## Services and Ports
+
+`hoarder.target` includes 4 services: `meilisearch.service`, `hoarder-web.service`, `hoarder-workers.service`, `hoarder-browser.service`. 
+
+- `meilisearch.service`: Provides full-text search, Hoarder Workers service connects to it, uses port `7700` by default.
+
+- `hoarder-web.service`: Provides the hoarder web service, uses `3000` port by default.
+
+- `hoarder-workers.service`: Provides the hoarder workers service, no port.
+
+- `hoarder-browser.service`: Provides the headless browser service, uses `9222` port by default.
+
+## Configuration, ENV file, database locations
+
+During installation, the script created a configuration file for `meilisearch`, an `ENV` file for Hoarder, and located config paths and database paths separate from the installation path of Hoarder, so as to allow for easier updating. Their names/locations are as follows:
+
+- `/etc/meilisearch.toml` - a basic configuration for meilisearch, that contains configs for the database location, disabling analytics, and using a master key, which prevents unauthorized connections.
+- `/var/lib/meilisearch` - Meilisearch DB location.
+- `/etc/hoarder/hoarder.env` - The Hoarder `ENV` file. Edit this file to configure Hoarder beyond the default. The web service and the workers service need to be restarted after editing this file:
+    
+    ```shell
+    sudo systemctl restart hoarder-workers hoarder-web
+    ```
+- `/var/lib/hoarder` - The Hoarder database location. If you delete the contents of this folder you will lose all your data.
+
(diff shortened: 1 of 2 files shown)


hoarder-linux.sh
```diff
@@ -0,0 +1,274 @@
...
+  HOARDER_SECRET="$(openssl rand -base64 36 | cut -c1-24)"
+  cat <<EOF >${ENV_FILE}
+SERVER_VERSION=${RELEASE}
+NEXTAUTH_SECRET="${HOARDER_SECRET}"
+NEXTAUTH_URL="http://localhost:3000"
+DATA_DIR=${DATA_DIR}
+MEILI_ADDR="http://127.0.0.1:7700"
+MEILI_MASTER_KEY="${MASTER_KEY}"
+BROWSER_WEB_URL="http://127.0.0.1:9222"
+CRAWLER_VIDEO_DOWNLOAD=true
```
MohamedBassem: This is not enabled by default in hoarder docker, so I'd keep it still opt in here as well.

hoarder-linux.sh
```diff
@@ -0,0 +1,274 @@
...
+#CRAWLER_VIDEO_DOWNLOAD_MAX_SIZE=
+#OLLAMA_BASE_URL=
```
MohamedBassem: Maybe add the `OPENAI_API_KEY` as well?

hoarder-linux.sh
```diff
@@ -0,0 +1,274 @@
...
+#INFERENCE_IMAGE_MODEL=
+EOF
+  chmod 600 $ENV_FILE
+  echo ${RELEASE} > ${INSTALL_DIR}/version.txt
+  echo "Configuration complete" && sleep 1
+
+  echo "Creating users and modifying permissions..."
+  useradd -U -s /usr/sbin/nologin -r -m -d ${M_DATA_DIR} meilisearch
+  useradd -U -s /usr/sbin/nologin -r -M -d ${INSTALL_DIR} hoarder
+  chown meilisearch:meilisearch ${M_CONFIG_FILE}
```
MohamedBassem: What about `M_DATA_DIR`? Also let's maybe add quotes around the dir names as well
```suggestion
  chown meilisearch:meilisearch "${M_CONFIG_FILE}" "${M_DATA_DIR}"
```

hoarder-linux.sh
```diff
@@ -0,0 +1,274 @@
...
+
+    echo "Updating Hoarder to v${RELEASE}..." && sleep 1
+    sed -i "s|SERVER_VERSION=${PREV_RELEASE}|SERVER_VERSION=${RELEASE}|" ${ENV_FILE}
+    rm -R ${INSTALL_DIR}
+    cd /tmp
+    wget -q "https://github.com/hoarder-app/hoarder/archive/refs/tags/v${RELEASE}.zip"
+    unzip -q v${RELEASE}.zip
+    mv hoarder-${RELEASE} ${INSTALL_DIR}
+    cd ${INSTALL_DIR}/apps/web && pnpm i --frozen-lockfile &>/dev/null
+    pnpm exec next build --experimental-build-mode compile &>/dev/null
```
MohamedBassem: Are we missing

```
mv .next/standalone/apps/web/server.js .
```
here as well?

docs/docs/02-Installation/06-debuntu.md
```diff
@@ -0,0 +1,76 @@
...
+
+### 3. Run the script
+
+> This script must be run as `root`, or as a user with `sudo` privileges.
+
+    If this is a fresh install, then run the installer by using the following command:
+
+    ```shell
+    ./hoarder-linux.sh install
+    ```
```
MohamedBassem: Given that this is a one time script, can we just do `bash hoarder-linux.sh install` instead of first making it executable?

MohamedBassem (approved):
This is amazing, thanks a LOT for porting it and opening a PR! Looks good to me, left a couple of small comments.

hoarder-linux.sh
vhsdream: Just confirming, you're referring to the video download flag? If so then yes very good point, I think just commenting it out would suffice?

MohamedBassem: yeah exactly.

hoarder-linux.sh
vhsdream: When the meilisearch user is created, the data dir is also created and is owned by meilisearch, so chowning it is not needed.

hoarder-linux.sh
vhsdream: That would be consistent, yes; but now it's raising the question of why does the webUI start if it's not there?
image

(Note: that warning appears even if server.js is in apps/web, but the webUI works regardless.)

So this implies that the mv of server.js is superfluous I guess.

Out of curiosity, I tried what the warning suggested (I had to move `server.js` into that dir) and:
image

Finally, I left `server.js` where it was (after the update without it being moved) and ran `node .next/standalone/apps/web/server.js`:
image

However this leads to a server-side error:
image

So I'm thinking I can either add the mv of `server.js` like in the install function, or remove it completely and ignore the warning.

docs/docs/02-Installation/06-debuntu.md
vhsdream: Yes that works and removes the additional instruction; I'll amend the docs to remove the chmod bit.

hoarder-linux.sh
MohamedBassem: hmmm, let's remove the `mv` then. I do it in a completely different way in the docker container to be honest, but I think `pnpm start` for now directly sounds good enough.

hoarder-linux.sh
```diff
@@ -0,0 +1,274 @@
...
+db_path = "${M_DATA_DIR}/data"
+dump_dir = "${M_DATA_DIR}/dumps"
+snapshot_dir = "${M_DATA_DIR}/snapshots"
+no_analytics = true
+EOF
+  chmod 600 $M_CONFIG_FILE
+
+  HOARDER_SECRET="$(openssl rand -base64 36 | cut -c1-24)"
+  cat <<EOF >${ENV_FILE}
+SERVER_VERSION=${RELEASE}
```
MohamedBassem, line 90: I noticed that we're missing `NODE_ENV=production` here as well.

hoarder-linux.sh
vhsdream: Apologies, my previous post was a bit misleading; `root` did not have the env which was causing the error. I still think leaving it as is and removing the `mv` command is ok, but I just tested by amending the service file to use `ExecStart=node .next/standalone/apps/web/server.js`  and that works just fine.

MohamedBassem:
I'm curious, is there any reason why we're swallowing the output of all commands? While the installation is happening, it seems to me as if process is stuck. Maybe that's the convention in proxmox scripts, but do we need to do that here? I know it'll probably make it harder to see the output messages, but at least it shows the progress of the installation?

vhsdream:
> I'm curious, is there any reason why we're swallowing the output of all commands? While the installation is happening, it seems to me as if process is stuck. Maybe that's the convention in proxmox scripts, but do we need to do that here? I know it'll probably make it harder to see the output messages, but at least it shows the progress of the installation?

That's the convention in the Proxmox scripts, so it looks all nice, but they have a flag to enable verbose mode. I lack the finesse to make a progress indicator, at least within a reasonable amount of time 😄  If you don't mind the script having all the output and looking quite busy as it runs, I'm happy to stop supressing the output.

MohamedBassem:
@vhsdream Yeah, I'm trying the script now and trying to figure out what it's doing. I don't think you need to do progress bars, I think just removing the suppression would be good enough.

MohamedBassem:
Thanks again @vhsdream!

xxjoe2:
hi @vhsdream 

may i ask how did you install hoarder with the script on ubuntu 24 successfully?

it produces a lot of errors during the installation which i have already mentioned part of them at https://github.com/hoarder-app/hoarder/issues/933 

thanks

PR #714, opened 2024-12-03
Trim whitespace from email on signin page

sweepies:
Sometimes when pasting or using a password manager, an extra space is inserted after an email address. I noticed this caused a sign in failure.

Whitespace isn't valid here, so it's safe to just trim this off.

### DIFF
--- a/apps/mobile/app/signin.tsx
+++ b/apps/mobile/app/signin.tsx
@@ -121,7 +121,7 @@ export default function Signin() {
                   placeholder="Email"
                   keyboardType="email-address"
                   autoCapitalize="none"
-                  value={usernameFormData.email}
+                  value={usernameFormData.email.trim()}
                   onChangeText={(e) =>
                     setUserNameFormData((s) => ({ ...s, email: e }))
                   }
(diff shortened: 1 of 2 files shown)


apps/mobile/app/signin.tsx
```diff
@@ -121,7 +121,7 @@ export default function Signin() {
                   placeholder="Email"
                   keyboardType="email-address"
                   autoCapitalize="none"
-                  value={usernameFormData.email}
+                  value={usernameFormData.email.trim()}
                   onChangeText={(e) =>
                     setUserNameFormData((s) => ({ ...s, email: e }))
```
MohamedBassem: This needs to happen in the setter otherwise it won't make the desired effect

```suggestion
                  value={usernameFormData.email}
                  onChangeText={(e) =>
                    setUserNameFormData((s) => ({ ...s, email: e.trim() }))
```

MohamedBassem (changes requested):
Hey, thanks for the PR. This makes sense to me. Left a small comment on the change in the mobile app.

sweepies:
Ah right, missed that.

MohamedBassem:
Sorry for the delay. Thanks again for the fix!

Issue #701, opened 2024-11-28
Improve note support : WYSIWYG markdown


Tukks:
### Describe the feature you'd like

I made a similar app here https://github.com/Tukks/cogito but I don't have enough time to make new feature, and now I discovered Hoarder who is what I wanted for Cogito.
I think, the note part would need a little upgrade.

For the note-taking and editing part, I used the stackover flow markdown editor : https://editor.stackoverflow.design/ who is complete and very well-designed.

Here's how is integrated in cogito : 
![image](https://github.com/user-attachments/assets/8a3c001c-4bc9-4b92-8bf1-b00874564799)
![image](https://github.com/user-attachments/assets/139c1fb4-f089-430d-8743-46aca9ea4975)

### Describe the benefits this would bring to existing Hoarder users

add visual shortcut to bold/italic etc.
Support code
Support image
full-featured note-taking app
WYSIWYG

Upgrade : when we open a note card, open in Edit mode or add a button to edit in the big view (now we need to go : option > Edit)
![image](https://github.com/user-attachments/assets/b3a50c56-2293-4827-bbfb-3436e83ff66d)

### Can the goal of this request already be achieved via other means?

We can note-take today but some features are not implemented like code support or visual shortcut

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### Additional context

I can try to work on that if validated
I'm open to other editor BTW like tiptap/Lexical (https://lexical.dev/)/mdx (https://mdxeditor.dev/editor/demo)
We can add a toolbar to the existing react-markdown too.

MohamedBassem:
Stacks Editor looks indeed very cool, and I'be be up to adding support for it in hoarder.

> I can try to work on that if validated

I'd be more than happy to merge it if you send a PR for it. You can join us in the #development channel in discord in case you need any help with developing in hoarder as well.

Gnosnay:
is that possible to add support for `Text` bookmark also?

Tukks:
> is that possible to add support for `Text` bookmark also?
> 
> for now it seems that hoarder only supports plain text

What is text bookmark? 
Hoarder actually support Markdown, but not in a wysiwig way.
If you put 

# title 1 

For example, you will see.

@MohamedBassem perfect, I will give it a try.

PR #715, opened 2024-12-03
#701 Improve note support : WYSIWYG markdown

Tukks:
First implementation with a wysiwyg markdown editor. Update:
- Add Lexical markdown editor
- consistent rendering between card and preview
- removed edit modal, replaced by preview with save action
- simple markdown shortcut: underline, bold, italic etc...


Tukks:
This is a first implementation, I will keep it simple for the moment.
Later I will add codeblock, Headings, and other format who I think can be interesting in Hoarder.  

here's a full implementation example : https://playground.lexical.dev/?showTreeView=false&isAutocomplete=true

Let me know if everything all right in the PR.

Tukks:
Added useTranslation to all string

MohamedBassem:
Ok, I spent a lot of time today reading about lexical and understanding how it works. It's pretty cool and thanks for introducing it into the codebase. I've pushed a bunch of modifications/fixes to the PR that I hope you don't mind:

1. Dropped alignment and underline support as those are not supported in markdown.
2. Dropped the history undo/redo buttons for now just to simplify things a bit. Kept the history plugin though as the keyboard shortcuts still work.
3. Added markdown shortcuts such that the markdown is rendered as you type it.
4. Fixed a bug in the toolbar not correctly reflecting highlights and code.

Now, to merge this PR. There's a couple of things that we need to do:

1. I don't like the flicker that happens in the bookmark cards in homepage on load. Seems like lexical only works on the client side (https://github.com/facebook/lexical/issues/4960). Which means that I think we should continue using `react-markdown` in the bookmark cards to avoid this flicker. Unfortunate but can't find a way around it.
2. We don't actually need to use a theme`. Just adding `prose dark:prose-invert prose-p:m-0` to the `ContentEditable` gets us most of the way there. The benefit of that is that it'll make the styling look similar to `react-markdown` and to what it'll look like on the mobile app.
3. Not a blocker, but I think we should move the `Save` button to the toolbar.
4. We should add an escape hatch in the editor to toggle editing raw markdown so that people don't get frustrated from fighting with the editor.

Tukks:
Hello,

I saw you removed the UpdateMarkdownPlugin, it was needed when you updated the markdown in the editor to make change appear in the grid. I will not be needed if we use react-markdown on the grid.



> * Dropped alignment and underline support as those are not supported in markdown.

You're right.
> * Dropped the history undo/redo buttons for now just to simplify things a bit. Kept the history plugin though as the keyboard shortcuts still work.

No problem, I understand.
> * Added markdown shortcuts such that the markdown is rendered as you type it.

Yes, I will add a little tooltip helper so


> 1. I don't like the flicker that happens in the bookmark cards in homepage on load. Seems like lexical only works on the client side
 
Yes, it bother me too, I didn't know how we can find a way around it.
I will give it a try with react-markdown and prose dark:prose-invert prose-p:m-0 (I did'nt understand how this can work, but let see)

> 3. Not a blocker, but I think we should move the Save button to the toolbar.

Why not, I have no opinion on that.

> 4. We should add an escape hatch in the editor to toggle editing raw markdown so that people don't get frustrated from fighting with the editor.

Definitely yes, [you can see a button in the lexical playground here](https://playground.lexical.dev/#doc=H4sIAAAAAAAACu1ZzW4bNxB-lenm4DbQj2OnDaxDEaAB0rQJkMJGiyIIEC452iXE5ajkrGQhMFD0FXrKKe_WJ-gjdEhZjhz0YKyaLmTnIIg7Sw5nvh1-Qw7fFmgsUzhlxVhM3haBiNO_rq0zAX0xefXRg0FW1hWTw0ExpdAozs2GjIwvfJK4YlBEXrkkkCbjufQp7qXmap6EWjoPa1vVTn4s8gWGaEn0PxgUV_Kzded56zW3itP7i0H3-eEXdJoaBCbgGmHu1KoK1HrT0TC2LFMkky5HO-uxDKhm1wbeqENXp77taHvpSM9-a4lxN0yfedAqIqyohSUGhCV5g8H6Cpa14gx06ZSeQUnnsBEQMzVgI_z1-59g-SBmscGyrWBhcTmAWNMyKUly3QaJPYaYYhRomoXrsB3BrzKxVh6Mjap0KNqgXME8YIxpPPncOyKzPEbQ5DmQA-u3LBk6nHJSLF4EiDogelDeSKRUlduY8cG8jbrRjcDvN0DOrkV6wlyJJw0B-oUN5JuEbNlax7C0XMONXepq0JvHDs-tVm4sfmp-0zGAc-fo7XyOvFMIj-AsrEBMSN9ZwuKTA3D_fs9EGDMJpnZX6D1L1OxkQ-8g_E_R3rebxk6nmNizz2_d96de9477QNcv1EyyVRvyLkXXKImT2nXWXKhgqY3C5m1lfdykMCZypdrKhMpFglZy8r1axZqVZD0K8HiYmF4MjGnEF3sAxbOcjw8MODvLcEytJLCERkOCjypT8_k6lwzyBkS8n-yBY8OO68HZuNtChJ9ttOtg-uTE96qjk21wO_l4GRCyGS3FV-yR9l5_2RMENfM8Tsbjy33WyOBi3JMpX_U0b6J8MKTbRHo5g-T9fCYO69eqRPgfZoT9I4PvrmWXNHk6LaUT0K0lh6eWv29LCDinmI6Pq7tMD5VsgNtypKkZT5XGkmi2oYw7xhZ3mgZefigOZA5IW8gShUCT5NYyQS2Hos-L_18X_5gD4rhR1o_nSs9UhXHzarhzzfQzRewhRfxA9pZvDJ7YqCkYOMUg09xlYjBrJDIzWL-QI9T4x-bpw-VP3j85uWOLPp8ZdLrGyMXCXG1B1exDHem5iuxWAzkGH4S0ufeRlRcJKGNSoVsTOfC4hCkqbgPG9a2YjVuXBSM4JWiuClK5wpLPDGW6z6nzbU-NPr_wMi1UyKASYF4j_P3-3R83ROr1oDA2oM6VPkGD0wrcuJucst6khZVc31L30dd3yletZCt526gwM7T0RRfV-Qb0mn0XSXnkU7VAI8-Pjo8ffX108s3Rw-MTwV6IUadxL7dz49Xo4nB09GB0WFz8A0SiLXpqHQAA)

Tukks:
Hello,
Make some change : 
- added ListPlugin because if absent, there's a bug where  you can't escape a list with enter + enter
-  added codeblock plugin
- added prose dark:prose-invert prose-p:m-0 like you said (there's room for improvement I think, don't took the time too deep dive in) and removed theme
- Added a switch to show raw markdown
- Added back the react markdown for card (SSR)

Todo ? 
- Save button in toolbar ?
- Some testing

Tukks:
Hello @MohamedBassem, 

Removing the theme for the code highlight make the code not so clear on the lexical part.
The react markdown 'reader' use a custom components for code (prism) and highlight color.

Tried to replace the codeNode in lexical with the prism components, but no luck.

I think we can add back only the code part from theme.ts and use prose from tailwind for the rest.

What do you think? 
Hope this is clear.

MohamedBassem:
@Tukks Sure, let's do that 👍

Tukks:
@MohamedBassem theme.ts is back.

let me know if evrythings is okay

MohamedBassem:
@Tukks I made the following changes:

1. Given that the editor is new, I didn't want to use it in the bookmark preview for now as there's no escape hatch there if something goes wrong. So I moved the editor back to the edit bookmark option. If it proves to be reliable enough, we can add it back to the bookmark preview page.
2. I didn't like the hack of handling raw markdown (aka using codeblocks). It wasn't very intuitive and was causing a bunch of trouble for me. So for raw markdown, I changed the code to use plain text editor instead of the rich one.
3. I moved the save button to the toolbar as we agreed.
4. I moved the markdown component out of the `ui` directory as this directory is meant to be generic, and that component was not (e.g. it had a dependency on the update bookmark hook).

I think this PR is now ready to get merged. Let me know what you think.

Tukks:
> 1. Given that the editor is new, I didn't want to use it in the bookmark preview for now as there's no escape hatch there if something goes wrong. So I moved the editor back to the edit bookmark option. If it proves to be reliable enough, we can add it back to the bookmark preview page.

I understand, let it play secure for now.


> 2\. I didn't like the hack of handling raw markdown (aka using codeblocks). It wasn't very intuitive and was causing a bunch of trouble for me. So for raw markdown, I changed the code to use plain text editor instead of the rich one.

I took the codeblock things from lexical playground, but you're right, I prefer your way.

> 3\. I moved the save button to the toolbar as we agreed.

Yes perfect

> 4\. I moved the markdown component out of the `ui` directory as this directory is meant to be generic, and that component was not (e.g. it had a dependency on the update bookmark hook).

ok

The PR can be merged for me too. 
Thank for your help.

PR #748, opened 2024-12-22
deps: Upgrade to Nextjs 15 and React 19

MohamedBassem:


MohamedBassem:
Hmmm, having multiple versions in react seems like is going to be painful. We can't upgrade the docs or the mobile apps just yet.

PR #749, opened 2024-12-22
Fix broken link in Debian/Ubuntu install doc

vhsdream:
Unfortunately I missed this mistake back when it was added.


MohamedBassem (approved):
Ah good catch. Mind fixing it in `docs/versioned_docs/version-v0.20.0/02-Installation/06-debuntu.md` as well so that it gets fixed in the 0.20 docs?

PR #762, opened 2024-12-25
Update 08-security-considerations.md

mark-pitblado:
Add missing "t" to "untrusted"


MohamedBassem (changes requested):
Good catch! However, this needs to be edited in `docs/docs/08-security-considerations.md` instead of the versioned ones. Otherwise the typo will come back in the next version.

mark-pitblado:
Ah I see! I have made the same change in `docs/docs`. I assume that leaving the change in the first file is harmless since a) it is a correction and b) that it will get overwritten anyways.

Let me know if there is anything else, and apologies for missing the versioning scheme

MohamedBassem:
Thank you!

PR #781, opened 2024-12-28
Added Turkish translation.

bcanata:
- Added Turkish translations for the application's interface in apps/web/lib/i18n/locales/tr/translation.json.
- Updated packages/shared/langs.ts to include Turkish ("tr") in the language mappings.
- No deletions were made, and the changes ensure full localization support for Turkish-speaking users.


MohamedBassem:
Thank you!

Issue #776, opened 2024-12-27
Confirmation before deleting item


bverkron:
### Describe the feature you'd like

Seems there is currently no confirmation required before deleting an item. This could be catastrophic if an item is deleted accidentally and the original source of the item is no longer available on the web and can't be re-added.

### Describe the benefits this would bring to existing Hoarder users

Avoids accidental and permanent deletion of an item that can no longer be found online 

### Can the goal of this request already be achieved via other means?

Not AFAIK.

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

lexafaxine:
Hello @MohamedBassem , I just found this repo and very interested in its concept. I am going to add a `DeleteBookmarkConfirmationDialog` component in the bookmarks components to handle this. I could open a PR in 24h, could you please assign this issue to me? Thanks!

MohamedBassem:
@lexafaxine thanks for picking this up, looking forward to the PR!

PR #787, opened 2024-12-30
improvement: add confirmation before deleting item

lexafaxine:
PR for #776 

For mobile,
added an alert when deleting bookmark

For web,
added an confirmation dialog when deleting bookmark

Proof:
https://drive.google.com/drive/folders/1067NHRKKIoip3vZRfgaJnbxBj1XHf5oc?usp=sharing


apps/web/components/dashboard/preview/ActionBar.tsx
```diff
@@ -100,14 +91,18 @@ export default function ActionBar({ bookmark }: { bookmark: ZBookmark }) {
...
           <ActionButton
-            loading={pendingDeletion}
+            loading={isPendingDelete}
             className="size-14 rounded-full bg-background"
             variant="none"
-            onClick={() => {
-              deleteBookmark({ bookmarkId: bookmark.id });
-            }}
+            onClick={() => setDeleteBookmarkDialogOpen(true)}
           >
```
MohamedBassem: Given that this button no longer itself takes an action, we should switch it to just a normal button instead of `ActionButton`. That way, you also won't need to propagate the pending state back to this component.

pnpm-lock.yaml
MohamedBassem, line 1: Something weird is happening in the lockfile. I assume you might have not used the same `pnpm` version defined in the `package.json`? Anyways, given that you're not changing any deps, this file should probably be reverted.

apps/web/components/dashboard/bookmarks/BookmarkOptions.tsx
```diff
@@ -118,6 +112,11 @@ export default function BookmarkOptions({ bookmark }: { bookmark: ZBookmark }) {
         setOpen={setTextEditorOpen}
       />
       <DropdownMenu>
+        <DeleteBookmarkConfirmationDialog
+          bookmark={bookmark}
+          open={deleteBookmarkDialogOpen}
+          setOpen={setDeleteBookmarkDialogOpen}
+        />
```
MohamedBassem: I think this probably should be moved out of the dropdown menu. Maybe move it beside the other modals on line 108?

apps/web/components/dashboard/bookmarks/DeleteBookmarkConfirmationDialog.tsx
```diff
@@ -0,0 +1,70 @@
...
+  useEffect(() => {
+    onPendingChange?.(isPending);
+  }, [isPending, onPendingChange]);
+
+  return (
+    <ActionConfirmingDialog
+      open={open}
+      setOpen={setOpen}
+      title={`Delete ${bookmark.title ?? "Bookmark"}?`}
+      description={`Are you sure you want to delete this bookmark?`}
```
MohamedBassem: Can you add this to the translation files in `apps/web/lib/i18n/locales/en.json`? So that people can also get a chance to translate it?

MohamedBassem (changes requested):
Thanks for the PR! I've left some small comments and then it'll be good to go!

apps/web/components/dashboard/bookmarks/DeleteBookmarkConfirmationDialog.tsx
lexafaxine: Sure, added it in [add delete bookmark confirmation dialog](https://github.com/hoarder-app/hoarder/pull/787/commits/5460f44147e30c29a9b6b2d4eff5bfb82efa7a7b)

apps/web/components/dashboard/bookmarks/BookmarkOptions.tsx
lexafaxine: fixed in [add delete bookmark confirmation dialog](https://github.com/hoarder-app/hoarder/pull/787/commits/5460f44147e30c29a9b6b2d4eff5bfb82efa7a7b)

apps/web/components/dashboard/preview/ActionBar.tsx
lexafaxine: Fixed in [add delete bookmark confirmation dialog](https://github.com/hoarder-app/hoarder/pull/787/commits/5460f44147e30c29a9b6b2d4eff5bfb82efa7a7b)

MohamedBassem:
Thank you and congrats on the first contribution in Hoarder!

PR #790, opened 2024-12-31
feat: show createdAt year in card footer when created over a year ago

jdhartley:
I imported a bunch of old saves from Pocket and found it hard to tell how old the saves were when browsing

I added the year to the Bookmark Card footer date format string when the bookmark is created over a year ago. It doesn't take up too much space, hope nobody minds the few characters

Happy to move code around or take another stab at this. Thanks!

<details>
<summary>screenshots showing grid and compact bookmark cards</summary>
grid preview<br>compact preview
</details>


MohamedBassem:
This makes sense, thanks a lot for the fix!

PR #798, opened 2024-12-31
Fix 'Fetch Failed' for next-auth when using OAuth behind proxy

rfrank:
Without the NEXTAUTH_URL_INTERNAL env variable pointing to 'http://localhost:DEFAULT_HOARDER_PORT', Traefik (and possibly others) proxy report a fetch failed and refuse to show OAuth login button or allow OAuth login.

### DIFF
--- a/docs/docs/03-configuration.md
+++ b/docs/docs/03-configuration.md
@@ -22,7 +22,7 @@ Only OIDC compliant OAuth providers are supported! For information on how to set
 :::
 
 :::info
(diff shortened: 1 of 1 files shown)


coversheets:
👍 - also happening to me with Traefik + Authentik but I checked the [NextAuth.js docs](https://next-auth.js.org/configuration/options) and this worked.  This should definitely be in the documentation and the [Authentik documentation for Hoarder](https://docs.goauthentik.io/integrations/services/hoarder/).

MohamedBassem:
hmmm, it kinda seem like this is something that can be baked in the docker image itself instead of asking the users to set it up. Its value will never change, so we probably can add it here (https://github.com/hoarder-app/hoarder/blob/main/docker/Dockerfile#L43) as `ENV NEXTAUTH_URL_INTERNAL http://localhost:3000`. What do you think?

MohamedBassem:
Given that multiple people has been facing this, I ended up baking it in the image in c5298cf. Thanks for pointing out this problem and the idea of having a default for it!

PR #812, opened 2025-01-03
feat: Add Bookmark Sorting Feature

dakshpareek:
This PR adds the ability to sort bookmarks by creation date in ascending or descending order.

## Changes
- Add sort order toggle in GlobalActions component
- Implement ascending/descending sort functionality
- Update translations for sorting feature in all supported languages
- Add sort order icons and dropdown menu
- Maintain sort preference in URL parameters

## Screenshots
ScreenRecording2025-01-03at1 50 11PM-ezgif com-video-to-gif-converter

## Testing Done
- [x] Verified sort order works in both directions
- [x] Confirmed sort preference persists across page reloads
- [x] Tested with different layouts (grid, list, etc.)
- [x] Verified translations work correctly

## Notes
- Sort preference is maintained in URL params for better sharing and persistence
- Default sort order is newest first (descending)
- Note: Sort order resets to default on page refresh, server-side persistence can be implemented in future enhancement


kamtschatka:
there is the new feature of smart lists/filtering based on queries. My assumption was, that sorting would be part of the query language in the future, since your solution will not work with pagination i assume?

dakshpareek:
I do not have knowledge about smart lists/filtering. Is this upcoming feature?
My solution will work with pagination I assume.

MohamedBassem:
Hey, thanks for the PR! My plan (still haven't given it a lot of thought) is to keep sorting and smart search separate (the way this PR introduces it) because sorting is needed even when smart searching is not user. Will hopefully be able to review this PR over the weekend.

pnpm-lock.yaml
MohamedBassem, line 1: I think you might be using a different pnpm version which introduced a lot of changes to the lock file. Those will probably need to be reverted.

packages/shared/types/bookmarks.ts
```diff
@@ -161,6 +167,8 @@ export const zGetBookmarksRequestSchema = z.object({
   // The value is currently not being used, but keeping it so that client can still set it to true for older
   // servers.
   useCursorV2: z.boolean().optional(),
+  sortBy: zSortBy.optional().default("createdAt"),
```
MohamedBassem: Given that this is currently unused, should we just drop it for now to avoid confusion?

packages/shared/types/bookmarks.ts
```diff
@@ -12,6 +12,12 @@ export const enum BookmarkTypes {
   UNKNOWN = "unknown",
 }
 
+export const zSortOrder = z.enum(["asc", "desc"]);
+export type SortOrder = z.infer<typeof zSortOrder>;
```
MohamedBassem: Please follow the same style in the rest of the file. All zod inferred type start with the `Z` prefix.

packages/shared/types/bookmarks.ts
```diff
@@ -12,6 +12,12 @@ export const enum BookmarkTypes {
...
+
+export const zSortBy = z.enum(["createdAt"]);
+export type SortBy = z.infer<typeof zSortBy>;
```
MohamedBassem: Same comment about the prefix.

apps/web/components/dashboard/SortOrderToggle.tsx
```diff
@@ -0,0 +1,64 @@
...
+        >
+          <SortDesc size={18} />
+        </ButtonWithTooltip>
+      </DropdownMenuTrigger>
+      <DropdownMenuContent className="w-fit">
+        <DropdownMenuItem
+          className="cursor-pointer justify-between"
+          onClick={() => updateSort("desc")}
+        >
+          <span>{t("actions.sort.newest_first")}</span>
```
MohamedBassem: Let's include some lucide sort icons here to make it look nicer: https://lucide.dev/icons/?search=sort

packages/trpc/routers/bookmarks.ts
```diff
@@ -721,7 +732,12 @@ export const bookmarksAppRouter = router({
             ),
           )
           .limit(input.limit + 1)
-          .orderBy(desc(bookmarks.createdAt), desc(bookmarks.id)),
+          .orderBy(
+            input.sortOrder === "asc"
+              ? asc(bookmarks.createdAt)
+              : desc(bookmarks.createdAt),
+            desc(bookmarks.id),
+          ),
```
MohamedBassem, line 774: Given that this order button is global, we'll need to modify the `searchBookmarks` endpoint as well, and modify `apps/web/lib/hooks/bookmark-search.ts` to pass the sorting order to it as well.

apps/web/components/dashboard/SortOrderToggle.tsx
```diff
@@ -0,0 +1,64 @@
...
+    // invalidate the query
+    await apiUtils.bookmarks.getBookmarks.invalidate();
+
+    const params = new URLSearchParams(searchParams.toString());
+    params.set("sort", newSort);
+    router.replace(`?${params.toString()}`);
+
+    // Force a refetch after URL update
+    await apiUtils.bookmarks.getBookmarks.refetch();
+  };
```
MohamedBassem: This and the part that reads the sort order sounds like something that should be extracted into its own shared hook. Specially that as I mentioned later, you'll have to use it for the search endpoint.

Also, the mobile app will eventually have search as well, so I feel like we can make this a zustand store (instead of using search params), and make the hook generic across the mobile and the web. You don't have to do the mobile app part though (I can take care of it).

MohamedBassem (changes requested):
Hey, thanks for the PR! Overall, the approach looks good to me. Left some comments, that we can merge the PR after they get addressed.

dakshpareek:
Hi @MohamedBassem, thanks for reviewing PR.
I will look into these comments when I reach back home on Thursday after my vacation.

pnpm-lock.yaml
dakshpareek: sure.

packages/shared/types/bookmarks.ts
dakshpareek: yes, removed it.

packages/shared/types/bookmarks.ts
dakshpareek: Done

packages/shared/types/bookmarks.ts
dakshpareek: Removed this as well.

packages/trpc/routers/bookmarks.ts
dakshpareek: Yes, search will also support sorting now.

apps/web/components/dashboard/SortOrderToggle.tsx
dakshpareek: Extracted it into it's own store. Take a look.

apps/web/components/dashboard/SortOrderToggle.tsx
dakshpareek: Added icon here. Let me know if you think some other icon should be there.

dakshpareek:
Hello @MohamedBassem,
I have adjusted sorting related code according to your suggestions. Please take a look when you have time and let me know what do you think. 

Thanks

MohamedBassem:
Thank you!

dakshpareek:
Hey @MohamedBassem btw when is next release planned?
Just wanted to know when I can view latest changed on my hoarder account. 😄

MohamedBassem:
@dakshpareek Given that the last release was last week, I think it'll be sometime until the next one. Maybe 3 weeks or so until we accumulate enough features for the release. There hasn't been a lot of changes since the release, so the nightly build should be stable if you want to switch to it today and have your feature available there today :D

PR #816, opened 2025-01-03
Fix openAI pricing link

JackBailey:
This link currently redirects to the ChatGPT subscription-based pricing page, I've updated it with the URL for the model pricing page.


MohamedBassem:
Thanks for the fix! However, you need to modify it in `docs/docs/06-openai.md` instead otherwise, the wrong URL will be back in the next docs release.

JackBailey:
Ahh that makes sense, don't know how I missed that 😆 , should be all good now

MohamedBassem:
Thanks!

PR #834, opened 2025-01-05
WIP: Bookmark embeddings

MohamedBassem:


thiswillbeyourgithub:
Just linking #1315  here as I think it might be better to make it part of the first release that contains embeddings, otherwise we'll have to deal with the issue of resetting all the embeddings when switching to binary or not.

thiswillbeyourgithub:
I hope you won't mind but I was wondering what were the blockers for this much awaited feature. What is the state of the code, what are questions left to answer, what can the community do to help, etc.

Because it's a major change, I'm afraid no one other than the maintainer would want to tackle it and so I think communication here would go a long way to make us help you without duplicating too much effort.

thiswillbeyourgithub:
I was suddenly reminded of that prominent quote in the readme:

> This app is under heavy development and it's far from stable.

So my opinion is now that, as long as there is no data loss (ever!), this disclaimer grants you the moral right to move faster on those roll out situations.

- ollama users will figure it out
- us contributors will help with PRs left and right to iron out the time consuming kinks (docs, guides, UI, perf, etc)
- this kind of major feature that touches both frontend and backend seems (to me) at this stage best handled by the maintainer, saving coordination overhead
- with this disclaimer: the goal is to move fast to reach quickly the "stable" state, as a lot of users are probably hesitant to join as long as the disclaimer is in place

Just my 2 cents on priorities of course. I'm aware you're very busy and am sure you'll do what's best for us all.

paul-tharun:
We are using the same OpenAI api-key and base url for both inference and embedding. I think have separate config for each of them would be better, as it allows for using different providers for each of these tasks. In addition, this also would remove the limitation of having to use a provider who supports both embeddings and chat.

thiswillbeyourgithub:
> We are using the same OpenAI api-key and base url for both inference and embedding. I think have separate config for each of them would be better, as it allows for using different providers for each of these tasks. In addition, this also would remove the limitation of having to use a provider who supports both embeddings and chat.

I sort of disagree as for more cutstomization you should instead  run a [litellm](https://docs.litellm.ai) instance

paul-tharun:
> > We are using the same OpenAI api-key and base url for both inference and embedding. I think have separate config for each of them would be better, as it allows for using different providers for each of these tasks. In addition, this also would remove the limitation of having to use a provider who supports both embeddings and chat.
> 
> I sort of disagree as for more cutstomization you should instead run a [litellm](https://docs.litellm.ai) instance

Litellm is one of the llm gateway providers that support embeddings , there are others that do not support embeddings but solely focus on llm inference (openrouter, cloudflare). Adding this would be good separation of concerns imo, if needed we can default to the inference parameters if the embedding params are not set.

MohamedBassem:
This PR is replaced with: https://github.com/karakeep-app/karakeep/compare/main...claude/add-embeddings-support-h2swV

PR #852, opened 2025-01-09
feat(webhook): Implement webhook functionality for bookmark events

hanguofeng:
- Added WebhookWorker to handle webhook requests.
- Integrated webhook triggering in crawlerWorker after video processing.
- Updated main worker initialization to include WebhookWorker.
- Enhanced configuration to support webhook URLs, token, and timeout.
- Documented webhook configuration options in the documentation.
- Introduced zWebhookRequestSchema for validating webhook requests.

### DIFF
--- a/apps/workers/webhookWorker.ts
+++ b/apps/workers/webhookWorker.ts
@@ -1,12 +1,16 @@
+import { eq } from "drizzle-orm";
 import { DequeuedJob, Runner } from "liteque";
+import fetch from "node-fetch";
+
 import { db } from "@hoarder/db";
 import { bookmarks } from "@hoarder/db/schema";
 import serverConfig from "@hoarder/shared/config";
 import logger from "@hoarder/shared/logger";
-import { WebhookQueue, zWebhookRequestSchema, ZWebhookRequest } from "@hoarder/shared/queues";
-import { eq } from "drizzle-orm";
-import fetch from 'node-fetch';
-import { Response } from 'node-fetch';
+import {
+  WebhookQueue,
+  ZWebhookRequest,
+  zWebhookRequestSchema,
+} from "@hoarder/shared/queues";
 
 export class WebhookWorker {
   static build() {
@@ -18,25 +22,30 @@ export class WebhookWorker {
         onComplete: async (job) => {
           const jobId = job.id;
           logger.info(`[webhook][${jobId}] Completed successfully`);
+          return Promise.resolve();
         },
         onError: async (job) => {
           const jobId = job.id;
           logger.error(
             `[webhook][${jobId}] webhook job failed: ${job.error}\n${job.error.stack}`,
-          )
+          );
+          return Promise.resolve();
         },
       },
       {
         concurrency: 1,
         pollIntervalMs: 1000,
-        timeoutSecs: serverConfig.webhook.timeout / 1000 * (serverConfig.webhook.retryTimes + 1) * serverConfig.webhook.urls.length + 1,  //consider retry times, urls, and timeout and add 1 second for other stuff
+        timeoutSecs:
+          serverConfig.webhook.timeoutSec *
+            (serverConfig.webhook.retryTimes + 1) +
+          1, //consider retry times, and timeout and add 1 second for other stuff
+        validator: zWebhookRequestSchema,
       },
     );
 
     return worker;
   }
 }
-
 
 async function fetchBookmark(linkId: string) {
   return await db.query.bookmarks.findFirst({
@@ -52,17 +61,16 @@ async function fetchBookmark(linkId: string) {
 async function runWebhook(job: DequeuedJob<ZWebhookRequest>) {
   const jobId = job.id;
   const webhookUrls = serverConfig.webhook.urls;
-  const webhookToken = serverConfig.webhook.token;
-  const webhookTimeout = serverConfig.webhook.timeout;
-
-  const request = zWebhookRequestSchema.safeParse(job.data);
-  if (!request.success) {
-    throw new Error(
-      `[webhook][${jobId}] Got malformed job request: ${request.error.toString()}`,
+  if (!webhookUrls) {
+    logger.info(
+      `[webhook][${jobId}] No webhook urls configured. Skipping webhook job.`,
     );
+    return;
   }
+  const webhookToken = serverConfig.webhook.token;
+  const webhookTimeoutSec = serverConfig.webhook.timeoutSec;
 
-  const { bookmarkId } = request.data;
+  const { bookmarkId } = job.data;
   const bookmark = await fetchBookmark(bookmarkId);
   if (!bookmark) {
     throw new Error(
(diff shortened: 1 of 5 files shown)


apps/workers/webhookWorker.ts
```diff
@@ -0,0 +1,109 @@
...
+          const jobId = job.id;
+          logger.error(
+            `[webhook][${jobId}] webhook job failed: ${job.error}\n${job.error.stack}`,
+          )
+        },
+      },
+      {
+        concurrency: 1,
+        pollIntervalMs: 1000,
+        timeoutSecs: serverConfig.inference.jobTimeoutSec,
```
MohamedBassem: Reusing inference timeout here is probably not intentional, let's maybe using the `webhook.timeout / 1000` server settings?

apps/workers/webhookWorker.ts
```diff
@@ -0,0 +1,109 @@
...
+    );
+  }
+
+  logger.info(
+    `[webhook][${jobId}] Starting a webhook job for bookmark with id "${bookmark.id}"`,
+  );
+
+  for (const url of webhookUrls) {
+    try {
+      const response = await Promise.race([
```
MohamedBassem: I think instead of Promise.race, we can use `AbortSignal.timeout` here so that the fetch itself get cancelled as well.

apps/workers/webhookWorker.ts
```diff
@@ -0,0 +1,109 @@
...
+        )
+      ]);
+
+      if (!(response instanceof Response)) {
+        throw new Error(`Webhook call to ${url} failed: response is not a Response object`);
+      }
+
+      if (!response.ok) {
+        logger.error(`Webhook call to ${url} failed with status: ${response.status}`);
+      }
```
MohamedBassem, line 133: What should we do if the webhook response is a non success status code? In that case, do we want to retry or not? I think we probably should?

docs/docs/03-configuration.md
```diff
@@ -95,3 +95,33 @@ Hoarder uses [tesseract.js](https://github.com/naptha/tesseract.js) to extract t
...
+| WEBHOOK_TIMEOUT | No       | 5000    | The timeout for the webhook request in milliseconds.                                           |
+
+:::info
+
+- If a url is add to hoarder , after it is crawled, the webhook will be triggered.
+- The WEBHOOK_TOKEN is used for authentication. It will appear in the Authorization header as Bearer token.
+  ```
+  Authorization: Bearer <WEBHOOK_TOKEN>
+  ```
+- The webhook will be triggered with the job id (used for idempotent), bookmark id, the user id, the url and the operation in JSON format in the body.
```
MohamedBassem:
```suggestion
- The webhook will be triggered with the job id (used for idempotence), bookmark id, the user id, the url and the operation in JSON format in the body.
```

packages/shared/config.ts
```diff
@@ -56,6 +56,12 @@ const allEnv = z.object({
   DATA_DIR: z.string().default(""),
   MAX_ASSET_SIZE_MB: z.coerce.number().default(4),
   INFERENCE_LANG: z.string().default("english"),
+  WEBHOOK_URLS: z.preprocess(
+    (val) => typeof val === "string" ? val.split(",") : val,
+    z.array(z.string().url())
```
MohamedBassem: Hmmm, not sure if the use of `preprocess` here is right. It seems more correct to do do:

```
z.string().transform(val => val.split(",")).pipe(z.array(z.string().url()))
```

packages/shared/queues.ts
```diff
@@ -158,3 +158,30 @@ export const AssetPreprocessingQueue =
       keepFailedJobs: false,
     },
   );
+
+
+//Webhook worker
+export const zWebhookRequestSchema = z.object({
+  bookmarkId: z.string(),
+  url: z.string(),
```
MohamedBassem: I don't think the `url` should be included here. We have bookmarks of other types, and you're already fetching the bookmark in the worker anyways.

packages/shared/queues.ts
```diff
@@ -158,3 +158,30 @@ export const AssetPreprocessingQueue =
...
+  operation: z.enum(["create"]),
+});
+export type ZWebhookRequest = z.infer<typeof zWebhookRequestSchema>;
+export const WebhookQueue = new SqliteQueue<ZWebhookRequest>(
+  "webhook_queue",
+  queueDB,
+  {
+    defaultJobArgs: {
+      numRetries: 1,
```
MohamedBassem: I feel like we probably should retry on failure? Maybe 3 retries?

apps/workers/crawlerWorker.ts
```diff
@@ -746,6 +747,9 @@ async function runCrawler(job: DequeuedJob<ZCrawlLinkRequest>) {
     // Trigger a potential download of a video from the URL
     await triggerVideoWorker(bookmarkId, url);
 
+    // Trigger a webhook
+    await triggerWebhookWorker(bookmarkId, url, "create");
+
```
MohamedBassem: This handles the case for URLs though. I think we should also trigger the webhook for non URL bookmarks which don't go through the crawling workflows. You'll need to trigger those from `packages/trpc/routers/bookmarks.ts`  (at the end of the createBookmark handler).

apps/workers/webhookWorker.ts
```diff
@@ -0,0 +1,109 @@
...
+    try {
+      const response = await Promise.race([
+        fetch(url, {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+            "Authorization": `Bearer ${webhookToken}`,
+          },
+          body: JSON.stringify({
+            jobId, bookmarkId, userId: bookmark.userId, url: bookmark.link.url, operation: job.data.operation
```
MohamedBassem: Let's include bookmark type here as well.

MohamedBassem (changes requested):
Overall, looks good to me! Left a couple of small comments, and then we can merge it!

Great job navigating the codebase without much assistance :)

MohamedBassem:
A more high level question. Do we want to only trigger the webhook after the crawling? Or should we trigger it once on create, and once after it gets crawled? We can also add a configuration setting for the events people want to subscribe to.

mkarliner:
> A more high level question. Do we want to only trigger the webhook after the crawling? Or should we trigger it once on create, and once after it gets crawled? We can also add a configuration setting for the events people want to subscribe to.

To further open the can of worms, what are the events in the life cycle of a bookmark, and should tags or lists also have life cycles? I'm not suggesting these should be done now, but at least asking the question now leaves further expansion options open.

apps/workers/webhookWorker.ts
hanguofeng: I changed it to :
```typescript
timeoutSecs: serverConfig.webhook.timeout / 1000 * (serverConfig.webhook.retryTimes + 1) * serverConfig.webhook.urls.length + 1,

```
This is considering retry times, urls, and timeout and add 1 second for other stuff

apps/workers/webhookWorker.ts
hanguofeng: Goods suggestions , I improved this

apps/workers/webhookWorker.ts
hanguofeng: I added a retry logic and also a config `WEBHOOK_RETRY_TIMES`

docs/docs/03-configuration.md
hanguofeng: Good , changed

hanguofeng: And also some other doc improvements

packages/shared/config.ts
hanguofeng: Yes, I changed it and with default value

packages/shared/queues.ts
hanguofeng: Yes , I removed it and also removed from `triggerWebhookWorker`

packages/shared/queues.ts
hanguofeng: I changed it to 3

apps/workers/crawlerWorker.ts
hanguofeng: Considering that the question of which events to support needs more debate and consideration, I've constrained it to "crawled." This means we only support the event that signals crawl completion for now. This preserves scalability and prevents current issues from becoming unnecessarily complex like @mkarliner saying.

apps/workers/webhookWorker.ts
hanguofeng: Yes, added

apps/workers/crawlerWorker.ts
mkarliner: Been thinking about this the last few days.
May I suggest a event of 'edited' or 'changed'.
Here is the use case: 
I intend to use these events to post a stream of my bookmarks to social media, in particular the Fediverse/Maston, through the ActivityPub protocol. I don't actually want an unfiltered stream to be posted, I want to curate it at some level. 
So, with an 'changed' event, I can manually change a bookmark in some way that will indicated to my bot that this is a bookmark I want to share and it will pick up the change and publish it. 

It's worth noting that there are now Fediverse - Threads - Bluesky bridges, so this solution would get Hoarder bookmarks to a potentially large audience

Alternatively, if the event specified which list the bookmark was added to, that would also work.

MohamedBassem:
Sorry that this taking long to review, I'll be able to have a look on it this weekend. @mkarliner your suggestion about more events makes a lot of sense now that we have the infra to do it. I think we can land this PR with the current events and followups with more PRs to add more events. Shouldn't be too hard.

MohamedBassem:
@hanguofeng Thanks a lot @hanguofeng for the high quality PR! I did some minor modifications before landing.

- If I get time before the next release (specially with the ongoing trademark drama), I might make this functionality configurable per user such that every user can configure their own webhooks from the UI.
- Regarding obsidian, someone actually just submitted an obsidian sync plugin to the obsidian plugin store (https://github.com/hoarder-app/hoarder/discussions/622).
- Vector search and offline reading are both planned. In fact I already have a non-polished PR for embeddings and using a vector database (https://github.com/hoarder-app/hoarder/pull/834).
- @mkarliner I'll followup with another PR to add `created/edited` events!

MohamedBassem:
Ended up pushing the changes to make this configurable by user, and added the `created`/`edited` lifecycle hooks.

erik-nilcoast:
This is great, I'll integrate this into my RSS aggregator! Thanks.

PR #866, opened 2025-01-12
Fix context menu handling to prioritize link URLs and include anchor text as bookmark titles

dakshpareek:
**Problem:** When users select anchor text and click "Add to Hoarder," the extension was capturing the selected text instead of the link URL. As a result, only the selected text was saved in Hoarder.

issue

**Solution:** **Modified `handleContextMenuClick` in `background.ts`:**
  - Prioritized `info.linkUrl` over `info.selectionText` when determining what to hoard.
  - Captured the selected anchor text as the `title` when saving a link bookmark.

solution

### DIFF
--- a/apps/browser-extension/src/background/background.ts
+++ b/apps/browser-extension/src/background/background.ts
@@ -56,7 +56,13 @@ async function handleContextMenuClick(info: chrome.contextMenus.OnClickData) {
     });
   } else if (menuItemId === ADD_LINK_TO_HOARDER_ID) {
     let newBookmark: ZNewBookmarkRequest | null = null;
-    if (info.selectionText) {
+    if (info.linkUrl) {
+      newBookmark = {
+        type: BookmarkTypes.LINK,
+        url: info.linkUrl,
+        title: info.selectionText ?? undefined,
+      };
+    } else if (info.selectionText) {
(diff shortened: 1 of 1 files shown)


apps/browser-extension/src/background/background.ts
```diff
@@ -56,7 +56,13 @@ async function handleContextMenuClick(info: chrome.contextMenus.OnClickData) {
...
   } else if (menuItemId === ADD_LINK_TO_HOARDER_ID) {
     let newBookmark: ZNewBookmarkRequest | null = null;
-    if (info.selectionText) {
+    if (info.linkUrl) {
+      newBookmark = {
+        type: BookmarkTypes.LINK,
+        url: info.linkUrl,
+        title: info.selectionText ?? undefined,
+      };
+    } else if (info.selectionText) {
```
MohamedBassem, line 65: hmmm, the ordering of the `ifs` makes me think that having text selection taking higher priority was some intentional decision. I'll need to remember why it was done that way :)

dakshpareek: Haha. Take your time. :)

dakshpareek: any update on this?

Issue #872, opened 2025-01-12
Bulk tag deletion


fabfabretti:
### Describe the feature you'd like

Hi! First of all, thanks for the awesome project :)

My request is to be able to delete multiple tags (or even all tags) easily, for example by being able to select multiple tags in the /tags page and then hitting "delete" once.

Alternatively, this could also be implemented by being able to easily delete tags from bookmarks when bulk-editing bookmarks: in Bulk Edit mode, it would be nice to be able to easily remove all tags from selected bookmarks (and this would allow easy deletion of tags from the /tags page, as they would be orphaned).

### Describe the benefits this would bring to existing Hoarder users

I believe it could be useful to be able to easilly delete a high number of tags that are not yet orphaned (e.g. overly specific or underly used). This could also be useful to "start from scratch" if one wants to change their tagging system/prompt and start from a clean slate, particularly so with the recent introduction of smart lists.

### Can the goal of this request already be achieved via other means?

The only ways I found to be able to delete tags is either:

- **Delete tags one at a time:** this is very slow, as at the moment we can't multi-select tags (from what I'm aware) and we need to click the x on each tag and then click on the confirmation button.
- **Merge all tags into one, and then delete the single remaining tag**. This is doable and I've done this in the past, but it's still quite time consuming if you have a lot of tags, which is usually the case when using the AI-tagging system.
- You can **select all articles, enter bulk edit mode and finally add+delete each tag, one by one**. This works, but again as per point 2 it takes quite a lot of time when you use AI-tagging system and have a lot of tags.
- From [this comment ](https://github.com/hoarder-app/hoarder/issues/111#issuecomment-2496173499), it should be possible to bulk delete tags by **rerunning inference tag service for all items and using a custom prompt that asks the AI for an empty/limited array.** I was personally unable to do this, as each time I rerun the tagging with the mentioned prompt I would be able to orphan some of the tags, but not all of them (I had 2 orphaned tags before this test; after the first retag it was 32 orphans, with the second one 95, final attempt 165 orphans). Still, from what I understand, this still has a big downside: this requires you to resend all of the content to your AI API again, potentially "wasting" a lot of tokens for something that could be done without AI. 

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

### Additional context

I've tried looking for similar issues with terms like "bulk tag edit","tag delete","tag deletion" but couldn't find this issue. I apologize if I missed a similar issue, and please let me know if you have any question!

[...]

PR #900, opened 2025-01-18
feat(web): #872 bulk tag deletion

lexafaxine:
#872 

# Add bulk tag deletion feature

## Changes
- Added bulk selection mode toggle for tags
- Added bulk actions toolbar with:
  - Select/Unselect all tags option
  - Delete selected tags action
  - Confirmation dialog for bulk deletion

## Demo video
https://drive.google.com/file/d/192R-6t2cZPy2g8Dn_9BIRtrONs-Ir38z/view?usp=sharing


apps/web/components/dashboard/tags/AllTagsView.tsx
```diff
@@ -75,18 +78,26 @@ export default function AllTagsView({
...
+  useEffect(() => {
+    const visibleTags: ZTagBasic[] = initialData.map((tag) => ({
+      id: tag.id,
+      name: tag.name,
+    }));
+    setVisibleTags(visibleTags);
+    return () => {
+      setVisibleTags([]);
+    };
+  }, [initialData]);
```
MohamedBassem: hmmm, I don't think you should depend on initial data because it means you'll miss the updates after the first server side rendering. You should depend on `data` that's coming out of react query later in the function.

apps/web/components/dashboard/tags/BulkTagAction.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const [isDeleteDialogOpen, setIsDeleteDialogOpen] = useState(false);
+  const pathname = usePathname();
+  const [currentPathname, setCurrentPathname] = useState("");
+
+  useEffect(() => {
+    if (pathname !== currentPathname) {
+      setCurrentPathname(pathname);
+      setIsBulkEditEnabled(false);
+    }
+  }, [pathname, currentPathname]);
```
MohamedBassem: Should we maybe here do an unmount hook instead?

apps/web/components/dashboard/tags/BulkTagAction.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+      {!isBulkEditEnabled ? (
+        <Toggle
+          variant="outline"
+          aria-label="Toggle bulk edit"
+          pressed={isBulkEditEnabled}
+          onPressedChange={setIsBulkEditEnabled}
+        >
+          <Pencil className="mr-2 size-4" />
+          Bulk Edit
```
MohamedBassem: We have a translation for this in `actions.bulk_edit` that we can use.

apps/web/components/dashboard/tags/BulkTagAction.tsx
```diff
@@ -0,0 +1,155 @@
...
+          <InfoTooltip size={15} className="my-auto ml-2" variant="explain">
+            <p>Bulk Edit Tags</p>
+          </InfoTooltip>
```
MohamedBassem: This is probably redundant, we can remove it

apps/web/lib/bulkTagActions.ts
```diff
@@ -0,0 +1,61 @@
+import { create } from "zustand";
+
+import type { ZTagBasic } from "@hoarder/shared/types/tags";
+
+interface TagState {
+  selectedTags: ZTagBasic[];
+  visibleTags: ZTagBasic[];
```
MohamedBassem: Does the store actually care about the name of the tag or is the ids enough?

MohamedBassem (changes requested):
Hey, I'm really sorry for the delay in the review. Overall, the change looks good to me. I've requested some changes that shouldn't take long to address and then we'll be able to merge this PR!

apps/web/components/dashboard/tags/AllTagsView.tsx
lexafaxine: Fixed in https://github.com/hoarder-app/hoarder/pull/900/commits/ffef39aa3584b2376b511310eaf61ccb818311e0

apps/web/components/dashboard/tags/BulkTagAction.tsx
lexafaxine: You are right. Fixed in https://github.com/hoarder-app/hoarder/pull/900/commits/ffef39aa3584b2376b511310eaf61ccb818311e0

apps/web/components/dashboard/tags/BulkTagAction.tsx
lexafaxine: Thank you! Change it to use translation in  https://github.com/hoarder-app/hoarder/pull/900/commits/ffef39aa3584b2376b511310eaf61ccb818311e0

apps/web/components/dashboard/tags/BulkTagAction.tsx
lexafaxine: Removed in  https://github.com/hoarder-app/hoarder/pull/900/commits/ffef39aa3584b2376b511310eaf61ccb818311e0

apps/web/lib/bulkTagActions.ts
lexafaxine: No we don't need the name. Fixed it in https://github.com/hoarder-app/hoarder/pull/900/commits/ffef39aa3584b2376b511310eaf61ccb818311e0

lexafaxine:
> Hey, I'm really sorry for the delay in the review. Overall, the change looks good to me. I've requested some changes that shouldn't take long to address and then we'll be able to merge this PR!

Thank you for your review! That's totally fine, if there's any issue please feel free to request other changes

MohamedBassem:
Thanks a lot @lexafaxine for addressing the feedback and for the PR!

PR #927, opened 2025-01-22
fix: margin for "Lists" tab content

danielvigaru:
It's a small thing, but it was bugging me a lot 😬

| Before: | After: |
|:---:|:---:|
|  |  |
|  |  |

### DIFF
--- a/apps/mobile/lib/utils.ts
+++ b/apps/mobile/lib/utils.ts
@@ -37,7 +37,7 @@ export function condProps(
     condition: boolean;
     props: Record<string, unknown>;
   }[]
-) {
+): Record<string, unknown> { 
   return condProps.reduce((acc, { condition, props }) => {
     return condition ? { ...acc, ...props } : acc;
   }, {});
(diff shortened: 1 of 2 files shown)


apps/mobile/lib/utils.ts
```diff
@@ -5,3 +5,40 @@ import { twMerge } from "tailwind-merge";
...
+ * ```
+ * <View style={ color: "red", fontWeight: "bold" } />
+ * ```
+ */
+export function condProps(
+  ...condProps: {
+    condition: boolean;
+    props: Record<string, unknown>;
+  }[]
+) {
```
imcodingideas: Nit: we could explicitly add the return type here 

```suggestion
): Record<string, unknown> { 
```

imcodingideas: Thanks @danielvigaru

MohamedBassem:
I'm really sorry for the delay in reviewing this PR, I promise I'll get to it at most over the weekend.

MohamedBassem:
Thanks for the fix!

PR #947, opened 2025-01-30
feat: Pluralize and format tags

sweepies:
This PR mainlines a couple of tagging rules I added to my personal instance:

- `Tags should be converted to Title Case.`
- `If applicable, tags should be plural. For example, 'recipes' instead of 'recipe'.`

These have worked great for me.


kamtschatka:
there are people who have the exact opposite and want lower case tags only. You can add those rules locally with the custom tagging rules, so this is where they should probably stay.

sweepies:
It's opinionated indeed but I think we need a default of *some* sort. The default behaviour should not open up potential for four different tags of the same subject.

i.e. `recipe`/`recipes`/`Recipe`/`Recipes`

Maybe putting these in the system prompt isn't the solution. Perhaps a better place for them would be a couple pre-filled but editable fields?

MohamedBassem:
I think changing those rules now can be a bit disruptive. How about we start a new docs page where people can share some of their custom prompts?

MohamedBassem:
We can even potentially bake this custom rules library in hoarder itself. A modal that opens with one row per rule, and a button to "Add" it to your rules.

sweepies:
> We can even potentially bake this custom rules library in hoarder itself. A modal that opens with one row per rule, and a button to "Add" it to your rules.

This sounds good to me, where should we store the rules? In a new hoarder-app/community-rules repo?

MohamedBassem:
@sweepies We can start with a JSON file in this repo, and if there's enough traction we can move it to a separate repo.

MohamedBassem:
Given that the suggested approach is going to be very different, I'll close this PR for now.

PR #955, opened 2025-02-01
Updated Firefox extension URL

jessyjones:
The old link was invalid, replaced with what I'm assuming is the right extension (TBC)


MohamedBassem:
The firefox extension is now down due to some trademark dispute, more details in https://github.com/hoarder-app/hoarder/issues/899. This is not the correct extension.

jessyjones:
Sorry about that !

PR #957, opened 2025-02-01
Update docker-compose.yml

mexicanamerican:


MohamedBassem:
The env file is part of the installation and it's what's recommended in the docs. If you don't need it, and you want to inline all env variables sure, but removing it in this PR is wrong.

PR #958, opened 2025-02-01
Update install script to use latest from Main

vhsdream:
Checks Github API for latest commit hash, uses hash as version, pulls down main.zip, for both install and update functions. This way the install script can take advantage of the latest fixes/improvements.

### DIFF
--- a/hoarder-linux.sh
+++ b/hoarder-linux.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
-# Copyright 2024
+# v2.0
+# Copyright 2025
 # Author: vhsdream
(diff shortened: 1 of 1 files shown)


MohamedBassem (changes requested):
@vhsdream It's kinda not recommended to use latest unless people explicitly opt into it because it's usually less stable. I'm ok with making this optional (e.g. by passing `--latest`) but I don't think making it the default is the right call.

vhsdream:
> @vhsdream It's kinda not recommended to use latest unless people explicitly opt into it because it's usually less stable. I'm ok with making this optional (e.g. by passing `--latest`) but I don't think making it the default is the right call.

I think that's a good compromise! I'm going to close this one and continue working on the script 👍

PR #980, opened 2025-02-04
Hoarder Debian/Ubuntu install script - Add error handling; other small fixes

vhsdream:
Script will now exit on errors (and provide some info) and when there is an unset variable. Added check for Corepack v0.31.0 in update function.

### DIFF
--- a/hoarder-linux.sh
+++ b/hoarder-linux.sh
@@ -1,9 +1,22 @@
 #!/usr/bin/env bash
 
-# Copyright 2024
+set -Eeuo pipefail
+
+# v2.0
+# Copyright 2024-2025
 # Author: vhsdream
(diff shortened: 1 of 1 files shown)


hoarder-linux.sh
```diff
@@ -254,6 +267,10 @@ function update {
     wget -q "https://github.com/hoarder-app/hoarder/archive/refs/tags/v${RELEASE}.zip"
     unzip -q v${RELEASE}.zip
     mv hoarder-${RELEASE} ${INSTALL_DIR}
+    # https://github.com/hoarder-app/hoarder/issues/967
+    if [[ $(corepack -v) < "0.31.0" ]]; then
```
MohamedBassem, line 271: This will break when corepack reaches `0.100` but I think it's the problem of future me :D

MohamedBassem (approved):

PR #983, opened 2025-02-05
feat(i18n): add Arabic language support

AhmadMuj:


MohamedBassem:
I didn't expect that making the whole website RTL would be that easy. This is amazing, thank you!

Issue #711, opened 2024-12-02
PDF is auto-downloaded on page load


LeskoIam:
### Describe the Bug

If PDF is uploaded to hoarder it is auto-downloaded on every page refresh.

### Steps to Reproduce

1. Uploaded PDF document to hoarder (drag&drop)
2. Refresh page.
3. Previously uploaded document is downloaded (without clicking anything. except refresh button in browser)

### Expected Behaviour

No file should be downloaded.

### Device Details

133.0 (64-bit) Windows 10 Pro (22H2)

### Exact Hoarder Version

v0.19.0

MohamedBassem:
Yeah. That's an unfortunate side effect from how we render the bookmark cards. We should definitely fix that.

MohamedBassem:
Fixed by #995

PR #995, opened 2025-02-08
feat: Add PDF screenshot generation and display

AhmadMuj:
- Add PDF to image conversion using pdf2pic library
- Install graphicsmagick and ghostscript dependencies 
- Generate and store PDF screenshots during preprocessing
- Add screenshot display support for web and mobile apps
- Add missing PDF screenshot generation via tidyAssets worker
- Update pdf2json to 3.1.5 for better parsing and removed old commits about the incorrect typing

This change improves PDF preview performance by generating and displaying static screenshots instead of loading full PDFs. 
A fallback to PDF iframe display is maintained when screenshots are unavailable.


apps/workers/package.json
```diff
@@ -1,68 +1,69 @@
 {
```
MohamedBassem: Can you revert the unnecessary changes in this file?

apps/workers/assetPreprocessingWorker.ts
```diff
@@ -67,24 +75,75 @@ async function readImageText(buffer: Buffer) {
...
 }
 
+export async function extractAndSavePDFScreenshot(
+  pdfBuffer: Buffer,
+  userId: string,
+  bookmarkId: string,
+  jobId: string,
+): Promise<boolean> {
+  try {
+    console.log("extracting =================================");
```
MohamedBassem: Let's not keep console logs here.

apps/mobile/components/bookmarks/BookmarkCard.tsx
```diff
@@ -300,11 +301,21 @@ function AssetCard({
   }
   const title = bookmark.title ?? bookmark.content.fileName;
 
+  let assetImage = bookmark.content.assetId;
```
MohamedBassem: I think to solve #711, let's stop rendering the pdf at all if it doesn't have a screenshot. We can render a blank image of the word "pdf" in html or something.

apps/web/components/dashboard/bookmarks/AssetCard.tsx
```diff
@@ -32,13 +35,42 @@ function AssetImage({
...
+                  </Link>
+                </p>
+              )}
+            </div>
+            <iframe
+              title={bookmarkedAsset.assetId}
+              className={className}
+              src={getAssetUrl(bookmarkedAsset.assetId)}
+            />
+          </div>
```
MohamedBassem: Same comment as above, let's just render a blank html saying "pdf" instead.

apps/web/components/dashboard/preview/AssetContentSection.tsx
```diff
@@ -1,42 +1,92 @@
...
+  SelectGroup,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
 
 import { BookmarkTypes, ZBookmark } from "@hoarder/shared/types/bookmarks";
 
 export function AssetContentSection({ bookmark }: { bookmark: ZBookmark }) {
+  const [section, setSection] = useState<string>("");
```
MohamedBassem: I feel like that it might not be worth it to have a selector for this? It's just an optimization for rendering the grid in an efficient way? What do you think?

apps/web/components/dashboard/preview/AssetContentSection.tsx
```diff
@@ -1,42 +1,92 @@
...
+  useEffect(() => {
+    const screenshot = bookmark.assets.find(
+      (item) => item.assetType === "screenshot",
+    );
+    if (screenshot) {
+      setSection("screenshot");
+    } else {
+      setSection("pdf");
+    }
+  }, [bookmark]);
```
MohamedBassem: As I mentioned above, this might not be worth it, but if we're going that route, `useMemo` is more correct here than `useEffect`

apps/workers/assetPreprocessingWorker.ts
```diff
@@ -16,6 +22,8 @@ import {
   triggerSearchReindex,
 } from "@hoarder/shared/queues";
 
+import { storeScreenshot } from "./crawlerWorker";
```
MohamedBassem: Not a big fan of cross worker imports. If you want to share that logic, move the parts you're interested in in the `assetDb` util class.

apps/workers/assetPreprocessingWorker.ts
```diff
@@ -67,24 +75,75 @@ async function readImageText(buffer: Buffer) {
...
+      logger.error(`[${jobId}] Failed to store PDF screenshot`);
+      return false;
+    }
+
+    // Insert into database
+    await db.insert(assets).values({
+      id: asset.assetId,
+      bookmarkId,
+      userId,
+      assetType: AssetTypes.LINK_SCREENSHOT,
```
MohamedBassem: `LINK_SCREENSHOT` are meant to be in particular for "link" screenshots. This is more like `ASSET_SCREENSHOT`.

apps/workers/tidyAssetsWorker.ts
```diff
@@ -95,6 +102,55 @@ async function runTidyAssets(job: DequeuedJob<ZTidyAssetsRequest>) {
     );
   }
 
+  if (request.data.generateMissingPDFScreenshots) {
```
MohamedBassem: Tidy assets is meant to be mainly dangling assets and that's why you're copying logic here. How about we allow in the admin panel to re-trigger asset preprocessing instead and have a special mode to "fix" missing pre-processing?

pnpm-lock.yaml
MohamedBassem, line 1: There's a lot of changes here which makes me believe you're using a different version of pnpm than the pinned one in the `package.json`. Use corepack to take you to the proper version we're using and regenerate this file.

apps/workers/assetPreprocessingWorker.ts
ahmadmucom: Ops

apps/web/components/dashboard/preview/AssetContentSection.tsx
ahmadmucom: I kept in mind the extreme cases of having VERY large PDF files when the user wants to edit the tags but don't want to load +100MB for example. What do you think ?

MohamedBassem: hmmm, good point. At this point, we know the asset size before rendering it. How about we by default show the pdf if it's size is say (<10MB), and render the screenshot instead if it's bigger than 10MB with a button on top say "download pdf" or something to load the full pdf in this case?

ahmadmucom: It's a valid option as well. I'll work on it.

apps/mobile/components/bookmarks/BookmarkCard.tsx
kamtschatka: would be better to tell the user that there is no preview generated yet and can be generated in the admin panel as mentioned below (or it failed due to a bug, then it should be fixed instead?)

ahmadmucom: In the mobile app, we are not rendering the PDF anyway. It's always showing as a black blank so the edit will actually solve this in the mobile app

ahmadmucom: @kamtschatka I already implemented that on the web UI, should this be the case in the mobile app as well ?

apps/workers/tidyAssetsWorker.ts
ahmadmucom: @MohamedBassem Basically you want to introduce a new Queue for fixing preprocessing issues in general ?

MohamedBassem: @AhmadMuj no, we can just trigger the asset preprocessing job from the admin panel in a "fix" only mode that fixes the missing preprocessing.

apps/mobile/components/bookmarks/BookmarkCard.tsx
kamtschatka: I think anything that also works in the mobile app is a big win. It is currently way behind in features compared to the web UI. I personally don't use it, but seems like there are plenty of people who are.

apps/workers/tidyAssetsWorker.ts
ahmadmucom: preprocessing job would send the item again for the OpenAIQueue which would cause an additional unjustified cost. That's why I avoided it in the first place.

MohamedBassem: We can make it conditionally send them or not depending on what got fixed.

AhmadMuj:
@MohamedBassem Should be good to go. Let me know if there's anything.

AhmadMuj:
@MohamedBassem Sounds good to me

PR #996, opened 2025-02-08
feat: add URL protocol validation

AhmadMuj:
Added URL protocol validation for mobile app and browser extension.


MohamedBassem:
We get a ton of issues because of this, thanks for putting an end to it!

Issue #1018, opened 2025-02-11
Allow changing the entire prompt


poculeka:
### Describe the feature you'd like

I would like to have a possibility to change default prompts (text, image, summarization) without recompiling the project.

### Describe the benefits this would bring to existing Hoarder users

With this feature, users will be able to fully customize the prompts used to communicate with LLMs.

There are some open issues, where current prompts are tweaked to work better with users' local language models, but each model works differently - prompts which work perfectly with llama3.2 might not work with deepseek-r1 or qwen.

In my case, the default prompt didn't always return satisfying results, even when I created some tagging rules using Hoarder settings.

### Can the goal of this request already be achieved via other means?

Yes, Hoarder can be rebuilt with custom prompt. It is very frustrating, though - in order to keep Hoarder version up to date, each new release must be rebuilt with prompt changes.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### Additional context

My suggestion would be to allow prompts customization through .env variables. This way the default prompts can be kept in the codebase and they could be updated on demand without rebuilding the application / container. Such approach would also be consistent with other inference configuration options - for example OpenAI / Ollama URLs, which are also global for the whole Hoarder instance.

I already have some code which implements these changes, I can open a PR if such solution is acceptable.

kamtschatka:
i am also running into issues, because the prompt contains things you might want to override.
- Aim for 3-5 tags.
3-5 tags is usually quite excessive, I would prefer 3 at max

MohamedBassem:
Please have a look at my comment here: https://github.com/hoarder-app/hoarder/pull/1029#issuecomment-2661602853
This is intentionally restricted. I'm wiling to add more flexibility to customize pre-packaged rules, but I'm not planning to give up the control on the prompt structure itself to allow me to evolve it in the future.

poculeka:
Let's drop one issue from the list 😉 
Thanks for all the work you're putting into this project!

PR #1029, opened 2025-02-15
Allow custom prompts through environment variables

poculeka:
### Description
This PR introduces prompt customization through environment variables mentioned in #1018.
It solves issues with local language models being incompatible with prompts hardcoded inside the application.
This feature is backward compatible - if none of new environment variables is defined, Hoarder should behave exactly as it behaved before. 

### Technical details
3 new optional environment variables are defined:
* `TAGGING_PROMPT` - tag generation for text content,
* `IMAGE_PROMPT` - tag generation for images,
* `SUMMARIZATION_PROMPT` - summarizing text.

Default prompts are moved into `config.ts` and they are used if corresponding environment variables were not defined.

#### Filling prompts with data
In current solution, content is being inflated into the prompt text using string interpolation. Since there is no string interpolation in environment variables, I introduced markers which are replaced by actual values when building a prompt. The marker values are: 
* `tags`, 
* `aiTags`, 
* `userTags`, 
* `lang`, 
* `customPrompts`  (tagging rules),
* `content`. 

Values `tags`, `aiTags`, `userTags` existed before and were used in "Tagging Rules" feature.
However, I decided to change current markers from using dollar sign (`$variableName`) to doubled curly braces (`{{ variableName }}`), because dollar sign would have to be escaped when defined inside `.env` file (I'm not insisting on this, we can discuss other templating solutions). To keep backward compatibility, markers `tags`, `aiTags` and `userTags` can also be defined with dollar-sign notation.

### Custom prompts through env variables vs custom prompts using tagging rules
Hoarder already allows some customization in prompts using "tagging rules" in AI settings. However, tagging rules don't change the entire prompt. But I don't think these features would be in conflict with each other - custom prompt defined through environment variable sets the prompt template for the entire Hoarder instance, while tagging rules are additional per-user settings which could be inflated into the global prompt template with `{{ customPrompts }}` marker. In this way they supplement each other.

### DIFF
--- a/apps/web/components/settings/AISettings.tsx
+++ b/apps/web/components/settings/AISettings.tsx
@@ -28,9 +28,8 @@ import { useForm } from "react-hook-form";
 import { z } from "zod";
 
 import {
-  buildImagePrompt,
-  buildSummaryPrompt,
-  buildTextPrompt,
+  buildContentPromptFromTemplate,
+  buildPromptFromTemplate,
 } from "@hoarder/shared/prompts";
 import {
   zNewPromptSchema,
@@ -297,9 +296,35 @@ export function TaggingRules() {
   );
 }
 
+function fetchCustomPrompts() {
+  const { data: prompts } = api.prompts.list.useQuery();
+
+  const textPrompts = (prompts ?? [])
+    .filter((p) => p.appliesTo == "text" || p.appliesTo == "all_tagging")
+    .map((p) => p.text);
+  const imagePrompts = (prompts ?? [])
+    .filter((p) => p.appliesTo == "images" || p.appliesTo == "all_tagging")
+    .map((p) => p.text);
+  const summaryPrompts = (prompts ?? [])
+    .filter((p) => p.appliesTo == "summary" || p.appliesTo == "all_tagging")
+    .map((p) => p.text);
+  return {
+    text: textPrompts,
+    images: imagePrompts,
+    summary: summaryPrompts,
+  };
+}
+
 export function PromptDemo() {
   const { t } = useTranslation();
-  const { data: prompts } = api.prompts.list.useQuery();
+
+  const tags = {
+    all: ["<ALL_TAGS>"],
+    ai: ["<AI_TAGS>"],
+    human: ["<USER_TAGS>"],
+  };
+  const customPrompts = fetchCustomPrompts();
+
   const clientConfig = useClientConfig();
   return (
     <div className="flex flex-col gap-2">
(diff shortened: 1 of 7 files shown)


MohamedBassem:
@poculeka Hey, I really appreciate the effort you put into this PR, but unfortunately, I'm intentionally not giving the people the ability to customize the entire prompt because this give me more flexibility to change it in the future without breaking things. For example, if I want to add automatic summariziation to the tagging prompt, I wouldn't be able to do it if people have overriden my original prompt. I want to give people control only over the rules not the prompt structure itself. If I allowed you to disable some of the pre-packaged rules would that be good enough for you?
I'm again really sorry you put so much effort into that, but I'd really suggest you join us in the `#development` channel on discord so that we can be aligned on features before you put effort into implementing them.

poculeka:
Hey, @MohamedBassem, no worries! I decided to open a PR, because I'm already using this code in my instance of Hoarder, so I don't feel like it's a wasted effort. Plus, this was the first time in when I did something webdev-related, so I learned something new :)

Going back to the topic - I think even if disabling some of the pre-packaged rules helped me, there would still exist other users needing deeper customization - I don't know if it's worth the effort.

Let me understand the argument about flexibility better - are you worried that if, in the future, expected response structure changes (together with the prompt), then overriden prompts will stop working? If that's the case, I'd say it is responsibility of owner of the Hoarder instance to keep prompts working with the rest of the application. And this is what I had in mind while developing this feature - the default prompt remains unchanged and it is only possible to change it through an environment variable. A regular user is not able to change anything other than tagging rules.

Additionally, both OpenAI ([click](https://platform.openai.com/docs/guides/structured-outputs)) and Ollama ([click](https://ollama.com/blog/structured-outputs)) requests can be configured in such way that they adhere to supplied JSON schema (this is different than `response_format: { type: "json_object" }`). Structured outputs might be worth looking into, even if this PR will not be merged.


Anyway, this is a great project - keep up the good work!

MohamedBassem:
@bentrop the discord channel is here: https://discord.gg/NrgeYywsFh

And there's no english only mandate, you can customize the language of the tags using the `INFERENCE_LANG` env variable. You can also set `INFERENCE_LANG` to `language of the content` and it'll work though it'll be a hack.

As for the amount of tags and the length, I'm open to add customization points/hooks there. But as I mentioned, I'm not willing to give up full control over the prompt.

bentrop:
Thank you very much for that link, no idea why even Google failed to find that for me.

I understand your desire to not give up _full_ control, but the language and tag amount/summary length like a rather limiting and entirely needless restrictions.

For example, like many multilingual users, I'd prefer to have my native language summarized and tagged in my native language while I would like all other languages to be tagged and summarized in English. That's simply not possible with your prompt _or_ your suggested hacky workaround.

Telling the LLM to ignore the hardcoded parts of the prompt, like so ...

"Ignore that the previous request that "The summary language must be in english", instead the summary language must be in German if the original content is in German, otherwise it must be in English."

... failed for me, at least for summaries. It looks like it worked for tags, though, for some reason.

PR #1036, opened 2025-02-16
Add Unsorted tab to dashboard

garygcchiu:
Add a new 'Unsorted' tab to the dashboard to include all items not associated with a list.

* **Dashboard Layout**:
  - Import `Inbox` icon from `lucide-react`.
  - Add 'Unsorted' tab to the `items` function with the `Inbox` icon and path `/dashboard/unsorted`.

* **Unsorted Page**:
  - Create `apps/web/app/dashboard/unsorted/page.tsx`.
  - Import and use the `Bookmarks` component to display items not associated with any list.
  - Set the `query` prop to filter items not associated with any list.
  - Add a header with the title 'Unsorted'.

### DIFF
--- a/apps/web/app/dashboard/layout.tsx
+++ b/apps/web/app/dashboard/layout.tsx
@@ -7,7 +7,7 @@ import { Separator } from "@/components/ui/separator";
 import { api } from "@/server/api/client";
 import { getServerAuthSession } from "@/server/auth";
 import { TFunction } from "i18next";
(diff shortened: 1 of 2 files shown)


MohamedBassem:
Hey, thanks for the PR. But I don't think this does what you think it does, and it probably doesn't work.

You can achieve what you want by creating a smart list with the filtering being `-is:inlist` instead.

garygcchiu:
Sorry, this was meant to be opened as a draft and was WIP. Thanks for the very quick review and guidance.

PR #1078, opened 2025-02-28
docs: Add instructions for node environment setup

mos3abof:
Hoarder uses node version 22, and makes use of `corepack` features. Instructions for setting those up correctly were missing.


MohamedBassem:
Thank you!

PR #1088, opened 2025-03-02
Adds summarize action to the public API (proxy to TRPC mutation)

erik-nilcoast:
I added this in two commits. One is just for me. I had a few issues getting this running locally on a Linux box. It could just be a me problem :|

The second commit has the addition of a subroute for the summary mutation. It should return the bookmark object with the summarized contents. Please feel free to add or remove any commits and I'll do my best to address any feedback.

Thanks for a fantastic project.

Addresses: https://github.com/hoarder-app/hoarder/issues/523

### DIFF
--- a/apps/web/app/api/v1/bookmarks/[bookmarkId]/summarize/route.ts
+++ b/apps/web/app/api/v1/bookmarks/[bookmarkId]/summarize/route.ts
@@ -1,0 +1,19 @@
+import { NextRequest } from "next/server";
+import { buildHandler } from "@/app/api/v1/utils/handler";
+
+export const dynamic = "force-dynamic";
+
+export const POST = (
+  req: NextRequest,
+  params: { params: { bookmarkId: string } },
+) =>
+  buildHandler({
+    req,
+    handler: async ({ api }) => {
+      const bookmark = await api.bookmarks.summarizeBookmark({
+        bookmarkId: params.params.bookmarkId,
+      });
+
+      return { status: 200, resp: bookmark };
+    },
+  });
--- a/packages/open-api/hoarder-openapi-spec.json
+++ b/packages/open-api/hoarder-openapi-spec.json
@@ -913,6 +913,85 @@
         }
       }
     },
+    "/bookmarks/{bookmarkId}/summarize": {
+      "post": {
+        "description": "Attaches a summary to the bookmark and returns the updated record.",
+        "summary": "Summarize a bookmark",
+        "tags": [
+          "Bookmarks"
+        ],
+        "security": [
+          {
+            "bearerAuth": []
+          }
+        ],
+        "parameters": [
+          {
+            "$ref": "#/components/parameters/BookmarkId"
+          }
+        ],
+        "responses": {
+          "200": {
+            "description": "The updated bookmark with summary",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "type": "object",
+                  "properties": {
+                    "id": {
+                      "type": "string"
+                    },
+                    "createdAt": {
+                      "type": "string"
+                    },
+                    "modifiedAt": {
+                      "type": "string",
+                      "nullable": true
+                    },
+                    "title": {
+                      "type": "string",
+                      "nullable": true,
+                      "maxLength": 250
+                    },
+                    "archived": {
+                      "type": "boolean"
+                    },
+                    "favourited": {
+                      "type": "boolean"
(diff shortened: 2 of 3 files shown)


apps/web/app/api/v1/bookmarks/summary/route.ts
```diff
@@ -0,0 +1,19 @@
...
+import { z } from "zod";
+
+import { buildHandler } from "../../utils/handler";
+
+export const dynamic = "force-dynamic";
+
+export const POST = (req: NextRequest) =>
+  buildHandler({
+    req,
+    searchParamsSchema: z.object({ bookmarkId: z.string() }),
```
erik-nilcoast: I put this in a query param for now testing, but let me know if it'd be more idiomatic for this code base to use a routeParam (e.g. /api/v1/bookmarks/summary/:bookmark-id), a queryParam, or a structured POST body like: `{ bookmarkId: foo-id }`. Let me know what you prefer.

package.json
```diff
@@ -32,10 +32,5 @@
     "turbo": "^2.1.2"
   },
   "prettier": "@hoarder/prettier-config",
-  "packageManager": "pnpm@9.0.0-alpha.8+sha256.a433a59569b00389a951352956faf25d1fdf43b568213fbde591c36274d4bc30",
```
erik-nilcoast: I'll remove all of these once the question above is resolved. I just need this for now to keep working :)

apps/web/app/api/v1/bookmarks/summary/route.ts
MohamedBassem: I think it'll be more idiomatic if we do it as

```
POST /bookmarks/[bookmarkId]/summarize
```

to do that, you can move this file to
```
apps/web/app/api/v1/bookmarks/[bookmarkId]/summarize/route.ts
```

MohamedBassem (changes requested):
Hey, thanks for opening the PR. Left a small comment about the endpoint!

apps/web/app/api/v1/bookmarks/summary/route.ts
erik-nilcoast: Works for me!

erik-nilcoast: Do you want a GET then?

MohamedBassem: No POST is still the right verb here.

erik-nilcoast:
Switched to a GET subroute, let me know if this needs anything else and I'll remove my build hacks.

apps/web/app/api/v1/bookmarks/summary/route.ts
erik-nilcoast: ack

erik-nilcoast:
Looks green. I'll test locally. Is there somewhere I need to add for docs or are those all autogenerated?

MohamedBassem:
@erik-nilcoast Oh good catch, we need to update the docs here: https://github.com/hoarder-app/hoarder/blob/main/packages/open-api/lib/bookmarks.ts

and then call `pnpm run generate` inside `packages/open-api` to generate the openAPI files.

Also, can we call the endpoint `summarize` instead of `summary`? I feel like here you're not posting a summary but instead triggering a summarize action.

erik-nilcoast:
Will do. I’ll see if I can actually get the build running locally. It didn’t work well with my setup. Probably a me problem.

erik-nilcoast:
Build works great on a mac. ¯\_(ツ)_/¯ 

@MohamedBassem docs are updated and I moved the route to `summarize`

erik-nilcoast:
I think this is likely good to go. Let me know if there's anything else I can do to get this on the release schedule (I'd love to get off my fork). 

I may need to add some additional API endpoints in the near future. I'm looking at deploying Hoarder for a client, but would love to give back where I can. 

RSS of bookmark collections and email newsletter hoarding are things I'm looking to add.

MohamedBassem:
Thanks a lot @erik-nilcoast (hope your client enjoys hoarder!). If you're planning to be active in the development, you might want to join us in the #development channel on discord :) Also I'd highly recommend before working on a new feature to make sure we have an issue for it with the `status/approved` tag, to make sure we're aligned on how this feature will look like!

erik-nilcoast:
@MohamedBassem Will do, I'll join y'all.

PR #1118, opened 2025-03-12
feat(workers): Adds publisher and author og:meta tags to Bookmark

erik-nilcoast:


apps/workers/crawlerWorker.ts
```diff
@@ -63,6 +65,8 @@ import { BookmarkTypes } from "@hoarder/shared/types/bookmarks";
 const metascraperParser = metascraper([
   metascraperAmazon(),
   metascraperReadability(),
+  metascraperAuthor(),
+  metascraperPublisher(),
```
MohamedBassem, line 74: While you're at it, do you want to maybe also store `metascraper-date`? People has been asking for us to extract and store the article's original publication date.

packages/db/drizzle/0042_right_madame_web.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE `bookmarks` ADD `author` text;--> statement-breakpoint
+ALTER TABLE `bookmarks` ADD `publisher` text;
```
MohamedBassem: Something seems wrong here. In the worker, you're adding those fields in `bookmarkLinks` but the migration  is adding them to `bookmarks`?

packages/db/schema.ts
```diff
@@ -111,6 +111,8 @@ export const bookmarks = sqliteTable(
     createdAt: createdAtField(),
     modifiedAt: modifiedAtField(),
     title: text("title"),
+    author: text("author"),
+    publisher: text("publisher"),
```
MohamedBassem: I don't think we need to store those in the root bookmark. Doing it in `bookmarkLinks` should be enough.

packages/db/schema.ts
```diff
@@ -148,6 +150,8 @@ export const bookmarkLinks = sqliteTable(
     // Crawled info
     title: text("title"),
     description: text("description"),
+    author: text("author"),
+    publisher: text("publisher"),
```
MohamedBassem, line 152: The migration file seems to have not been refreshed.

packages/trpc/routers/bookmarks.ts
```diff
@@ -1083,6 +1083,8 @@ export const bookmarksAppRouter = router({
 Title: ${bookmark.title ?? ""}
 Description: ${bookmark.description ?? ""}
 Content: ${bookmark.content ?? ""}
+Publisher: ${bookmark.publisher ?? ""}
+Author: ${bookmark.author ?? ""}
```
MohamedBassem, line 1087: Let's also start indexing them in search by adding the fields to `packages/shared/search.ts`

MohamedBassem (changes requested):
The build will probably fail because you have to regenerate the OpenAPI spec after adding new fields.

apps/workers/crawlerWorker.ts
erik-nilcoast: Yup, I'll add it.

packages/db/drizzle/0042_right_madame_web.sql
erik-nilcoast: There's a 100% chance that I did something wrong. You're right. I'll fix that up. I've never worked on a backend TS app before.

packages/db/schema.ts
erik-nilcoast: ack

packages/trpc/routers/bookmarks.ts
erik-nilcoast: will do

erik-nilcoast: done

packages/db/schema.ts
erik-nilcoast: I think I got it this time.

packages/db/drizzle/0042_flawless_whistler.sql
```diff
@@ -0,0 +1,5 @@
+ALTER TABLE `bookmarkLinks` ADD `author` text;--> statement-breakpoint
+ALTER TABLE `bookmarkLinks` ADD `publisher` text;--> statement-breakpoint
+ALTER TABLE `bookmarkLinks` ADD `datePublished` text;--> statement-breakpoint
+ALTER TABLE `bookmarkLinks` ADD `dateModified` text;--> statement-breakpoint
+ALTER TABLE `bookmarks` ADD `modifiedAt` integer;
```
erik-nilcoast: This snuck in but isn't related to my changes. Maybe some local state or a bad migration?

apps/workers/crawlerWorker.ts
erik-nilcoast: added

packages/db/drizzle/0042_flawless_whistler.sql
MohamedBassem: This is because you edited the snapshot file of migration 41 instead of creating a new one. There's no dangling migrations on the main branch.

What'd I'd recommend to fix this is to revert everything in the `db/drizzle` back to how it was in the main branch, and then re-reun drizzle generate.

packages/shared/search.ts
```diff
@@ -18,6 +18,10 @@ export const zBookmarkIdxSchema = z.object({
   note: z.string().nullish(),
   summary: z.string().nullish(),
   tags: z.array(z.string()).default([]),
+  publisher: z.string().nullish(),
+  author: z.string().nullish(),
+  datePublished: z.string().nullish(),
+  dateModified: z.string().nullish(),
```
MohamedBassem, line 24: You added them to the schema, but you need to fill them up in the `searchWorker`, sorry forgot to mention this eariler.

erik-nilcoast:
> Thank you, the changes look good to me! We just need to fix the migration file problems before merging. If you can give me permissions to edit the PR, I can help as well.

Will do, sorry I've never used this migration framework before and it's real stateful :)

packages/db/drizzle/0042_flawless_whistler.sql
erik-nilcoast: Think I got it.

packages/shared/search.ts
erik-nilcoast: Got it

MohamedBassem:
LGTM! Thanks a lot @erik-nilcoast! Can you please rebase so that I can merge?

MohamedBassem:
Btw, tried running it locally, and I noticed that the crawler working are failing with:

```
2025-03-16T22:07:20.226Z error: [Crawler][838] Crawling job failed: TypeError: rule is not a function
TypeError: rule is not a function
    at findRule (/home/mbassem/repos/hoarder/node_modules/@metascraper/helpers/index.js:434:35)
    at /home/mbassem/repos/hoarder/node_modules/metascraper/src/get-data.js:11:27
    at arrayMap (/home/mbassem/repos/hoarder/node_modules/lodash/lodash.js:653:23)
    at map (/home/mbassem/repos/hoarder/node_modules/lodash/lodash.js:9622:14)
    at getData (/home/mbassem/repos/hoarder/node_modules/metascraper/src/get-data.js:10:5)
    at /home/mbassem/repos/hoarder/node_modules/metascraper/src/index.js:29:12
    at extractMetadata (/home/mbassem/repos/hoarder/apps/workers/crawlerWorker.ts:3:3483)
    at crawlAndParseUrl (/home/mbassem/repos/hoarder/apps/workers/crawlerWorker.ts:3:9512)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async runCrawler (/home/mbassem/repos/hoarder/apps/workers/crawlerWorker.ts:3:12840)
```

Does it work for you?

erik-nilcoast:
> Btw, tried running it locally, and I noticed that the crawler working are failing with:
> 
> ```
> 2025-03-16T22:07:20.226Z error: [Crawler][838] Crawling job failed: TypeError: rule is not a function
> TypeError: rule is not a function
>     at findRule (/home/mbassem/repos/hoarder/node_modules/@metascraper/helpers/index.js:434:35)
>     at /home/mbassem/repos/hoarder/node_modules/metascraper/src/get-data.js:11:27
>     at arrayMap (/home/mbassem/repos/hoarder/node_modules/lodash/lodash.js:653:23)
>     at map (/home/mbassem/repos/hoarder/node_modules/lodash/lodash.js:9622:14)
>     at getData (/home/mbassem/repos/hoarder/node_modules/metascraper/src/get-data.js:10:5)
>     at /home/mbassem/repos/hoarder/node_modules/metascraper/src/index.js:29:12
>     at extractMetadata (/home/mbassem/repos/hoarder/apps/workers/crawlerWorker.ts:3:3483)
>     at crawlAndParseUrl (/home/mbassem/repos/hoarder/apps/workers/crawlerWorker.ts:3:9512)
>     at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
>     at async runCrawler (/home/mbassem/repos/hoarder/apps/workers/crawlerWorker.ts:3:12840)
> ```
> 
> Does it work for you?

I'm getting the same now. Something is up with that metadata helper. I'll see if there's any open issues in the morning. Nothing is jumping out at the moment. These are the lines failing:

```
const findRule = async (rules, args) => {
  let index = 0
  let value

  do {
    const rule = rules[index++]
    const test = rule.test || truthyTest
    if (test(args)) value = await rule(args) // line 434, this is the failure
  } while (!has(value) && index < rules.length)

  return value
}
```

Which is called from here:

https://github.com/nilcoast/hoarder/blob/685adcc4cc0bacd194b9a81bfe1ef80b975af56b/apps/workers/crawlerWorker.ts#L361-L367

So my guess is that library has an issue or it's being initialized improperly?

MohamedBassem:
@erik-nilcoast the issue seems to go away after updating the packages:

```diff
diff --git a/apps/workers/package.json b/apps/workers/package.json
index 08ffd1b..b03ae6d 100644
--- a/apps/workers/package.json
+++ b/apps/workers/package.json
@@ -18,7 +18,7 @@
     "execa": "9.3.1",
     "jsdom": "^24.0.0",
     "liteque": "^0.3.2",
-    "metascraper": "^5.45.24",
+    "metascraper": "^5.46.5",
     "metascraper-amazon": "^5.45.22",
     "metascraper-author": "5.46.5",
     "metascraper-date": "^5.46.5",
@@ -26,7 +26,7 @@
     "metascraper-image": "^5.45.22",
     "metascraper-logo": "^5.45.22",
     "metascraper-logo-favicon": "^5.45.22",
-    "metascraper-publisher": "5.45.22",
+    "metascraper-publisher": "5.46.5",
     "metascraper-readability": "^5.45.22",
     "metascraper-title": "^5.45.22",
     "metascraper-twitter": "^5.45.6",
```

erik-nilcoast:
Great, that's what I was going to try first as well. I'll make that change now.

MohamedBassem:
One last thing :) I noticed that `metascraper-date` doesn't actually work by default. Although their documentation says that it should, they only work if you do:

```diff
diff --git a/apps/workers/crawlerWorker.ts b/apps/workers/crawlerWorker.ts
index e8842e9..6384115 100644
--- a/apps/workers/crawlerWorker.ts
+++ b/apps/workers/crawlerWorker.ts
@@ -64,7 +64,10 @@ import {
 import { BookmarkTypes } from "@hoarder/shared/types/bookmarks";

 const metascraperParser = metascraper([
-  metascraperDate(),
+  metascraperDate({
+    dateModified: true,
+    datePublished: true,
+  }),
   metascraperAmazon(),
   metascraperReadability(),
   metascraperAuthor(),
```

erik-nilcoast:
That makes more sense :)

MohamedBassem:
Also, can you enable maintainer edits for this PR? I want to change the `dateModified` and `datePublished` to be timestamps in the schema (similar to `createdAt`) by parsing the returned dates from metascraper, but at the same time, I've asked you for too many changes, so I can do this change myself :)

erik-nilcoast:
Absolutely. I have no idea how. I'll look around.

PR #1151, opened 2025-03-23
Close browser if connect on demand

Summon528:
Fix https://github.com/hoarder-app/hoarder/discussions/869

Close the browser instance after crawling if browserConnectOnDemand is enabled.
This should help users that are using service like browserless to reduce API usage.


MohamedBassem:
Thanks for the fix, and sorry that this took too long to accept!

PR #1169, opened 2025-03-26
docs: Fixed missing semicolon in update admin password section

jferch:
Thanks for the great app and helpful FAQ entry! I'm by no means a database user but needed to change the password and didn't get it to work initially. Figured it our pretty quickly by myself but might be helpful to other people, too. I took the liberty to slightly adjust the formatting as well so that the command fits in one line but feel free to change as you like. Thanks!


MohamedBassem:
thank you!

PR #1179, opened 2025-03-30
fix(feed): fix URL parsing for Wallabag RSS feeds

### LINKED ISSUE #684: Wallabag RSS fetch bug
### Describe the Bug

I have subscribed to a RSS feed generated by my wallabag instance. However, all the link crawed are wallabag login portal

### Steps to Reproduce

1. Select any rss feed generated from wallabag
2. subscribe in hoarder

### Expected Behaviour

Original link and content fetched

CleanShot 2024-11-22 at 00 10 43@2x

### Device Details

_No response_

### Exact Hoarder Version

v0.19.0

yinan-c:
Thanks for the great project!
This should fix #684, let me know what you think!

It deals with wallabag feed and link parser separately to make sure get original url instead of wallabag instance url.

### DIFF
--- a/apps/workers/feedWorker.ts
+++ b/apps/workers/feedWorker.ts
@@ -126,6 +126,94 @@ async function run(req: DequeuedJob<ZFeedRequestSchema>) {
   });
   const unparseFeedData = await parser.parseString(xmlData);
 
+  // Check if this is a Wallabag feed
+  const isWallabagFeed = unparseFeedData.items.some(
+    (item) =>
+      item.id && typeof item.id === "string" && item.id.includes("wallabag:"),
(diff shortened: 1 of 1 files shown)


MohamedBassem (changes requested):
Hey, @yinan-c appreciate taking the time and sending a PR to fix the wallabag issue. I'm also really sorry with the delay in the review.

However, honestly, I'm not a big fan of hardcoding wallabag specific logic. Is what you're introducing in this PR generic (e.g. part of the RSS spec), or is this specific to wallabag? Also, parsing the RSS XML with regexes is a bit fragile. Is there no better way for doing so?

PR #1195, opened 2025-04-04
Enable Ukrainian language

denysdovhan:
As I see in the codebase, Hoarder has already been translated to Ukrainian: https://github.com/hoarder-app/hoarder/blob/main/apps/web/lib/i18n/locales/uk/translation.json

But it's not possible to select Ukrainian language from the UI.


MohamedBassem:
Thank you!

PR #1196, opened 2025-04-05
ui(web): Clicking view original now opens in a blank tab

### LINKED ISSUE #1114: [FR] Clicking "View Original" in bookmark details should open in a blank tab

To be consistent with the behavior of clicking on a link in the main Hoarder page, if I click to see the details popup for an item and then click "View Original", I believe this should open in a new tab (`_blank` )

UX consistency when viewing a link at its source.  That way, when I'm done reading, I can hit ⌘W to close the tab in both situations without losing my spot in the Hoarder UI.

Yes, I can train myself to always hold down the ⌘ key when opening links so that the behavior is always consistent.

### Additional context

_No response_

ekambains:
Fix #1114 
Added target="_blank" attribute to the Link tag.
Now on clicking view original in bookmark details opens in a new tab.

### DIFF
--- a/apps/web/components/dashboard/preview/BookmarkPreview.tsx
+++ b/apps/web/components/dashboard/preview/BookmarkPreview.tsx
@@ -120,6 +120,7 @@ export default function BookmarkPreview({
           {sourceUrl && (
             <Link
               href={sourceUrl}
+              target="_blank"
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thank you!

PR #1209, opened 2025-04-06
ui(web): Copy link button greyed out when clipboard not available

### LINKED ISSUE #846: Copy Link menu in bookmark options is not greyed out when clipboard is not available
### Describe the Bug

The context menu called "Copy Link" (which is supposed to copy the link attached to the note) is not working. If it is clicked, then nothing happens (and the context menu does not disappear).

### Steps to Reproduce

1. Create a new note providing a link to an external website
2. Click on the three dots in the bottom right part of the tile and select "Copy Link"

### Expected Behaviour

The link is copied into the clipboard (and the context menu disappears)

This issue seems to be due to an issue in Javascript. See below the error which is raised in JS when the "Copy Link" button is pressed:

image

### Device Details

Google Chrome on Windows Client

### Exact Hoarder Version

v0.21.0

- [X] I have checked the troubleshooting guide and I haven't found a solution to my problem

ekambains:
Fix #846 
Now the copy link button is greyed out(disabled) when clipboard is unavailable.

### DIFF
--- a/apps/web/components/dashboard/bookmarks/BookmarkOptions.tsx
+++ b/apps/web/components/dashboard/bookmarks/BookmarkOptions.tsx
@@ -48,6 +48,7 @@ export default function BookmarkOptions({ bookmark }: { bookmark: ZBookmark }) {
   const linkId = bookmark.id;
 
   const demoMode = !!useClientConfig().demoMode;
+  const isClipboardAvailable = !!navigator.clipboard;
 
   const { setOpen: setTagModalIsOpen, content: tagModal } =
     useTagModel(bookmark);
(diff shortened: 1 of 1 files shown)


apps/web/components/dashboard/bookmarks/BookmarkOptions.tsx
```diff
@@ -48,6 +48,7 @@ export default function BookmarkOptions({ bookmark }: { bookmark: ZBookmark }) {
   const linkId = bookmark.id;
 
   const demoMode = !!useClientConfig().demoMode;
+  const isClipboardAvailable = !!navigator.clipboard;
```
MohamedBassem: Don't we need to do `navigator && !!navigator.clipboard` as well? Otherwise it might crash during SSR?

ekambains: Yes, you are absolutely right. I am very sorry for missing this, i will fix it right now.

ekambains:
@MohamedBassem, I made the requested changes.

MohamedBassem:
Thanks @ekambains!

PR #1218, opened 2025-04-07
 docs: add Arch Linux migration instructions

devome:
docs: add Arch Linux migration instructions


docs/docs/02-Installation/03-archlinux.md
```diff
@@ -43,6 +43,48 @@
...
+**2. Uninstall Hoarder**  
+After uninstalling, you can manually remove the old `hoarder` user and group if needed.
+```shell
+paru -R hoarder
+```
+
+**3. Rename the old data directory**
+```shell
+sudo mv /var/lib/hoarder /var/lib/karakeep
+```
```
MohamedBassem, line 72: Shouldn't we ask people to move their env file as well? Otherwise they might lose their configuration?

docs/docs/02-Installation/03-archlinux.md
```diff
@@ -43,6 +43,53 @@
...
+```shell
+sudo chown -R karakeep:karakeep /var/lib/karakeep
+```
+
+**6. Set Karakeep**  
+Edit `/etc/karakeep/karakeep.env` according to [configuration page](/configuration). **The environment variables that are not specified in `/etc/karakeep/karakeep.env` need to be added by yourself.**
+
+Or you can copy old hoarder env file to karakeep:
+```shell
+sudo cp -f /etc/hoarder/hoarder.env /etc/karakeep/karakeep.env
```
MohamedBassem, line 89: Do you know if the env file needs to be owned by the karakeep user as well or does systemd take care of applying it?

PR #1219, opened 2025-04-07
hoarder-linux.sh > karakeep-linux.sh v2.1

vhsdream:
- Adds a migrate function so users can convert their Hoarder install to a Karakeep install
- The function will also check for an update after finishing the migration
- Various fixes and optimisations
- Update docs with migration info
- Replace mentions of Hoarder with Karakeep in Debuntu install page


MohamedBassem:
Thanks @vhsdream for being an amazing maintainer for this script!

PR #1227, opened 2025-04-09
move auth redirect to middleware

jaworek:
Add middleware and setup next-auth to avoid having multiple places in the app to protect against unauthorized access.

https://next-auth.js.org/configuration/nextjs#middleware

Test steps:
1. Open app at "localhost:3000"
2. Verify that you get redirected to "/signin" page
3. Login with your credentials or singup with new account
4. Verify that you get redirected to "/dashboard/bookmarks"
5. Try to access "/signin" page and verify that you get redirected to "/dashboard/bookmarks" while logged it
6. Logout and verify that you're redirected to "/signin" page


apps/web/middleware.ts
```diff
@@ -0,0 +1,6 @@
+import { withAuth } from "next-auth/middleware";
+import { pages } from "./server/auth";
+
+export default withAuth({
+  pages,
```
jaworek, line 6: I had to extract `pages` into separate variable and pass it again in middleware to avoid infinite redirect as mentioned in NextAuth documentation.

https://next-auth.js.org/configuration/nextjs#pages

image

MohamedBassem:
Hey, @jaworek thanks for taking the time to send the PR! I think the changes in general make sense, however, NextJs recently had a CVE where an attacker could bypass middlewares ([link](https://vercel.com/blog/postmortem-on-next-js-middleware-bypass)) and they don't recommend using middlewares as the sole auth mechanism. I understand that in our case it's not given that most interactions are via trpc which itself is authed, but still I like the comfort of ensuring that the user is authenticated manually and guaranteeing that by typescript's `session` checks. The CVE is leaving a bad taste in my mouth basically :D

Given that we don't really do a lot of auth checks anyways (mostly only in the root layouts), I honestly, don't see this as a problem worth solving. I hate not accepting work that people put effort into, but as I mentioned in the our [contribution guidelines](https://github.com/karakeep-app/karakeep/blob/main/CONTRIBUTING.md) it would be better to align first before picking up more work to avoid such things. Sorry again :(

jaworek:
Hey, @MohamedBassem thanks for the reply. I agree with you points. If I come up with a better solution that allows keeping auth checks from specific components, I'll open another PR. Thanks for creating such an awesome project, I'm using it everyday and it meets all of my needs :D

Issue #1060, opened 2025-02-21
merging lists


Fuckingnameless:
### Describe the feature you'd like

the way i found to update/sync the bookmarks in hoarder was import them again, but that gives me 2 lists with thousands of duplicate links, please add the option to merge them so i can deduplicate

Thank you.

### Describe the benefits this would bring to existing Hoarder users

convenience when having too many bookmarks

### Can the goal of this request already be achieved via other means?

maybe

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

gomnitrix:
Hi @MohamedBassem @Fuckingnameless,

I’d love to work on this issue and contribute to Hoarder. Here’s my proposed solution:

**Implementation Plan**

- Add a "Select" icon next to the "+" icon in the sidebar.

- Enable batch selection mode, displaying checkboxes for each list.

- Show a floating action bar with "Merge" (and optionally "Cancel") upon selection.

- When merging, prompt the user to either:

  - Select an existing target list, or

  - Create a new list, with automatic deduplication of content.

**Scope Consideration**
As a first-time contributor and new to this project’s tech stack, I would like starting with flat merging of nested lists for simplicity. We can later enhance this with more granular controls if needed.

What do you think about this approach? If approved, I’d be happy to start working on it!

MohamedBassem:
@gomnitrix Thanks for picking this up, regarding the approach, I'd keep it simple:
1. Inside the list view, there's a list options drop down menu. When you click on this, you get "edit", "delete", and we add "merge". This is simpler than "bulk" merging.
2. When you click merge, a dialog opens, and you select another list to merge into. There's no value in creating a new "list" during merging in my opinion.

Regarding nested lists, you don't have to worry about that. When merging two lists, just move the items from the source list id to the destination list id. Nested lists don't show the content of their children anyways.

MohamedBassem:
Done in #1231

MohamedBassem:
@Fuckingnameless Karakeep itself deduplicates links on addition. So you can't have "exactly" duplicate bookmarks, and therefore merging lists shouldn't dedup. Notice the emphasis on exactly, karakeep only dedups "exact" links. A difference in slashes at the end of the URL is considered a different URL (at least that's how things are today). So my assumptions is that your links are not "unique" from karakeep's perspective in the first place.

PR #1231, opened 2025-04-10
feat(web): Add support for merging lists

gomnitrix:
This PR implements the feature requested in #1060, enabling users to merge contents from one list into another selected non-smart list with deduplication functionality.

#### Key Changes
- Added new merge functionality with deduplication between lists in trpc/models
- Added corresponding UI components to support the merge feature (see screenshot below)

#### Screenshot
<div style="display: flex; flex-direction: column; gap: 10px;">
  Merge list interface
  Merge confirmation dialog
</div>

#### Internationalization (i18n) Notes
New form labels have been added with translations for:
 - English (en)
 - Chinese (zh)
 
Translations for other supported languages are currently missing. I’m unsure if it’s expected for contributors to provide full translations, or if there’s a separate process for translations. I'm happy to add AI-generated translations if that's acceptable--just let me know.

#### Related Issue Discovery
During implementation, I noticed an edge case in how text-type bookmarks are handled during import, and it could affect the expected user flow:
When importing bookmarks from files that exported from this app, the system creates new text-type bookmarks with identical content but different bookmark IDs in each import. In a word, the import function does not deduplicate text-type bookmarks. 

This leads to:
  - Multiple imported lists containing duplicate bookmarks with different IDs.
  - The merge function is unable to deduplicate these bookmarks, as it relies on bookmark IDs for comparison.
  - That might break the ideal workflow described in #1060, where users re-import to sync updates and expect merging to help clean up duplicates.
    
If this behavior isn’t intended, I think it might make sense to track it as a separate issue.

### DIFF
--- a/apps/web/components/dashboard/lists/MergeListModal.tsx
+++ b/apps/web/components/dashboard/lists/MergeListModal.tsx
@@ -22,12 +22,13 @@ import { Input } from "@/components/ui/input";
 import { Switch } from "@/components/ui/switch";
 import { toast } from "@/components/ui/use-toast";
 import { useTranslation } from "@/lib/i18n/client";
-import { useMergeLists } from "@hoarder/shared-react/hooks/lists";
-import { ZBookmarkList, zMergeListSchema } from "@hoarder/shared/types/lists";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { X } from "lucide-react";
 import { useForm } from "react-hook-form";
 import { z } from "zod";
+
+import { useMergeLists } from "@karakeep/shared-react/hooks/lists";
+import { ZBookmarkList, zMergeListSchema } from "@karakeep/shared/types/lists";
 
 import { BookmarkListSelector } from "./BookmarkListSelector";
 
--- a/apps/web/components/dashboard/lists/MergeListModal.tsx
+++ b/apps/web/components/dashboard/lists/MergeListModal.tsx
@@ -1,0 +1,231 @@
+import { useEffect, useState } from "react";
+import { ActionButton } from "@/components/ui/action-button";
+import { Button } from "@/components/ui/button";
+import {
+  Dialog,
+  DialogClose,
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+  DialogTrigger,
+} from "@/components/ui/dialog";
+import {
+  Form,
+  FormControl,
+  FormField,
+  FormItem,
+  FormLabel,
+  FormMessage,
+} from "@/components/ui/form";
+import { Input } from "@/components/ui/input";
+import { Switch } from "@/components/ui/switch";
+import { toast } from "@/components/ui/use-toast";
+import { useTranslation } from "@/lib/i18n/client";
+import {
+  useDeleteBookmarkList,
+  useMergeLists,
+} from "@hoarder/shared-react/hooks/lists";
+import { ZBookmarkList, zMergeListSchema } from "@hoarder/shared/types/lists";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { X } from "lucide-react";
+import { useForm } from "react-hook-form";
+import { z } from "zod";
+
+import { BookmarkListSelector } from "./BookmarkListSelector";
+
+export function MergeListModal({
+  open: userOpen,
+  setOpen: userSetOpen,
+  list,
+  children,
+}: {
+  open?: boolean;
+  setOpen?: (v: boolean) => void;
+  list: ZBookmarkList;
+  children?: React.ReactNode;
+}) {
+  const { t } = useTranslation();
+  if (
+    (userOpen !== undefined && !userSetOpen) ||
+    (userOpen === undefined && userSetOpen)
+  ) {
+    throw new Error("You must provide both open and setOpen or neither");
+  }
+  const [customOpen, customSetOpen] = useState(false);
+  const form = useForm<z.infer<typeof zMergeListSchema>>({
+    resolver: zodResolver(zMergeListSchema),
+    defaultValues: {
+      sourceId: list?.id ?? "",
+      targetId: "",
+      deleteAfterMerge: true,
+    },
+  });
+  const [open, setOpen] = [
+    userOpen ?? customOpen,
+    userSetOpen ?? customSetOpen,
+  ];
+
+  useEffect(() => {
+    form.reset({
+      sourceId: list?.id ?? "",
+      targetId: "",
+      deleteAfterMerge: true,
+    });
+  }, [open]);
+
+  const { mutate: mergeLists, isPending: isMerging } = useMergeLists({
+    onSuccess: () => {
+      toast({
+        description: t("toasts.lists.merged"),
+      });
+      setOpen(false);
+      form.reset();
+    },
+    onError: (e) => {
+      if (e.data?.code == "BAD_REQUEST") {
+        if (e.data.zodError) {
+          toast({
+            variant: "destructive",
+            description: Object.values(e.data.zodError.fieldErrors)
+              .flat()
+              .join("\n"),
+          });
+        } else {
+          toast({
+            variant: "destructive",
+            description: e.message,
+          });
+        }
+      } else {
+        toast({
+          variant: "destructive",
+          title: t("common.something_went_wrong"),
+        });
+      }
+    },
+  });
+
+  const { mutate: deleteList, isPending: isDeleting } = useDeleteBookmarkList({
+    onSuccess: () => {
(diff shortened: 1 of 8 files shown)


apps/web/components/dashboard/lists/MergeListModal.tsx
```diff
@@ -0,0 +1,247 @@
...
+  open: userOpen,
+  setOpen: userSetOpen,
+  list,
+  prefill,
+  children,
+}: {
+  open?: boolean;
+  setOpen?: (v: boolean) => void;
+  list?: ZBookmarkList;
+  prefill?: ZBookmarkList;
```
MohamedBassem: Prefill is only needed when we're reusing the dialog for both editing and creation. In this case, we can just drop it completely.

apps/web/components/dashboard/lists/MergeListModal.tsx
```diff
@@ -0,0 +1,247 @@
...
+export function MergeListModal({
+  open: userOpen,
+  setOpen: userSetOpen,
+  list,
+  prefill,
+  children,
+}: {
+  open?: boolean;
+  setOpen?: (v: boolean) => void;
+  list?: ZBookmarkList;
```
MohamedBassem: The source list in this case is a required field?

apps/web/components/dashboard/lists/MergeListModal.tsx
```diff
@@ -0,0 +1,247 @@
...
+                control={form.control}
+                name="sourceIcon"
+                render={({ field }) => (
+                  <FormItem>
+                    <FormControl>
+                      <Popover>
+                        <PopoverTrigger className="h-full rounded border border-input px-2 text-2xl">
+                          {field.value}
+                        </PopoverTrigger>
+                      </Popover>
```
MohamedBassem: Does this need to be a popover? I assume it can just be a normal `<span>` or something?

apps/web/components/dashboard/lists/MergeListModal.tsx
```diff
@@ -0,0 +1,247 @@
...
+  ) {
+    throw new Error("You must provide both open and setOpen or neither");
+  }
+  const [customOpen, customSetOpen] = useState(false);
+  const form = useForm<z.infer<typeof zMergeListSchema>>({
+    resolver: zodResolver(zMergeListSchema),
+    defaultValues: {
+      listId: list?.id ?? prefill?.id ?? "",
+      sourceName: list?.name ?? prefill?.name ?? "",
+      sourceIcon: list?.icon ?? prefill?.icon ?? "🚀",
```
MohamedBassem: Source name and icon should probably not be part of the form and get rendered as normal text instead of `FormField`s

packages/shared/types/lists.ts
```diff
@@ -74,3 +74,26 @@ export const zEditBookmarkListSchemaWithValidation = zEditBookmarkListSchema
       "Smart lists cannot have unqualified terms (aka full text search terms) in the query",
     path: ["query"],
   });
+
+export const zMergeListSchema = z
+  .object({
+    sourceName: z.string(),
+    sourceIcon: z.string(),
```
MohamedBassem: I don't think the source name and icon should be part of the request input schema

packages/shared/types/lists.ts
```diff
@@ -74,3 +74,26 @@ export const zEditBookmarkListSchemaWithValidation = zEditBookmarkListSchema
...
+    listId: z.string().min(1, "Source list must be selected"),
```
MohamedBassem: Let's maybe call this `sourceId` instead?

packages/trpc/routers/lists.ts
```diff
@@ -39,6 +41,13 @@ export const listsAppRouter = router({
     .mutation(async ({ input, ctx }) => {
       return await ctx.list.update(input);
     }),
+  merge: authedProcedure
+    .input(zMergeListSchema)
+    .output(zMergeListResponseSchema)
+    .use(ensureListOwnership)
```
MohamedBassem: This is missing an ownership check on the destination list.

You might have to call it inside the mutation function itself.

packages/trpc/models/lists.ts
```diff
@@ -75,6 +77,52 @@ export abstract class List implements PrivacyAware {
     return this.fromData(ctx, result);
   }
 
+  static async merge(
```
MohamedBassem: This doesn't need to be static. It can be a member function the list called `mergeInto`

packages/trpc/models/lists.ts
```diff
@@ -75,6 +77,52 @@ export abstract class List implements PrivacyAware {
...
+    ctx: AuthedContext,
+    input: z.infer<typeof zMergeListSchema>,
+  ): Promise<z.infer<typeof zMergeListResponseSchema>> {
+    const sourceList = await List.fromId(ctx, input.listId);
+    const targetList = await List.fromId(ctx, input.targetId);
+    if (targetList.type !== "manual") {
```
MohamedBassem: Shouldn't we also check that the source list is not a smart list? I feel like it doesn't make much sense merging a smart list into a manual one?

packages/trpc/models/lists.ts
```diff
@@ -75,6 +77,52 @@ export abstract class List implements PrivacyAware {
...
+              e instanceof TRPCError &&
+              e.message.includes("already in the list")
+            ) {
+              duplicateCount++;
+            } else {
+              failedCount++;
+            }
+          }),
+      ),
+    );
```
MohamedBassem: I'm curious, why are we moving items one by one instead of doing it in one db query.

Something along the lines of:
```
update bookmarkInLists set listId='target' where listId='source' on conflict do nothing.
```

packages/trpc/models/lists.ts
```diff
@@ -75,6 +77,52 @@ export abstract class List implements PrivacyAware {
...
+
+    return {
```
MohamedBassem: Should we delete the source list when the merge is completed?

MohamedBassem (changes requested):
Thanks a lot for the PR! I've requested a couple of changes :)

apps/web/components/dashboard/lists/MergeListModal.tsx
gomnitrix: Got it, removed

apps/web/components/dashboard/lists/MergeListModal.tsx
gomnitrix: Yes, updated to required—missed that one, thanks!

apps/web/components/dashboard/lists/MergeListModal.tsx
gomnitrix: make sense, I've simplified it.

apps/web/components/dashboard/lists/MergeListModal.tsx
gomnitrix: Agreed, removed from the form now

packages/shared/types/lists.ts
gomnitrix: removed

packages/trpc/routers/lists.ts
gomnitrix: Done—now checking ownership of both sourceList and targetList inside the mutation.

packages/trpc/models/lists.ts
gomnitrix: You’re right—I changed it to bulk insert. I used insert instead of update because deleting the source list after merging is now the default but optional. If the user chooses to keep it, preserving its bookmarks could be useful.

packages/trpc/models/lists.ts
gomnitrix: Good point—I’ve added a switch (default on) to delete the source list after merging. Attached a screenshot in this comment—let me know if you have any UI suggestions.
CleanShot 2025-04-15 at 01 59 17@2x

packages/trpc/models/lists.ts
gomnitrix: changed to abstract method.

packages/trpc/models/lists.ts
gomnitrix: fixed

gomnitrix:
> Thanks a lot for the PR! I've requested a couple of changes :)
Thanks for your review! I’ve made the requested changes. If you spot any other issues, please let me know.

gomnitrix:
@MohamedBassem Thanks! Moving the deletion to the backend totally makes sense. I’m still getting familiar with the stack and its best practices, so really appreciate your reviews and tweaks — learned a lot from them.
And yep, those screenshots are from CleanShot 😄

PR #1239, opened 2025-04-12
i18n: Add Korean language

jkywalker:
### Summary

- Added `"ko": "Korean"` to `langNameMappings` in `langs.ts`.
- Korean translation file already exists in `locales/ko/translation.json`.

This allows users to select Korean in the language dropdown from the settings page.


MohamedBassem:
Thank you!

PR #1247, opened 2025-04-12
Replace hoarder with karakeep in github issue templates

Mxrk:
While creating an issue yesterday I noticed that it still contains the old name. 
I replaced hoarder with karakeep in all three github templates.


MohamedBassem:
Thanks for the help, I still have a long way to go with this rebranding :D

PR #1250, opened 2025-04-12
Changed Hoarder to Karakeep

darxreborn:
Including links


MohamedBassem:
Hey, thanks for the PR, but you also need to change `docs/docs/02-Installation/02-unraid.md`. Otherwise those changes will be reverted in the next docs release.

PR #1257, opened 2025-04-13
fix: fit popover content box to emoji picker size

Mxrk:
Currently when adjusting the emoji for a list, the Popover-Content box is smaller than the actual emoji picker "box" as seen on the screenshot:
image

I fixed this problem with `PopoverContent className="w-auto">` which was also used in https://github.com/karakeep-app/karakeep/blob/cf97bace33fdd14f29ce947d55d17cba8fa85c11/apps/web/components/dashboard/bookmarks/EditBookmarkDialog.tsx#L283

With this fix it looks like this:
image


MohamedBassem:
Thank you!

Issue #1189, opened 2025-04-02
Oauth auto login


dregini:
### Describe the feature you'd like

if i set
DISABLE_SIGNUPS true
DISABLE_PASSWORD_AUTH true
and i have a Oauth Provider configured the login flow should start automatically redirecting to the identity provider instead of showing a landing page with the 2 phrases "Password authentication is currently disabled." and "Signups are currently disabled.".

as an alternative implement an additional variable to skip the landing page and start the login flow redirecting to the Identity provider

### Describe the benefits this would bring to existing Hoarder users

It removes unnecessary user friction by skipping a useless landing page.
Improves user experience for apps that rely only on external authentication.

### Can the goal of this request already be achieved via other means?

No

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

SandraCHC:
+1 :)
Maybe also possible to hide the login form

MohamedBassem:
@claude implement this and gate it with a client setting

Issue #411, opened 2024-09-23
Bug: OIDC Account Linking Issue with Case-Sensitive Emails


IamTaoChen:
I have a local account with the email `AaBbCcDd@example.com`, and when I use OIDC with `OAUTH_ALLOW_DANGEROUS_EMAIL_ACCOUNT_LINKING = true`, it doesn't link to the same account because the email from OIDC is `aabbccdd@example.com`.

kamtschatka:
This is not really an issue with OIDC per se.
This also happens when you create accounts locally with "Asdf@example.com" and "asdf@example.com". We'll have to discuss how to solve this issue internally, but basically email addresses are case insensitive, so they should all be 1 account.

MohamedBassem:
@IamTaoChen As @kamtschatka said, unfortunately currently we do email comparisons in a case sensitive way. It might take us some effort to change this behavior, so we might not have enough time to get to it soon to be honest. If you want, we can help you modify your email in the database so that linking can work :)

JereWoo:
@MohamedBassem 

For my own edification. Can you offer insight into the challenges delaying resolution here? Beyond the obvious resource/time constraints that come with working on an open source project. 

This seems like a simple matter of normalizing the email to lowercase or similar before making the comparison. If its just time I may be able to assist -- if you're open to it as this seems a doable task for me. 

Also it appears when registering with OIDC  new accounts are created using email the claim for both Name and Email. Would likely be best using Username, Name claim for this and with failback to email if claim not provided.

Eragos:
IMO it worked as designed ans is not a bug (check out [RFC 2822](https://www.rfc-editor.org/rfc/rfc2822.html#section-2.2), section 2.2 for more details of the correct email address format)

Issue #410, opened 2024-09-23
OAuth login failed: No profile


chain710:
![image](https://github.com/user-attachments/assets/12d37683-0827-4607-9c3b-b93db076449d)

My oauth provider: authelia 4.38.12

- click `Sign in with authelia` redirect to authelia sso page
- click `accept` redirect back to hoarder login page, but got error `OAuth login failed: No profile`

my authelia instance works well with other self-host services like linkwarden and immich

found no useful log via `docker logs hoarder`

```
....

2024-09-23T14:47:13.944Z info: Workers version: 0.17.0
2024-09-23T14:47:13.954Z info: [Crawler] Connecting to existing browser instance: http://chrome:9222
(node:121) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
2024-09-23T14:47:13.960Z info: [Crawler] Successfully resolved IP address, new address: http://172.20.0.29:9222/
2024-09-23T14:47:15.981Z info: Starting crawler worker ...
2024-09-23T14:47:15.981Z info: Starting inference worker ...
2024-09-23T14:47:15.981Z info: Starting search indexing worker ...
```

env 

```
REDIS_HOST=redis
REDIS_DB_IDX=15
DATA_DIR=/data
NEXTAUTH_SECRET=secret
BROWSER_WEB_URL=http://chrome:9222
MEILI_ADDR=http://meilisearch:7700
MEILI_NO_ANALYTICS=true
MEILI_MASTER_KEY=key
OPENAI_API_KEY=key
INFERENCE_IMAGE_MODEL=gpt-4o-mini
INFERENCE_TEXT_MODEL=gpt-4o-mini
OAUTH_WELLKNOWN_URL=https://auth.example.com/.well-known/openid-configuration
OAUTH_CLIENT_SECRET=secret
OAUTH_CLIENT_ID=hoarder
OAUTH_PROVIDER_NAME=authelia
NEXTAUTH_URL=https://h.example.com
```

kamtschatka:
this happens, when authelia does not provide an email address or a name for the user you are using: https://github.com/hoarder-app/hoarder/blob/main/apps/web/server/auth.ts#L150

you'll have to check on authelia side to fill those in.

MohamedBassem:
@kamtschatka I think we should probably remove the name restriction. If it's empty, we can just use the email and not fail the login on it.

[...]

PR #1268, opened 2025-04-15
Auth/case insensitive email

JereWoo:
This pull request addresses:

[Issue #410 (comment)](https://github.com/karakeep-app/karakeep/issues/410#issuecomment-2740961459)

[Issue #411 (comment)](https://github.com/karakeep-app/karakeep/issues/411#issuecomment-2744976404)

[Issue #1189](https://github.com/karakeep-app/karakeep/issues/1189)

✅ What’s fixed:
Email matching now uses case-insensitive comparisons. 
OAuth no uses userinfo so it will much more reliably have name/nickname/username claims
DangerousAccountLinking is useable if you capitalized your email at account creation!

📌 Why this matters:
Email is technically case-sensitive, but in practice, treating it as case-insensitive is expected behavior.
Casing mismatches from normalizing email in database can cause subtle bugs and unexpected failures especially when using foreign characters. 

ℹ️ Notes:
Unique index ensure no duplicate emails 
Users still see their email as they entered it — this change only affects backend logic.
This is my first real PR to this repo — happy to revise anything that needs improvement. If i did anything poorly let me know I'll do better next time. 

Some changes will break things: 
WellknownAddress is excepted to point to well known address
Schema change to unique index will cause break if different cases of the "same" email are present.

### DIFF
--- a/packages/trpc/auth.ts
+++ b/packages/trpc/auth.ts
@@ -1,8 +1,9 @@
 import { randomBytes } from "crypto";
 import * as bcrypt from "bcryptjs";
+import { sql } from "drizzle-orm";
 
 import { db } from "@karakeep/db";
-import { apiKeys } from "@karakeep/db/schema";
+import { apiKeys, users } from "@karakeep/db/schema";
 import serverConfig from "@karakeep/shared/config";
 import { authFailureLogger } from "@karakeep/shared/logger";
 
@@ -84,8 +85,11 @@ export async function validatePassword(email: string, password: string) {
   if (serverConfig.auth.disablePasswordAuth) {
     throw new Error("Password authentication is currently disabled");
   }
+
+  const normalizedEmail = email.toLowerCase();
   const user = await db.query.users.findFirst({
-    where: (u, { eq }) => eq(u.email, email),
+    //    where: (u, { eq }) => eq(u.email, email),
+    where: sql`LOWER(${users.email}) = ${normalizedEmail}`,
   });
 
   if (!user) {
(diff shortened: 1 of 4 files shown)


JereWoo:
> Assuming that an email address is case-sensitive is a bit of a mistake. You can check out [RFC 2822](https://www.rfc-editor.org/rfc/rfc2822.html#section-2.2), section 2.2 for more details on the correct format.
> 
> Best Michael


Hey Michael!

I completely agree, that's the point of the pull request. I was very tired when I wrote this and did not explain that well. I've updated the original post for clarity. **Currently karakeep treats the email as case sensitive** which causes issues with: oAuth, account-linking, or just remembering what email case you used to login.

packages/trpc/auth.ts
```diff
@@ -84,8 +85,11 @@ export async function validatePassword(email: string, password: string) {
   if (serverConfig.auth.disablePasswordAuth) {
     throw new Error("Password authentication is currently disabled");
   }
+
+  const normalizedEmail = email.toLowerCase();
   const user = await db.query.users.findFirst({
-    where: (u, { eq }) => eq(u.email, email),
+    //    where: (u, { eq }) => eq(u.email, email),
```
mikhail5555, line 91:
```
where: (u, { eq, lower }) => eq(lower(u.email), lower(email))
```
Should do the same are you are trying to achieve, without having to use the `sql`

JereWoo:
@MohamedBassem thoughts on this?

MohamedBassem:
It's unlikely that we'll be able to change this behavior any time soon as it's backward incompatible (specially the addition of the unique index).

MohamedBassem:
I'm also planning to switch to better-auth soon, so don't want to add further deps on next-auth. I also want to apologize for how long it took me to get to this PR :) Please let's make sure we're aligned on what you're planning to work on (as per the contribution guidelines) to avoid wasting any effort. Sorry one more time.

PR #1274, opened 2025-04-17
FIxed sidebar link to github

hasansino:
A. v${serverConfig.serverVersion} for docker images is translated to `vnightly` which is absent.

B. Link points to hoarder, which redirects after, probably legacy.

C. For me, it is more convenient to go directly to repository home and navigate from there.

### DIFF
--- a/apps/web/components/shared/sidebar/Sidebar.tsx
+++ b/apps/web/components/shared/sidebar/Sidebar.tsx
@@ -32,7 +32,11 @@ export default async function Sidebar({
       </div>
       {extraSections}
       <Link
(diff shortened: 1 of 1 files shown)


MohamedBassem (changes requested):
@hasansino It translates to `vnightly` only if you're on the nightly version. It links to the correct release if you're on one of the stable releases. We can check if the version is `== nightly`, then point to the root of the repo, else keep the release version link?

hasansino:
@MohamedBassem Agree, could you please verify the change as I am not proficient in frontend.

MohamedBassem:
Thank you!

PR #1296, opened 2025-04-21
Feat/playwright only

maelp:


element0xE:
A full migration to Playwright should be seriously considered, as the current Puppeter-stealth / chrome approach is easily detectable and out of date , resulting in failed crawls.

Patchright might be considered: https://github.com/Kaliiiiiiiiii-Vinyzu/patchright
In the Patchright README.md under the Stealth section there is a list of headless/bot detectors , Karakeep doesn’t really pass those tests.

Issue #414, opened 2024-09-23
Bypassing cookie and GDPR banner


Dwelled2593:
When i use hoarder on a youtube link, the crawler get stuck with the cookie banner, any idea on how to solve this ? 

![image](https://github.com/user-attachments/assets/0c50b791-6abc-45f0-9894-4b81e08da983)

CrypticC3s4r:
@Dwelled2593 can you provide some example links ?

pix:
![image](https://github.com/user-attachments/assets/08fbca24-15e7-40d2-b7f1-f333e160f04f)

From: https://www.youtube.com/watch?v=E-5b1iGNraM (Probably a EU thing)

ctschach:
Okay, I'm running into the same issue.

One way to get around this would be the ability to provide a cookie file which is used when crawling sites. With this in place, YT should know that the cookies are already excepted. You can use a chrome extension called "Get cookies.txt locally" to get the required cookies.  This would probably be helpful for other sites to.

cakonopka:
Got the same issue with Instagram Posts or X Posts

fcorvelo:
I don't know if this is the best solution, but I managed to get this to work using the "i-dont-care-about-cookies" Chrome extension. This extension automatically accepts and hides all the cookie banners.

The way I made this work was to make the following changes in the Chrome container:

1. Use the `--headless=new` flag to allow loading extensions in headless mode.
2. Create a bind mount between a folder on the host containing the unzipped extension code and some folder inside the container.
3. Use the `--load-extension=/i-dont-care-about-cookies` flag with the path to the previously mounted folder to tell Chrome from where to load the extension.

And that's it! This way, when Hoarder calls Chrome it will load the extension and the extension will accept all the cookie banners, making them disappear. 

One caveat (and this is why I say that maybe this is not the best solution) is that enabling the `--headless=new` flag creates the following error on the Hoarder web container that keeps appearing in the logs every 5 seconds. I didn't see any issues using Hoarder with this error, but perhaps there is something that I'm not aware of.

`[Crawler] Failed to connect to the browser instance, will retry in 5 secs: TypeError: Failed to fetch browser webSocket URL from http://172.18.0.3:9222/json/version: fetch failed`

So, I hope that this can help while there is no official way of doing it 😃

kassyss:
@fcorvelo you made my day « ! » 
[...]

PR #1297, opened 2025-04-21
Feat/playwright chromium vnc

maelp:
This should be merged on top of the playwright branch, to add the VNC + chromium + cookie sharing

if you deploy this, you would need to secure the VNC access using VNC_PASSWORD=some-strong-password so that people can't connect to it (or don't expose it and just use a SSH tunnel / Tailscale to access it)


thiswillbeyourgithub:
Just a thought: i thought there were stacks that allow connecting to a vnc like session using the browser. That might be better and more accessible: instead of installing a vnc client you use your own browser to connect to that session. Wouldn't that be better?

SettingDust:
What's the status of this one?

llCorvinSll:
@MohamedBassem i think, what adding cookie sharing solve this issue #414 users (at least in self host mode) can "previsit" most used sites via vnc and don`t struggle about scraping popups

PR #1302, opened 2025-04-22
feat (workers): adding a local metascraper plugin for Reddit posts

birnam:
In the past, the preview images for bookmarks from Reddit links were poorly chosen. Reddit does not use opengraph tags, so metascraper-images simply looked for all images on the page and returned the first. This tended to be the profile picture for the poster for the Reddit link.

This new plugin, using the existing metascraper framework, provides a better selection of image for the bookmark when the URL domain is 'reddit'.                                                                 
                                                                          
In addition, recent changes (I believe this was a side effect of adding the metascraper-author and/or the metascaper-publisher plugins, but it could also be related to the metascraper-readibility plugin) broke what used to be a good choice of bookmark title. Previously, titles looked like 'Tinyauth just reached 1000 stars! : r/selfhosted' with both thread title and subreddit mentioned. After this update, all Reddit posts now have the same title: 'The heart of the internet'.

To return to the better format, this new metascraper-reddit plugin now attempts to retrieve the better title from reddit URLs. Note that in order to gain precedence in title selection, the 'metascraperReddit()' inclusion in the crawlerWorkers.ts metascraper instantiation list had to be moved above metascraperReadability().


apps/workers/metascraper-plugins/metascraper-reddit.ts
```diff
@@ -0,0 +1,112 @@
...
+ * Post Title for a Reddit post instead of use the
+ * default.
+ *
+ * As of writing this, Reddit posts do not define
+ * an open-graph image (og:image) attribute, so
+ * metascraper resorts to looking for images in
+ * the HTML DOM, and selects the first one.
+ *
+ * In Reddit posts, the first image is typically
+ * the profile picture of the OP, which Hoarder
```
Mxrk: Replace Hoarder here

birnam: good catch, I missed that one

birnam:
My merge request into metascraper was accepted. I've updated the versions to use the latest which means I could remove the @ignore line that expected the typecheck error. I also updated a path to reflect the new directory structure.

MohamedBassem:
I tried it, and it works perfectly, thanks a lot!

PR #1303, opened 2025-04-22
build(Debian): use nightly build of yt-dlp in installation script

vhsdream:
- It's the [recommended build](https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#update)
- The script will update with `yt-dlp -U` if on Bookworm


vhsdream:
@MohamedBassem I'm closing this PR and making a new one with this change as well as adding the MCP server build commands. Is it correct to assume that the MCP server will require a systemd service of it's own?

PR #1316, opened 2025-04-26
chore: rename docs hoarder to karakeep

ekambains:
Renamed a file from hoarder to karakeep


MohamedBassem:
With the new release, the 0.24 docs have this fixed. Thanks for the PR though.

PR #1329, opened 2025-04-28
build(Debian): add mcp build; use nightly yt-dlp build; update system

vhsdream:
- Adds build command for the new MCP server feature 🚀
- Uses the recommended nightly of yt-dlp and will check for updates 📼
- Checks for system updates prior to karakeep install/update to avoid any surprises 😯

So I have to admit I am not at all on the MCP train, and I'm a little confused whether we need to build the server, or if it gets built and run when an MCP client initiates communication with karakeep. If it's the latter, then let me know and I'll get rid of the build commands for MCP 🙂

### DIFF
--- a/karakeep-linux.sh
+++ b/karakeep-linux.sh
@@ -85,9 +85,6 @@ install() {
   cd "$INSTALL_DIR"/apps/cli
   pnpm i --frozen-lockfile
   pnpm build
-  cd "$INSTALL_DIR"/apps/mcp
-  pnpm i --frozen-lockfile
-  pnpm build
   cd "$INSTALL_DIR"/packages/db
   pnpm migrate
   echo "Installed Karakeep" && sleep 1
@@ -286,8 +283,6 @@ update() {
     pnpm build
     cd "$INSTALL_DIR"/apps/workers && pnpm i --frozen-lockfile
     cd "$INSTALL_DIR"/apps/cli && pnpm i --frozen-lockfile
-    pnpm build
-    cd "$INSTALL_DIR"/apps/mcp && pnpm i --frozen-lockfile
     pnpm build
     cd "$INSTALL_DIR"/packages/db && pnpm migrate
     echo "$RELEASE" >"$INSTALL_DIR"/version.txt
--- a/karakeep-linux.sh
+++ b/karakeep-linux.sh
@@ -2,7 +2,7 @@
 
 set -Eeuo pipefail
 
-# v2.1
+# v2.2
 # Copyright 2024-2025
 # Author: vhsdream
 # Adapted from: The Karakeep installation script from https://github.com/community-scripts/ProxmoxVE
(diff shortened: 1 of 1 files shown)


MohamedBassem (changes requested):
Hey @vhsdream sorry for the delay in reviewing the PR.
1. I understand that yt-dlp prefers users to use the nightly build, but honestly, I don't see a strong reason why we should move away from the stable build that's tested by everyone?
2. As for MCP, it's not meant to be running on the server, typically MCP clients are spawned by their clients as a subprocess (which usually happens on the client side).
3. Upgrading the distro when updating karakeep would be very unexpected. You're assuming that everyone who uses this script runs karakeep on a dedicated VM, but that's not necessarily the case. I'd say, this should be the responsibility of the caller not our own responsibility honeslt.y

vhsdream:
@MohamedBassem No worries - I know you've got a lot on your plate 😄 

1. For yt-dlp - how about a compromise where we keep the stable repo, but I also keep the added update check in the update function? I guess that was my main concern anyway; the script didn't originally check for an update and I think that strikes a good balance.

2. Ah so for the MCP, do we even need the build command? I thought at least there would need to be something 'server-side' for an MCP client to interface with Karakeep?

3. Noted. Yeah personally I do the LXC container thing, but you're right - not everyone has the ability to do that, and Karakeep could be running in a multi-purpose system. I'll remove those update commands and leave it up to the end-user to maintain that.

MohamedBassem:
@vhsdream 

1. Update check sounds like a good middle ground to me.
2. The MCP server interfaces with karakeep over Karakeep's API, so yeah, nothing server side beside the karakeep server itself.
3. Sounds good.

vhsdream:
All right - I've made the changes, have a look. Also added a service checker to notify if any karakeep services have failed to start after installing or updating.

karakeep-linux.sh
```diff
@@ -85,6 +85,9 @@ install() {
   cd "$INSTALL_DIR"/apps/cli
   pnpm i --frozen-lockfile
   pnpm build
+  cd "$INSTALL_DIR"/apps/mcp
+  pnpm i --frozen-lockfile
+  pnpm build
```
MohamedBassem: Let's stop building the mcp server?

karakeep-linux.sh
```diff
@@ -285,14 +287,16 @@ update() {
     cd "$INSTALL_DIR"/apps/workers && pnpm i --frozen-lockfile
     cd "$INSTALL_DIR"/apps/cli && pnpm i --frozen-lockfile
     pnpm build
+    cd "$INSTALL_DIR"/apps/mcp && pnpm i --frozen-lockfile
+    pnpm build
```
MohamedBassem: Same here.

MohamedBassem (approved):

karakeep-linux.sh
vhsdream: Oh so you meant that it's not at all necessary? I'm so confused to be honest, but give me a few minutes and I'll remove both commands.

MohamedBassem:
Thanks @vhsdream!

Issue #1220, opened 2025-04-08
FR: [bounty $50] support for highlights on android


thiswillbeyourgithub:
### Describe the feature you'd like

Hi!

Karakeep was very promising last time I checked but now is clearly awesome!

I am still missing one crucial feature for my workflow: the ability to create new highlights on android. I am willing to bounty $50 to whoever pulls this off :)

As far as I am concerned, to earn the bounty, this can be either from the android app or the browser, including in "desktop mode" of the android browser. I also can accept minimal highlights like only text highlights, no color selectors, no area highlighter etc. 

Also I was surprised to not find any issue already tracking this feature, I apologize if it exists already but I couldn't find it in issues, PR nor discussions.

### Describe the benefits this would bring to existing Hoarder users

- A way to enjoy all features of karakeep from your phone or tablet
- A real alternative to the good old omnivore.
- A usable setup for users that want to hoard highlights instead of mostly static documents.

### Can the goal of this request already be achieved via other means?

No, I tried adding highlights from the android app, from the brave browser, and from the brave browser with desktop mode enabled.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### Additional context

Sort of related to #620 and #1014

Mxrk:
As I would also like to have this feature, I found a good fix which also supports mobile browsers:

https://github.com/user-attachments/assets/a7d8ecd8-84d4-4652-8bfc-40157fb5a412

I will check if that’s enough and won’t break anything else, but seems working so far. Is this how it should be on mobile or would you prefer something else?

thiswillbeyourgithub:
Hi!

Thanks for the reply. I'm not sure I understood. Did you find a workaround to make the highlighting work from android or did you code it yourself? I do believe it is fair that only the latter qualifies for the bounty. We can debate this of course!

Regarding the video example: it seems alright but on android I don't know if the fact that it highlights the text right away is the way to go, normally there is a contextual menu that appears so we can copy the text, do a web search on it etc. Would we be losing that if we added your fix?

And to clarify, it does work on the android app as well as on the browser, correct?

FYI I have a track record of being honest in my bounties, [see that one for example](https://github.com/thiswillbeyourgithub/SleepTk_pinetime_sleep_tracker/issues/15)

thiswillbeyourgithub:
Hi @Mxrk any update on this?

Mxrk:
> Hi @Mxrk any update on this?

[...]

PR #1348, opened 2025-05-02
feat: position highlight menu based on device type

Mxrk:
Hi, 

this PR adds the highlight functionality to touch devices, mostly mobile devices for the web version.
The main fix as discussed in the issue #1220  is using `onPointerUp={handlePointerUp} `instead of `onMouseUp`. 
With that change, the highlight menu appears on e.g. mobile. 

Besides that, I also adjusted `setMenuPosition` to move the menu below the selected text according to the `isMobile` check as in most cases the device context menu is above. The mobile check was found here and seems to be good for this use case:
https://stackoverflow.com/a/52855084/9950786

Here's how it now looks on mobile:

image

This doesn't adjust the web version:
**Current latest stable after selecting Qualcomm:**

image

**My PR:**

image

### DIFF
--- a/apps/web/components/dashboard/preview/BookmarkHtmlHighlighter.tsx
+++ b/apps/web/components/dashboard/preview/BookmarkHtmlHighlighter.tsx
@@ -19,6 +19,7 @@ interface ColorPickerMenuProps {
   onDelete?: () => void;
   selectedHighlight: Highlight | null;
   onClose: () => void;
+  isMobile: boolean;
 }
 
 const ColorPickerMenu: React.FC<ColorPickerMenuProps> = ({
@@ -27,6 +28,7 @@ const ColorPickerMenu: React.FC<ColorPickerMenuProps> = ({
   onDelete,
   selectedHighlight,
   onClose,
+  isMobile,
 }) => {
   return (
     <Popover
@@ -44,7 +46,10 @@ const ColorPickerMenu: React.FC<ColorPickerMenuProps> = ({
           top: position?.y,
         }}
       />
-      <PopoverContent side="top" className="flex w-fit items-center gap-1 p-2">
+      <PopoverContent
+        side={isMobile ? "bottom" : "top"}
+        className="flex w-fit items-center gap-1 p-2"
+      >
         {SUPPORTED_HIGHLIGHT_COLORS.map((color) => (
           <Button
             size="none"
@@ -113,6 +118,9 @@ function BookmarkHTMLHighlighter({
   const [selectedHighlight, setSelectedHighlight] = useState<Highlight | null>(
     null,
   );
+  const isMobile = useState(
+    () => typeof window !== 'undefined' && window.matchMedia("(pointer: coarse)").matches
+  )[0];
 
   // Apply existing highlights when component mounts or highlights change
   useEffect(() => {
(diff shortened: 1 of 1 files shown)


thiswillbeyourgithub:
Hi!

After tremendous pain I was able to build the app! Unfortunately it does not seem to be usable.

After trying many different things with my thumbs here's what I gathered:
- if I long press with my thumb on some text, it gets selected and I get the android context popup. So far no highlighting ability.
- if while the text is selected I move my thumb randomly/swipe or something I am able to see the color popup, for a fleeting microsecond then it disappears. I cannot reproduce it at will and it seems glitchy.

I am on android using brave browser.

Any idea of what's going on?

Mxrk:
I didn’t test the Android + Brave combination but it seems like brave on Android is (maybe) directly triggering a scroll event which will cancel the Pointer event. 

Could switch to touch events on mobile only, that could help.

thiswillbeyourgithub:
Thanks!

I'm seeing progress, yet too soon to share. I'll try to report back in the coming days (and have not forgotten about the bounty).

Mxrk:
I installed Brave on an android device and sadly can't reproduce your problem:

https://github.com/user-attachments/assets/2a79a36d-f0c9-427f-9ebd-48f1bbbd7a43

thiswillbeyourgithub:
(I'm going quite far on my own with the docker building so unless you're doing it for yourself I suggest not wasting your time on duplicating efforts on this until I push a clean code to suggest for review.)

thiswillbeyourgithub:
Well for some reason it seems to be working **flawlessly** now. I have no idea what the issue was. Thanks a lot!

> would you accept giving me your email to start discussing payment? Or a link to a temporary pastebin with your mail? Or I can create a private repo, invite you there so we can exchange privately. Your call!

thiswillbeyourgithub:
Progress: i now have the previous glitchy behavior. I suspect that its a combination of my specific device + having powersaving enabled (pretty sure it changes the digitizer sensitivity) + maybe brave plays a role. The resulting behavior is that if I "touch-click" on the selected blue text, it makes the color picker appear and disappear right away. But if I do even the most minuscule swiping motion on it, then the picker appears normally. Same for modifying a highlight, swiping works always, touching it depends.

All in all I am satisfied for the criteria I put in the bounty and @Mxrk will get it (we'll do the exchange in private, but I would appreciate it if they could publicly acknowledge here when they receive the money as this helps my reputation and hence my motivating power :))

MohamedBassem:
Thanks a lot @Mxrk!

Mxrk:
Thanks to @thiswillbeyourgithub for the bounty and I confirm that everything went flawlessly.

thiswillbeyourgithub:
Hey @Mxrk , I noticed that although it works fine on the browser on android, the pop up does not appear at all if the preview is not a modal (I believe that's the word).

Basically when reading a preview if I reload the page (or just multitask and my browser reloads the stale page) the preview appears as a page instead of a popup on top of the dashboard. In that configuration, the highlight pop up never appears.

Do you think it can be trivially fixed or should I create a new issue to track this?

Thanks!

Mxrk:
> Hey @Mxrk , I noticed that although it works fine on the browser on android, the pop up does not appear at all if the preview is not a modal (I believe that's the word).
> 
> 
> 
> Basically when reading a preview if I reload the page (or just multitask and my browser reloads the stale page) the preview appears as a page instead of a popup on top of the dashboard. In that configuration, the highlight pop up never appears.
> 
> 
> 
> Do you think it can be trivially fixed or should I create a new issue to track this?
> 
> 
> 
> Thanks!

Create a new issue, probably good with some kind of recording to better understand the problem! I feel like it would need a fix in the general highlighting engine.

thiswillbeyourgithub:
Thanks!


Edit: sadly I can't reproduce on other setups than my android brave browser on my phone. Will investigate some more

PR #1350, opened 2025-05-02
feat: Add Helm chart

sheyabernstein:
Adds a Helm chart for Kubernetes.

### DIFF
--- a/charts/karakeep/README.md
+++ b/charts/karakeep/README.md
@@ -1,0 +1,38 @@
+# Karakeep Helm chart
+
+Helm chart for deploying Karakeep along with:
+
+- **[Meilisearch](https://github.com/meilisearch/meilisearch-kubernetes)**: for fast and lightweight full-text search
+- **[Headless Chrome](https://github.com/jlandure/alpine-chrome)**: enables web page previews
+
+This chart inherits from the [bjw-s/common](https://github.com/bjw-s/helm-charts/tree/main/charts/library/common) library.
+
+### Configuration
+
+| Key                    | Description                              | Default                  |
+| ---------------------- | ---------------------------------------- | ------------------------ |
+| `applicationHost`      | Hostname used in ingress/service         | `karakeep.domain`        |
+| `applicationProtocol`  | Protocol for internal service references | `http`                   |
+| `applicationSecretKey` | Secret used for app authentication       | Auto-generated if `null` |
+| `meilisearchMasterKey` | Meilesearch master key                   | Auto-generated if `null` |
+
+#### Example with OIDC Authentication
+
+```yaml
+controllers:
+  karakeep:
+    containers:
+      karakeep:
+        env:
+          DISABLE_PASSWORD_AUTH: "true"
+          OAUTH_ALLOW_DANGEROUS_EMAIL_ACCOUNT_LINKING: "true"
+          OAUTH_PROVIDER_NAME: OIDC
+          OAUTH_SCOPE: openid email profile
+          OAUTH_WELLKNOWN_URL: https://auth.company/application/o/karakeep/.well-known/openid-configuration
+
+secrets:
+  karakeep:
+    stringData:
+      OAUTH_CLIENT_ID: your-client-id
+      OAUTH_CLIENT_SECRET: your-client-secret
+```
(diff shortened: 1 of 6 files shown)


timstott:
Thank you for your contribution. I am looking forward to have an official chart. 

Would you be open to remove the dependency on the bjw-s library to ease maintenance? I am happy to help

sheyabernstein:
Hey @timstott, I used the bjw-s common library as a base to avoid reinventing the wheel — it provides a lot of flexible patterns and handles much of the boilerplate, which makes it easier to maintain and extend. It's also well maintained and widely used in the self-hosting community. That said, I'm happy to discuss what a minimal, standalone chart would look like if you’d prefer to keep dependencies to a minimum. Appreciate your openness to help!

charts/karakeep/values.yaml
```diff
@@ -0,0 +1,145 @@
...
+      chrome:
+        image:
+          repository: gcr.io/zenika-hub/alpine-chrome
+          tag: 124
+
+        args:
+          - --remote-debugging-address=0.0.0.0
+          - --remote-debugging-port=9222
+          - --headless
+          - --disable-gpu
```
MohamedBassem, line 72: Should we add the missing `--disable-dev-shm-usage` and `--hide-scrollbars`?

charts/karakeep/README.md
```diff
@@ -0,0 +1,38 @@
+# Karakeep Helm chart
```
MohamedBassem, line 1: Can we add some instructions on how to install the chart and how to upgrade it?

MohamedBassem (approved):

MohamedBassem:
Thanks a lot for the PR! I'd personally also prefer if we don't have the `bjw-s` dependency, but at the same time, I'm not very familiar with helm and not very familiar with how much boilerplate it saves.

charts/karakeep/README.md
sheyabernstein: This bit is a little tricky, it requires repo permissions to set up Github Pages and CI.

Docs: https://github.com/helm/chart-releaser-action

Once CI and Pages are configured, we can add the Helm repository URL in the README with instructions. For now, you can checkout this branch, navigate to the chart directory, and run the command below

```shell
helm upgrade -i --create-namespace -n karakeep karakeep .
```

charts/karakeep/values.yaml
sheyabernstein: Added in f168984

charts/karakeep/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+Helm chart for deploying Karakeep along with:
```
timstott, line 3: What are your thoughts on moving the helm chart README to _charts/karakeep/README.md_ where it is traditionally kept? We can then reference it in the root README.

As an example the [`meilisearch` chart README](https://github.com/meilisearch/meilisearch-kubernetes/blob/main/charts/meilisearch/README.md)

charts/karakeep/values.yaml
```diff
@@ -0,0 +1,156 @@
...
+        port: 3000
+  chrome:
+    controller: chrome
+    ports:
+      http:
+        port: 9222
+
+secrets:
+  karakeep:
+    enabled: true
```
timstott, line 121: Do I understand correctly that setting the parameter to `false` enables users to declare their own `Secret` as long as it has the name `"{{ .Release.Name }}"`?.

This is a sought after feature when using gitops to checkin secrets into source control using [sealed-secrets](https://github.com/bitnami-labs/sealed-secrets) for example.

sheyabernstein: Correct, that will prevent the chart from creating secrets.

charts/karakeep/README.md
sheyabernstein: The meili repo is a dedicated chart repo while karakeep would have the chart in the main repo under a charts subdir. Ultimately I'll defer to the maintainers on this.

MohamedBassem:
I'm sorry this took that long to be merged :) Thanks a lot for the PR.

sheyabernstein:
Thanks for merging @MohamedBassem :)

Can Github Pages be set up with actions to build and release the chart for new versions?

https://helm.sh/docs/howto/chart_releaser_action/

MohamedBassem:
The charts are now released in a separate repo here: https://karakeep-app.github.io/helm-charts/

PR #1351, opened 2025-05-03
fix(tags): normalise leading hashes in tag names (#1317)

### LINKED ISSUE #1317: Prevent the creation of '##test' hashtag
### Describe the Bug

Today I ran into the following issue:

### Steps to Reproduce

I created a new manual tag `#glt25` internal Karakeep created a new tag `##glt25`. Next, I created a custom list and wondered why the search result was empty. The search query was `#glt25`

### Expected Behaviour

Maybe it’s a good idea to prevent the creation of a double hashtag at the beginning of a phrase.

Best
Michael

Image

### Device Details

Safari

### Exact Karakeep Version

latest

Mostafa-Wahied:
## Summary
Fixes #1317 — typing `#tag`, `##tag`, or even `###tag` could save a double‑hash tag to the database, breaking search.
We now trim whitespace and strip every leading `#` on every tag‑creation path:
- standalone Tags screen (`tags.create`, `tags.update`)
- bookmark editor (`bookmarks.updateTags`)

UI continues to display #tag; only storage is normalized.

## Key changes
`packages/trpc/utils/tag.ts`: new `normalizeTagName` helper
`packages/trpc/routers/tags.ts`: call helper in create and conditional update
`packages/trpc/routers/bookmarks.ts`: apply helper (bulk) + ignore empty names to avoid `values() must be called with at least one value` error
`packages/trpc/routers/tags.test.ts`: +2 tests (create & update) confirming hashes are stripped

## Tests
`pnpx vitest run packages/trpc`
- All existing suites pass.
- New cases:
  - `##demo` → stored `demo` (create)
  - rename `#foo`→`##bar` → stored `bar` (update)


MohamedBassem:
Thanks for the PR @Mostafa-Wahied and welcome to the contributors list!

PR #1362, opened 2025-05-05
Feature: add support for archiving and previewing WARC web session recordings

yapishu:
I spent this weekend hacking together a feature I wanted in karakeep. [WARC](https://archive-it.org/post/the-stack-warc-file/) format is used for creating archival copies of web pages; it is for instance what archive.org uses for the Wayback Machine. Rather than just making copies of static assets, it records sessions which can be played back in a viewer. If I am to be preserving bookmarks in a self-hosted manner, WARC files are exactly what I want, so I was kind of surprised at how little existing tooling there was for capturing and viewing them. I think Karakeep is awesome and I wanted to add this as a feature.

### Demo

[demo video](https://youtu.be/MQUMIWbJLo4)

### Architecture

The heart of this is a new sidecar container called [`warcer`](https://github.com/yapishu/karakeep/blob/reid/warc/docker/warcer/main.go), which is a pretty simple little custom go application that makes use of the existing headless chrome sidecar to crawl, similarly to how the existing crawler works, but it records the contents of the session to a single file which can be played back.

I added a new worker called `warcerWorker`, which basically watches the crawl queue for completed crawls and then looks up the URLs for them, then POSTs them to the warcer sidecar, which in turn starts piloting chrome. The warcer finishes, writes the file, then returns the path to the caller, which feeds it into the asset preprocessor. 

I extended the bookmark json schema to include a new `linkWarcArchiveAssetId` key, which will enable the new 'WARC View' preview option:

image

The Warc View embeds the [ReplayWeb.Page](https://github.com/webrecorder/replayweb.page) service worker, which renders a nice little fake web browser to play back the session:

image

### Other considerations

- This PR ended up being really big, sorry
- The current state of the warcer sidecar is pretty primitive -- it works, and you can even feed it multiple URLs that you can later browse in the reader, but e.g I'm not positive that it is actually like ISO-accordant or anything, it was pretty PoC weekend-project-grade
- I tried to adhere to the previously existing conventions but I'm sure there's stuff I didn't do correctly. I'm not a frontend guy so I probably did something stupid in there
- The warc preview doesn't work without HTTPS or screwing with your CORS
- The headless chrome tagged in the compose in master (123) has a bad entrypoint, bumping it to 124 fixes it. This is not specific to this PR, it's straight up broken at the moment


dehlen:
Just wanted to leave my +1 here. Warc support actually would elevate Karakeep to a real archivebox alternative for me. I hope this gets merged!

MohamedBassem:
Hey, I'm sorry it took me that long to get to this PR. You seem to have put a lot of effort into it (including recording the demo video!). However, at the same time, this PR is introducing a ton of new functionality in Karakeep that I'd have preferred if we discussed before you spending that amount of time on the PR. The [contribution guidelines](https://github.com/karakeep-app/karakeep/blob/main/CONTRIBUTING.md) suggests opening an issue first and make sure you're aligned with the maintainer to avoid exactly this scenario.

Regarding the functionality itself, it's honestly still not super clear to me why WARC is better than the singleFile/monolith archive? Even if it is, for something like archivebox it makes sense because it's its core functionality, but for Karakeep, I'm afraid that this introduces too much complexity for something very niche from karakeep's perspective. Specially with the introduction of a completely new sidecar and a completely new language to the project.

I unfortunately don't think I'll be able to accept this PR. It makes me feel terrible given the amount of effort you seem to have put into this, but please let's make sure we're aligned on an issue before introducing such a major change.

PR #1369, opened 2025-05-07
Add icon to show the archive status of a bookmark

### LINKED ISSUE #1368: Show the archive status of a bookmark

I want to add an archive icon to a bookmark when it is archived, similar to the star icon.

When using search or searching by tags (going to the tags page and clicking on a tag), the results don't show if the bookmark is archived or not. A star icon is used to show if the bookmark is starred. So adding an archive icon will help to know if the search/tag search results contain a bookmark which is archived. 

No

### Additional context

_No response_

PunitLodha:
Adds an archive icon to the bookmark card to show the archive status of a bookmark

Closes #1368 

Before:
Screenshot_20250507_173348

After:
Screenshot_20250507_173410

### DIFF
--- a/apps/web/components/dashboard/bookmarks/BookmarkActionBar.tsx
+++ b/apps/web/components/dashboard/bookmarks/BookmarkActionBar.tsx
@@ -1,7 +1,7 @@
 import Link from "next/link";
 import { buttonVariants } from "@/components/ui/button";
 import { cn } from "@/lib/utils";
-import { Maximize2 } from "lucide-react";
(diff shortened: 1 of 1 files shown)


MohamedBassem:
First of all, thanks for the PR! However, I'm not really a big fan of showing the archive status the way it is implemented here as for people who have auto archival enabled it'll introduce a lot of clutter. I had another PR (#743) which moves those icons to be on hover, but I haven't had a chance to work on it again since.

I appreciate you taking the time to send the PR though!

PR #1374, opened 2025-05-09
feat: Add NETSCAPE-Bookmark-file-1 export format support

yuikisaito:
## Description

This Pull Request supports exporting bookmarks in NETSCAPE-Bookmark-file-1 format.This feature will allow users to import bookmarks stored in Karakeep into most browsers.

## Changes

1. Added a new `toNetscapeFormat` function in `exportBookmarks.ts` to convert bookmarks to NETSCAPE format
2. Updated the export endpoint to support format selection via query parameter
3. Enhanced the UI to allow users to choose between JSON and NETSCAPE formats

## Testing

- Manually tested export in both formats
- Verified the exported NETSCAPE format can be imported back into Karakeep
- Confirmed the exported file can be imported into major browsers (Chrome, Firefox)
- Ensured special characters in URLs, titles, and tags are properly encoded

## Screenshots

image

### DIFF
--- a/apps/web/app/api/bookmarks/export/route.tsx
+++ b/apps/web/app/api/bookmarks/export/route.tsx
@@ -1,3 +1,4 @@
+import { NextRequest } from "next/server";
 import {
   toExportFormat,
   toNetscapeFormat,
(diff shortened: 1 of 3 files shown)


MohamedBassem (approved):
Thank you!

PR #1392, opened 2025-05-11
fix(search): add new relevance sort order

### LINKED ISSUE #1335: ‘Newest First’ sorting does not work when search is active
### Describe the Bug

Image

### Steps to Reproduce

- apply search for some tags
- apply sort: "newest first"

### Expected Behaviour

newer logs come first

_No response_

### Device Details

_No response_

### Exact Karakeep Version

0.24.1

xuatz:
## Summary

* add new sort order `relevance`
* retain existing(legacy) search's sort logic as `relevance`
* add new `asc` and `desc` sort logic

fixes #1335

## Notes

I'm wasn't super sure how karakeep handles the translation, so let me know if there are follow up tasks, and anything else that I should tweak for this PR, thanks!

### DIFF
--- a/apps/web/app/dashboard/search/page.tsx
+++ b/apps/web/app/dashboard/search/page.tsx
@@ -1,13 +1,21 @@
 "use client";
 
-import { Suspense } from "react";
+import { Suspense, useEffect } from "react";
 import BookmarksGrid from "@/components/dashboard/bookmarks/BookmarksGrid";
 import { FullPageSpinner } from "@/components/ui/full-page-spinner";
 import { useBookmarkSearch } from "@/lib/hooks/bookmark-search";
+import { useSortOrderStore } from "@/lib/store/useSortOrderStore";
 
 function SearchComp() {
   const { data, hasNextPage, fetchNextPage, isFetchingNextPage } =
     useBookmarkSearch();
+
+  const { setSortOrder } = useSortOrderStore();
+
+  useEffect(() => {
+    // also see related cleanup code in SortOrderToggle.tsx
+    setSortOrder("relevance");
+  }, []);
 
   return (
     <div className="flex flex-col gap-3">
--- a/apps/web/components/dashboard/SortOrderToggle.tsx
+++ b/apps/web/components/dashboard/SortOrderToggle.tsx
@@ -1,3 +1,7 @@
+"use client";
+
+import { useEffect } from "react";
+import { usePathname } from "next/navigation";
 import { ButtonWithTooltip } from "@/components/ui/button";
 import {
   DropdownMenu,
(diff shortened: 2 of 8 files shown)


MohamedBassem (changes requested):
Hey, that was fast @xuatz!

Overall, the PR looks good to me, however there's a couple of changes I think we need to do:

1. Relevance doesn't make much sense when we're not in a the "search" page, can we hide the relevance option when we're not in the search page?
2. Similarly, I think we should change the `getBookmarks` endpoint to not accept `relevance` as one of its sort orders.

thiswillbeyourgithub:
Hi,

Can I ask why we can't choose the sort order when [querying from the API](https://docs.karakeep.app/API/search-bookmarks)? I can imagine uses for it.

MohamedBassem:
@thiswillbeyourgithub it's an oversight, if you send a PR, I'll happily accept it :)

thiswillbeyourgithub:
Thanks for the quick reply. Unfortunately I'm useless at typescript (although Karakeep is a good reason for me to focus more on it!) so I just want to explicitely state that I won't make a PR (i.e. don't wait for me).
If @xuatz doesn't either then I'll create a separate issue to track this feature request.

xuatz:
> Thanks for the quick reply. Unfortunately I'm useless at typescript (although Karakeep is a good reason for me to focus more on it!) so I just want to explicitely state that I won't make a PR (i.e. don't wait for me).
> If @xuatz doesn't either then I'll create a separate issue to track this feature request.

If it's okay to slightly bloat the PR scope I can include it here!

xuatz:
> > Thanks for the quick reply. Unfortunately I'm useless at typescript (although Karakeep is a good reason for me to focus more on it!) so I just want to explicitely state that I won't make a PR (i.e. don't wait for me).
> > If @xuatz doesn't either then I'll create a separate issue to track this feature request.
> 
> If it's okay to slightly bloat the PR scope I can include it here!

@thiswillbeyourgithub actually, i change my mind. it involves updating the openapi docs at which I'm not opposed to, but i'm not familiar with. i'll do it in a separate PR.

xuatz:
> Hey, that was fast @xuatz!
> 
> Overall, the PR looks good to me, however there's a couple of changes I think we need to do:
> 
> 1. Relevance doesn't make much sense when we're not in a the "search" page, can we hide the relevance option when we're not in the search page?
> 2. Similarly, I think we should change the `getBookmarks` endpoint to not accept `relevance` as one of its sort orders.

@MohamedBassem 

for point number 2, I initially went for the approach to make 2 zSortOrders

```ts
const zSearchBookmarksSortOrder = z.enum(["asc", "desc", "relevance"]);
const zGetBookmarksSortOrder = z.enum(["asc", "desc"]);

export type ZSortOrder = z.infer<typeof zSearchBookmarksSortOrder>;
```

But because they both share the same `useSortOrderStore()`, `getBookmarks()` is gonna complain that we are potentially giving it `relevance`.

I guess we can make 2 stores, one specifically for `/search` and the existing one for the rest, but I got a feeling we probably don't want that.

Instead I've added a cleanup useEffect in `SortOrderToggle.tsx`, to reset the sort order, if the page is not /search and if the sortOrder is still "relevance"

```tsx
  // also see related on page enter sortOrder.relevance init
  // in apps/web/app/dashboard/search/page.tsx
  useEffect(() => {
    if (!isDashboardSearchPage && currentSort === "relevance") {
      // reset to default sort order
      setSortOrder("desc");
    }
  }, [isDashboardSearchPage, currentSort]);
```

Hard to say which approach you prefer, or if you have a 3rd idea, let me know 👍

PR #1422, opened 2025-05-17
minor: allow specifying source location when building via docker

thiswillbeyourgithub:
- **new: allow to specify the source directory when building**
- **update the docker compose files to use SOURCE_DIR**

Hi,

**This is low importance, maybe a mistake, if the reviewer is low on bandwidth please consider skipping this one and reviewing other PRs instead, thanks!**

I'm fairly new to docker and had some trouble getting the building process to work after modifying the sources.

My issue was mainly that because the sources are mounted to /app, and the building process outputs to /app, I could build the container but to run it I'd have to comment the runtime volume mounting at /app (because it happens after docker loads the built output).

I ended up figuring out how to modify the Dockerfile files to make them accept a source dir.

This has been working well for me but I'm not sure if it can break things for others.

For reference, here's how my own docker-compose.yml file is setup that I run with `docker compose up web --build:

```yml
services:
  web:
    restart: unless-stopped
    volumes:
      - ./data/web:/data
      - ./thefork:/appsource

    # To either run from image or build from scratch, uncomment one of the two sections

    # # START OF FROM IMAGE
    # image: ghcr.io/karakeep-app/karakeep:${KARAKEEP_VERSION:-release}
    # # END OF FROM IMAGE

    # # To build the app:
    # 0. comment the image: above
    # 1. uncomment the building lines below
    # # START OF BUILDING
    build:
      context: ./thefork
      target: aio
      dockerfile: docker/Dockerfile
      args:
        SOURCE_DIR: /appsource
    working_dir: /app
    # # END OF BUILDING

    # ...rest of the config...
```

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,7 +1,8 @@
 ################# Base Builder ##############
 FROM node:22-alpine AS base
 
-WORKDIR /app
+ARG SOURCE_DIR=/app  # Default value if not specified
+WORKDIR ${SOURCE_DIR}
 ENV PNPM_HOME="/pnpm"
 ENV PATH="$PNPM_HOME:$PATH"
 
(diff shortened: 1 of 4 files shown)


MohamedBassem:
You don't really need to modify the source dir location. E2E have an example here (https://github.com/karakeep-app/karakeep/blob/main/packages/e2e_tests/docker-compose.yml#L3) on how to do a docker compose with local image build.

thiswillbeyourgithub:
This isn't the issue: building will always work but currently the`/app` dir is both the location of the source code AND where the built app is being called. So currently we can't navigate between the sources and the compiled code inside the container (the built app is "mounted upon" the source dir).

I don't recall exactly why I needed this but it made it easier to debug stuff on my end.

I you still don't want this, I won't insist more to save you time though.

PR #1424, opened 2025-05-17
doc: mention karakeep-python-api in community projects

thiswillbeyourgithub:
- **doc: mention karakeep-python-api in the community projects**
- **doc: also mention karakeep-python-api in the cli section**


MohamedBassem:
Thank you and looking forward to what people will build with the python API!

Issue #1285, opened 2025-04-20
[BUG] Screen real estate utilization on android browser


thiswillbeyourgithub:
### Describe the Bug

Hi,

The current rendering of article on android browser is unfortunately not really usable on my phone, allow me to illustrate:

![Image](https://github.com/user-attachments/assets/c4af8f8d-a38f-4e70-bda6-ba130d85515d)

The readable content is only the red rectangle towards the middle, here are the issues I have with the current UI:
- On both sides, the cyan rectangles illustrate wasted margins
- When scrolling the text content, the browser thinks we are scrolling an element without scrolling the page, hence the bottom section with tabs and the top section with the url remain in sight instead of going away (see bottom and top cyan rectangles)
- Similarly, the green rectangle is shown all the times, squeezing the reading area permanently.

My suggested solution:
- remove the lateral margins
- make the text scrolling be interpreted by the browser in a way to makes the top and bottom bar go away
- hide the green section most of the time. I believe the best UI would be to have a "tab" view at the top with section "Article" and "Notes". Or a [kebab UI manu](https://external-content.duckduckgo.com/iu/?u=https%3A%2F%2Fi.pinimg.com%2Foriginals%2F15%2F82%2F8f%2F15828fbf117e5684ce25fb737acdb751.png&f=1&nofb=1&ipt=e753cf2997ce76007b3f86c3562359f313d57c0b97220e797daa5a664b5dc639) at the top right corner that upon clicking shows a "Notes" button that would make the current green screen appear as a split (like currently). This way we still keep the ability to add notes while reading the text but don't have to have the green screen at all times.

This is on brave browser btw.

### Steps to Reproduce

Open any link on brave browser on android.

### Expected Behaviour

Most of the UI should be dedicated to reading by default, the notes and metadata should not be permanently displayed.

### Exact Karakeep Version

v0.23.2

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

thiswillbeyourgithub:
Progress: it seems that if I turn on "desktop mode" then turn it off, the page that is loaded is much more sane: the whole AI section is now at the very bottom after the entirety of the article. This is tested on android brave browser.

This is still not satisfactory because it makes it impossible to access those things *without* losing where you are in the article.

MohamedBassem:
Fixed by @thiswillbeyourgithub in [#1425](https://github.com/karakeep-app/karakeep/pull/1425)

Issue #1423, opened 2025-05-17
FR: Keeping track of reading progress of bookmarks


thiswillbeyourgithub:
### Describe the feature you'd like

Hi,

Surprised to not have seen this requested already.

Basically it would be nice if karakeep could:
1. Remember the approximate scrolling of your last time you opened that bookmark (and work across platforms) and auto apply it upon reopening the bookmark.
2. Make that a percentage value, accessible as a metadata, so accessible to the rule engine (see #1406 ) as well as the API (and not read only)
3. Make it so that when the percentage gets to 100%, it just remembers the last scroll place instead of always going to the bottom. For example a long document that you read entirely but where you sometimes reread section.

For having seen issues with the how omnivore did it back in the days, I think it can be non trivial to get perfect, especially the multiplatform aspect, but I don't think it's particularly hard to get *very usable* (a 10% approximation on the reading scroll is not that bad, although painful for very long documents, it's still better than nothing).

### Describe the benefits this would bring to existing Karakeep users

Better reading experience

### Can the goal of this request already be achieved via other means?

Don't think so.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

esimkowitz:
@MohamedBassem I am interested in tackling this, can I be assigned this?

MohamedBassem:
@esimkowitz it seems like you're going to be working on a bunch of issues, don't hesitate to join the #developement channel on discord :)

esimkowitz:
Thanks, will do! Yeah I've been using Karakeep for a few months now and I love it. As a former Pocket user, Karakeep is so close to being a perfect replacement, I hope these Quality of Life enhancements help bridge that parity gap even more!

esimkowitz:
I've got a PR out to address the primary concern of this feature request (synchronized reading progress).

I chose not to address the metadata for the rules engine so as not to bloat the PR. @thiswillbeyourgithub If you still care about this part, can you break it out as a separate issue?

As for the last suggestion, this felt too niche. It feels like this can be solved currently by saving highlights of the areas you want to reread.

thiswillbeyourgithub:
Hi @esimkowitz 

I'm so sorry for not replying earlier. It appears I closed that tab the other day by mistake so forgot to reply.

1. Thanks a lot for the PR!
[...]

PR #1425, opened 2025-05-17
web(UX): new: swipeable tabs when reading on touch devices

thiswillbeyourgithub:
- **feat: Implement tabbed layout for bookmark preview content and details**
- **feat: Enable swipe gesture to switch between content and details tabs**
- **feat: add scroll-based tab bar hide/show functionality**
- **fix: adjust tab bar layout to maintain content space when hidden**
- **fix: bg-muted was stopping midway in the background**
- **fix: disable mouse tracking for swipe detection in BookmarkPreview**
- **fix: improve scroll behavior to prevent glitchy tab bar animation**

Hi,

I knew nothing about react nor typescript just yesterday, and now barely have a grasp, so if you are opposed to LLM based contributions, don't waste more time and close this :).

I made this to solve #1285, or at least do some progress.

image
image

Otherwise, this PR is a change to the "reading a bookmark" page. It moves the bookmark details to a separate tab, and needs the dependency react-swipeable to allow switching from reading to details and back via a horizontal swipe.

I also had to remove one `bg-muted` color that was mentioned in a seemingly unrelated file (sidebar/SidebarLayout.tsx) but was glitching the UI (the color was stopping midway behind the article).

I have been testing this both using `pnpm web` and after building the docker container entirely and it held up remarkably well. I also used the chrome dev tools to simulate various other touch and non touch devices and couldn't spot any issues.

This was done using [aider-chat](https://github.com/Aider-AI/aider).

I am a python dev originally, and although I aspire to figure out react and typescript, I have very limited knowledge of the implications of adding the `react-swipeable` library in particular.

Open to any modifications about this, especially if it makes me a better dev!


apps/web/components/dashboard/preview/BookmarkPreview.tsx
```diff
@@ -1,11 +1,13 @@
 "use client";
 
-import React from "react";
+import React, { useState, useRef, useEffect, useCallback } from "react";
```
xuatz:
```suggestion
import { useState, useRef } from "react";
```

apps/web/components/dashboard/preview/BookmarkPreview.tsx
```diff
@@ -68,6 +70,27 @@ export default function BookmarkPreview({
   initialData?: ZBookmark;
 }) {
   const { t } = useTranslation();
+  const [activeTab, setActiveTab] = useState<string>("content");
+  const [showTabBar, setShowTabBar] = useState<boolean>(true);
```
xuatz:
```suggestion
  const [showTabBar, setShowTabBar] = useState(true);
```
`useState` can infer that its boolean from `true`
same applies for the other similar cases as well

apps/web/components/dashboard/preview/BookmarkPreview.tsx
xuatz: not necessary to default import `React`, + remove unused imports.

xuatz (commented):
I can say that at least the feature works, pretty well ^^

MohamedBassem:
Thanks a lot @thiswillbeyourgithub for going out of your and implementing a feature that you want in a stack that you're not familiar with.

I like how it looks on small screens, but not a big fan of the tabs idea on large screens. Can we keep the current layout on desktops and only switch to tabs in smaller screens?

thiswillbeyourgithub:
Thanks!

Your suggestion is fine by me but please reconsider for a moment: I agree that the style can be criticized, but there is a practical aspect that I want to make sure I explicitly point out. If the article is long, even very long, then to get to see the details of the bookmark and to get to see a list of highlights, you have to scroll all the way down. This does not feel practical and even annoying. Furthermore, I don't think it would integrate well with feature #1423 that I asked.

Or maybe I am missing something?

MohamedBassem:
@thiswillbeyourgithub On large screens the article pane scrolls independently from the bookmark details. So you can see the bookmark details side by side with the article as such:

CleanShot 2025-05-18 at 4  30 58@2x

So this tabbed layout is mostly useful in mobile screens.

thiswillbeyourgithub:
Okay thanks for clarifying. I do have non touch vertical screens, so your layout would not be appropriate (i.e. I can't just use your layout for touch device, I need to take into account screen ratios).

I will do that but it's going to take me a few days as I'll be busy, unless someone has explicit recommendation for the best way to determine the layout to use, which I'm all ears.

thiswillbeyourgithub:
Well actually aider + sonnet 3.7 is just too good so it seems to work already. @xuatz would you have the time to give it a go?

MohamedBassem:
@thiswillbeyourgithub I hope you don't mind, I did a bunch of changes to make the code simpler.

1. Dropped the fancy stickiness of the tab bar. Not worth the scroll handler (I was hitting some bugs)
2. Drop the swiping functionality, less fancy but less dependencies.
3. Used media queries instead of javascript to detect screen width. This eliminates the UI shifts that happens on refresh.

Overall, the amount of changes left is reasonable, and I'd be happy to merge it.

thiswillbeyourgithub:
I certainly don't mind and am grateful for the time you take.

I am taking a look. I totally understand getting rid of the swipe, ~~but the lack of stickiness is a problem. It means I will always have to go to the top of the article if I want to re read the highlights for example.~~ You seem to have made that bar permanent so it's fine now.

I do notice a UI glitch: the black area underneath the tab bar:
image

Also, is there a reason the left sidebar is not collapsible? Intuitively I felt that it would be better if it were.

thiswillbeyourgithub:
Addendum: actually on my phone the tab bar disappears when I scroll down but not on my vertical desktop. I feel it's a problem if we can't reach to the details easily and without interrupting reading (actually that's probably why you even put the details page here in the first place).

MohamedBassem:
@thiswillbeyourgithub The bar seems to be sticky when it's a "modal" (e.g. opening a page from the grid), and not sticky when you open the bookmark preview page directly (when it's no longer a modal). Let me see if there's something I can do about this, but honestly, won't have much time debugging so will probably merge the PR if I can't find a quick fix for it.

MohamedBassem:
ah, the stickiness is a bigger problem than this PR. It seems that in the non modal-mode, even the details pane gets scrolled away. Will try to fix that independent from this PR.

MohamedBassem:
Thanks a lot @thiswillbeyourgithub, I agree that the final outcome is much better than the original state. Thanks again for taking the time and experimenting in a stack you're not comfortable with!

PR #1427, opened 2025-05-17
fix(api): make PUT bookmark to a list idempotent

### LINKED ISSUE #1402: Add a bookmark to a list API should be idempotent
Adding bookmark to the list where it is already present returns 400 error, which prevents (or at least makes it more complicated) for exporters to work.

xuatz:
Fixes #1402

### DIFF
--- a/packages/open-api/lib/lists.ts
+++ b/packages/open-api/lib/lists.ts
@@ -232,14 +232,6 @@ registry.registerPath({
     204: {
       description: "No content - the bookmark was added",
     },
-    400: {
-      description: "Bookmark already in list",
-      content: {
-        "application/json": {
-          schema: ErrorSchema,
-        },
-      },
-    },
     404: {
       description: "List or bookmark not found",
       content: {
--- a/packages/trpc/models/lists.ts
+++ b/packages/trpc/models/lists.ts
@@ -260,11 +260,10 @@ export class ManualList extends List {
     } catch (e) {
       if (e instanceof SqliteError) {
         if (e.code == "SQLITE_CONSTRAINT_PRIMARYKEY") {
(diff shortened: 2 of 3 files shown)


packages/trpc/models/lists.ts
```diff
@@ -260,11 +260,10 @@ export class ManualList extends List {
...
       if (e instanceof SqliteError) {
         if (e.code == "SQLITE_CONSTRAINT_PRIMARYKEY") {
-          throw new TRPCError({
-            code: "BAD_REQUEST",
-            message: `Bookmark ${bookmarkId} is already in the list ${this.list.id}`,
-          });
+          // this is fine, it just means the bookmark is already in the list
+          return;
         }
+        console.error("Unhandled SqliteError");
```
MohamedBassem: Can we remove this console log?

MohamedBassem (approved):
Thank you! 

Can you also update the openapi specs in `packages/open-api/lib/bookmarks.ts` and then run `pnpm generate` there to update the openapi specs?

packages/trpc/models/lists.ts
xuatz: sure 👍

packages/open-api/lib/lists.ts
```diff
@@ -232,14 +232,6 @@ registry.registerPath({
...
       description: "No content - the bookmark was added",
     },
-    400: {
-      description: "Bookmark already in list",
-      content: {
-        "application/json": {
-          schema: ErrorSchema,
-        },
-      },
-    },
```
xuatz, line 242: @MohamedBassem I think you meant this file instead?

MohamedBassem:
Thank you1

PR #1435, opened 2025-05-18
doc: new comunity project: freshrss to karakeep

thiswillbeyourgithub:
Just adding a link to my script that allows to send the RSS item I mark as *favourite* directly to karakeep as bookmarks.

Btw I don't know if you plan on doing this but on every new karakeep release I think it would be nice to shoutout new community projects made since the latest release in the release message, as I don't want to miss new contributions!


MohamedBassem:
Thanks a lot @thiswillbeyourgithub, I do shoutout new projects in every release :)

Issue #1403, opened 2025-05-14
doc: add page to detail the db structure


thiswillbeyourgithub:
### Describe the feature you'd like

Hi,

Similarly to how [open-webui](https://docs.openwebui.com/) has a documentation page detailing its own [database structure](https://docs.openwebui.com/tutorials/tips/sqlite-database) I think it would be useful to karakeep too.

It would explain how and where is each data stored etc.

### Describe the benefits this would bring to existing Karakeep users

Easier onboarding for other devs to collaborate, towards fostering a rich karakeep ecosystem.

### Can the goal of this request already be achieved via other means?

Not really

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### Additional context

Related to #1290

PR #1441, opened 2025-05-19
doc: add page detailling db structure and tables

thiswillbeyourgithub:
This was made to help solve #1403

This was made purely by prompting two state of the art LLMs based on `./packages/db/schema.ts` until they converged and stopped modifying anything, as I figured their combined error rates is far below mine. Surprisingly it took more than 10 iterations.

Hence this PR must be proof read by someone who is familiar with sqlite, and was done to reduce the burden on the dev team's shoulders.

Let me know if that's an issue!

### DIFF
--- a/docs/docs/07-Development/03-database.md
+++ b/docs/docs/07-Development/03-database.md
@@ -1,4 +1,6 @@
-# Database Migrations
+# Database
+
+## Database Migrations
 
 - The database schema lives in `packages/db/schema.ts`.
 - Changing the schema, requires a migration.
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thanks for working on this, but I think the question is why is this better than just adding a link to the `schema.ts` file? The schema changes frequently and this will quickly go out of date. And I don't think it adds much info more than that of the `schema.ts`? If something is not clear in `schema.ts` we can document it there instead of in the docs?

thiswillbeyourgithub:
You are right. Thank you.

Issue #1442, opened 2025-05-19
FR: use video subtitles as article content


thiswillbeyourgithub:
### Describe the feature you'd like

Currently when bookmarking a youtube video, the subtitles are not downloaded. It would be nice if it were and was part of the "article content", this way we could highlight it.

### Describe the benefits this would bring to existing Karakeep users

Enhanced bookmark

### Can the goal of this request already be achieved via other means?

No

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

Eragos:
Hey!

You can use `CRAWLER_YTDLP_ARGS` here the [Documentation](https://docs.karakeep.app/configuration#crawler-configs). 
In the [yt-dlp Documentation](https://github.com/yt-dlp/yt-dlp?tab=readme-ov-file#general-options) you can find the possible subtitles options.

Best Michael

thiswillbeyourgithub:
Thanks but that's still unclear: if karakeep receives a youtube link, it will call yt-dlp and expect a video. But can karakeep also get the subtitles? As they're not a video. If so where would it appear in the UI once it's downloaded?

beergeekdotcom:
It would fit well in the "Reader View" or "Archive" section of the item UI.

Reader View currently has the text scraped from the YT page -- which included partial description (view more.. cut off by default on YT). and a mush of of the comments (no usernames, dates, etc.)

Ideally it would be like the following:

Video Description:

... full video description ...

Video Transcript:

... full video transcript ...

Top Comments:

Username - Date - Comment
Username - Date - Comment
Username - Date - Comment
...

thiswillbeyourgithub:
That's pretty much what I had in mind yeah! I think "Reader View' would be the only appropriate choice at that's the only place we can highlight stuff.

Altranoetic:
I tried to download subs using the built in arguments, and it worked but doesn't show up on the UI at all. Doing `--embed-subs` broke it, so no luck there.

I personally would LOVE this feature. It'd make all the difference for me in downloading videos.

dimitrieh:
Required behaviour i would say, right now the summarisation and tagging isn't based on a lot of data.

current config:
```
INFERENCE_CONTEXT_LENGTH=16384
CRAWLER_VIDEO_DOWNLOAD=true
CRAWLER_VIDEO_DOWNLOAD_MAX_SIZE=-1
[...]

PR #1443, opened 2025-05-20
fix: Improve videoWorker error handling and use data directory for do…

erik-nilcoast:
…wnloads


thiswillbeyourgithub:
Hey I just happened to create #1442 so given you seem to have recently looked at the yt-dlp code, would you know if it's possible to, by adding yt-dlp args through the env variable, download the subtitles in a way thay makes them appear in the reader view?

apps/web/components/dashboard/bookmarks/UpdatableBookmarksGrid.tsx
```diff
@@ -51,7 +51,7 @@ export default function UpdatableBookmarksGrid({
 
   const grid = (
     <BookmarksGrid
-      bookmarks={data.pages.flatMap((b) => b.bookmarks)}
+      bookmarks={data?.pages.flatMap((b) => b.bookmarks) || []}
```
xuatz: if you want to use an empty array fallback, you can declare it as a const at the top level, or memo `bookmarks`, otherwise it [may cause unnecessary rerenders in certain situations](https://github.com/TanStack/table/issues/4240#issuecomment-1200893711) 

Or alternatively, I prefer if we update `<BookmarksGrid />` to accept undefined `bookmarks` gracefully.

erik-nilcoast: Happy to do whatever works. This was flagged as lint in the repo. I'll just back out the change if I can get the build to run and commit locally.

xuatz: I only commented because I noticed the change from your improvement 😄 

"declare it as a const at the top level" is probably the most straightforward option without introducing extra refactoring outside the scope of your intended PR.

thiswillbeyourgithub:
> > Hey I just happened to create #1442 so given you seem to have recently looked at the yt-dlp code, would you know if it's possible to, by adding yt-dlp args through the env variable, download the subtitles in a way thay makes them appear in the reader view?
> 
> The only reliable way I know to get the subtitles file is to scrape the xml on the client (transcripts aren't available in the API afaik). This is should be doable with the Chrome extension, but would require changes similar to the singlefile plugin settings to enable.

Thanks. So my  FR at #1442 still holds as the hard stuff (integrating yt-dlp, having the UI ready for videos) is already done, it's just a matter of connecting the pipes together. Thanks.

apps/web/components/dashboard/bookmarks/UpdatableBookmarksGrid.tsx
erik-nilcoast: Oh totally, I'll see if I can back this out and still commit.

erik-nilcoast:
Correct, grabbing subtitles from youtube is only possible on the client without trying to bypass their controls. This is doable, and I've done it using this code in a browser extension. This is not in the karakeep extension.

```
 * @param transcriptContent XML content from YouTube transcript
 * @returns Formatted transcript with timestamps
 */
function parseTranscript(transcriptContent: string): string {
  console.log('consoleContent -- starts here', transcriptContent);
  try {
    // Parse the XML content using fast-xml-parser
    const parser = new XMLParser({
      ignoreAttributes: false,
      attributeNamePrefix: ''
    });
    const xmlObj = parser.parse(transcriptContent);

    // Format the transcript
    let formattedTranscript = '';

    // Get all the text elements
    const textElements = xmlObj.transcript?.text || [];

    // Handle both array and single object cases
    const textArray = Array.isArray(textElements) ? textElements : [textElements];

    for (const textElement of textArray) {
      // Get the start time and duration
      const start = parseFloat(textElement.start || '0');
      const duration = parseFloat(textElement.dur || '0');
      const end = start + duration;

      // Format the timestamps
      const startFormatted = formatTimestamp(start);
      const endFormatted = formatTimestamp(end);

      // Get the text content and replace any HTML entities
      let text = textElement['#text'] || '';
      text = text.replace(/&#39;/g, "'");

      // Add to the formatted transcript
      formattedTranscript += `[${startFormatted} - ${endFormatted}] ${text}\n`;
    }

    console.log('formattedTranscript', formattedTranscript);

    return formattedTranscript;
  } catch (error) {
    console.error('Error parsing transcript:', error);
    return '';
  }
}

```

thiswillbeyourgithub:
I don't understand: yt-dlp is already included in karakeep and allows to direct subtitles directly.

For example to download the subs of [that reticulum conference](https://www.youtube.com/watch?v=pnyuSWy66cI') but not the video itself:
`yt-dlp 'https://www.youtube.com/watch?v=pnyuSWy66cI' --write-subs --skip-download --write-auto-subs`

So what I'm saying is that **if** I add `--write-subs --write-auto-subs` into the env variable, then it would be great if all the new vtt files created could be parsed into text and into the reader view.

I even coded [something](https://github.com/thiswillbeyourgithub/wdoc/blob/82b7f8150dab63b6cc13fc133df6622ab0021f23/wdoc/utils/loaders.py#L2559) in the passed to turn vtt format into something human readable yet without losing timestamps by just keeping the timestamp in HH:MM:SS format every 15 seconds. The output is super readable and could be useful to highlight.

erik-nilcoast:
> I even coded [something](https://github.com/thiswillbeyourgithub/wdoc/blob/82b7f8150dab63b6cc13fc133df6622ab0021f23/wdoc/utils/loaders.py#L2559) in the passed to turn vtt format into something human readable yet without losing timestamps by just keeping the timestamp in HH:MM:SS format every 15 seconds. The output is super readable and could be useful to highlight.

Yup that's basically what I'm doing (in the linked code), if you've find a reliable way to do it via ytdlp, that's great, use that!

thiswillbeyourgithub:
I am sorry for taking more of your time but I must be missing something. yt-dlp is *embedded* into karakeep right? Then why do you bring up the browser extension at all?

> if you've find a reliable way to do it via ytdlp

ytdlp can already reliably download the subtitles, and the rest is just a bunch of regex.


Basically if you could just tell me: is there any technical obstacle I am not seeing in order to pipe "output of ytdlp" to "bookmark's reader view"? Depending on your answer I *might* someday make the PR myself if I'm the only one to need subtitles, but if you tell me that it would somehow involve the browser extension then it would certainly be out of my league.

If you happen to know the files I should take a look in priority to be more familiar with the way videos are handled I would really appreciate it but it's okay if you don't have the time.

erik-nilcoast:
@thiswillbeyourgithub sorry, no clue. I'm not sure I can provide all of that context. I just occasionally fix things in this codebase. Videos weren't downloading correctly, with this change, now they do. I suspect that'll make whatever you're trying to do with subtitles a lot simpler. I've provided examples in the docs for how to modify the parameters of ytdlp as I originally wrote that feature. I can confirm that the CLI flags that I am passing:

```
        CRAWLER_VIDEO_DOWNLOAD_TIMEOUT_SEC = 1800
        # Updated configuration for more robust YouTube downloads
        CRAWLER_YTDLP_ARGS = "--no-call-home%%--geo-bypass%%--force-ipv4%%--no-check-certificate%%--prefer-insecure"
      }

```

That I use are correctly being passed through. I can pass through literally anything that ytdlp accepts and that is what will be used when invoking that tool in the `videoWorker.ts` file.

thiswillbeyourgithub:
Thank you very much. I might do that someday then.

apps/workers/workers/videoWorker.ts
```diff
@@ -24,7 +24,10 @@ import {
 import { withTimeout } from "../utils";
 import { getBookmarkDetails, updateAsset } from "../workerUtils";
 
-const TMP_FOLDER = path.join(os.tmpdir(), "video_downloads");
+// Use the configured data directory if available, fallback to tmp directory otherwise
+const TMP_FOLDER = serverConfig.dataDir
```
MohamedBassem: It's fair to assume that the data dir is always available.

apps/workers/workers/videoWorker.ts
```diff
@@ -24,7 +24,10 @@ import {
 import { withTimeout } from "../utils";
 import { getBookmarkDetails, updateAsset } from "../workerUtils";
 
-const TMP_FOLDER = path.join(os.tmpdir(), "video_downloads");
+// Use the configured data directory if available, fallback to tmp directory otherwise
+const TMP_FOLDER = serverConfig.dataDir
+  ? path.join(serverConfig.dataDir, "video_downloads")
+  : path.join(os.tmpdir(), "video_downloads");
```
MohamedBassem: can we name this `tmp_video_downloads` now that it's going to be in a persistent directory? Also, can we clean everything inside it on worker startup? Otherwise we might be leaking data there.

MohamedBassem (changes requested):
Thanks for the PR, left some small comments.

ebenoist:
@MohamedBassem np, all fixed up.

MohamedBassem:
@ebenoist thanks for addressing the comments, I can't seem to be able to edit the PR, so can you apply the following patch, so that I can merge the PR?

```diff
diff --git a/apps/workers/index.ts b/apps/workers/index.ts
index 1cc1ce49..fccda434 100644
--- a/apps/workers/index.ts
+++ b/apps/workers/index.ts
@@ -34,7 +34,7 @@ async function main() {
     OpenAiWorker.build(),
     SearchIndexingWorker.build(),
     TidyAssetsWorker.build(),
-    VideoWorker.build(),
+    await VideoWorker.build(),
     FeedWorker.build(),
     AssetPreprocessingWorker.build(),
     WebhookWorker.build(),
diff --git a/apps/workers/workers/videoWorker.ts b/apps/workers/workers/videoWorker.ts
index f2f61cfc..6f1772a4 100644
--- a/apps/workers/workers/videoWorker.ts
+++ b/apps/workers/workers/videoWorker.ts
@@ -1,5 +1,4 @@
 import fs from "fs";
-import * as os from "os";
 import path from "path";
 import { execa } from "execa";
 import { DequeuedJob, Runner } from "liteque";
@@ -27,10 +26,10 @@ import { getBookmarkDetails, updateAsset } from "../workerUtils";
 const TMP_FOLDER = path.join(serverConfig.dataDir, "tmp_video_downloads");
 
 export class VideoWorker {
-  static build() {
+  static async build() {
     logger.info("Starting video worker ...");
-    
-    cleanupTmpVideoDownloads();
+
+    await cleanupTmpVideoDownloads();
 
     return new Runner<ZVideoRequest>(
       VideoWorkerQueue,
@@ -215,17 +214,26 @@ async function findAssetFile(assetId: string): Promise<string | null> {
   return null;
 }
 
-function cleanupTmpVideoDownloads(): void {
+async function cleanupTmpVideoDownloads() {
   try {
-    if (fs.existsSync(TMP_FOLDER)) {
-      const files = fs.readdirSync(TMP_FOLDER);
-      for (const file of files) {
-        const filePath = path.join(TMP_FOLDER, file);
-        fs.rmSync(filePath, { force: true });
-      }
-      logger.info(`[VideoWorker] Cleaned up ${files.length} files from tmp_video_downloads`);
+    if (!fs.existsSync(TMP_FOLDER)) {
+      return;
     }
+    const files = await fs.promises.readdir(TMP_FOLDER);
+
+    Promise.all(
+      files.map(async (file) => {
+        const filePath = path.join(TMP_FOLDER, file);
+        await fs.promises.unlink(filePath);
+      }),
+    );
+
+    logger.info(
+      `[VideoWorker] Cleaned up ${files.length} files from tmp_video_downloads`,
+    );
   } catch (error) {
-    logger.error(`[VideoWorker] Failed to cleanup tmp_video_downloads: ${error}`);
+    logger.error(
+      `[VideoWorker] Failed to cleanup tmp_video_downloads: ${error}`,
+    );
   }
 }
```

MohamedBassem:
Closing the PR due to inactivity. Feel free to resend it if it's still needed.

PR #1451, opened 2025-05-22
doc: Update database docs

haappi:
Noticed a few command were wrong also removed a bunch of +'s that caused horizontal scrolling (but then again who doesn't love pluses!)


haappi:
Severely borked my git repo I think

MohamedBassem:
Thanks @haappi!

PR #1471, opened 2025-05-26
karakeep-linux.sh fix: use 'append' instead of 'file' for logs

vhsdream:
I noticed that using the log for StdOut and StdError would stop working under certain circumstances. Using 'append' instead of 'file' allows the file to be read while opened.


MohamedBassem:
My slight concern of the use of append is that the log file will probably grow indefinitely because there's no log rotation set up, right?

vhsdream:
> My slight concern of the use of append is that the log file will probably grow indefinitely because there's no log rotation set up, right?

Sorry about the delay - there is now 😄

MohamedBassem:
Thank you!

Issue #1406, opened 2025-05-14
FR: taking rules to the next level


thiswillbeyourgithub:
### Describe the feature you'd like

Regarding links/text bookmarks: I'm suggesting the following new rule engine features:

To the "Whenever" section:
New:
- An asset was added/removed to the bookmark
- A highlight was added/removed to the bookmark

Enhancements: 
- Favourite -> A bookmark's Favourite status has changed
    - add the options "added to/removed from favourite"
    - same for archive
- A bookmark is added to this list -> A bookmark is added to **a** list
    - and include an "Any list" option at the top (akin to wildcards)
- This tag was added to a bookmark -> A tag was added to a bookmark
    - and add options "Any tag", "Any human tag", "Any AI tag"
    - Same for removing a tag

To the "If" section:
- "contains archive": whether this bookmark has an "Archive" content (not sure if bookmarks can actually be missing this)
- condition on the actual text content (i.e. the reader view content, which is the thing you can add highlights to)
    - matches regex
    - character count + greater/smaller than + int
    - word count greater/smaller than + int
    - line count greater/smaller than + int
- same but for the raw html content (i.e. I think "archive" is not always available, hence the "contains archive")
- "contains video": whether the bookmark contains a video
    - video length
    - video title
    - contains subtitle or not
- "contains screenshot": whether the bookmark contains a screenshot

To the actions:
- prepend/append to the reader view + some html
- create new list/sublist/smartlist

Additionally:
- I think the "Whenever" should be renamed "triggers" to make them easy to find in the code, documentation, api and discussion

### Describe the benefits this would bring to existing Karakeep users

To me the most important is the rules pertaining to the textual content of bookmarks, as it would allow adding tags that dynamically adjust to the reading time of the article.
Also it's empowering users.

### Can the goal of this request already be achieved via other means?

Not really, the rule engine has great potential but cannot for now do complex things

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### Additional context

I would also be nice to add a "estimated reading time has changed" when #1079 is done

Fmstrat:
Seconding the raw archive HTML content. A good example of this is I have a list called Lemmy. I'd like to know of the HTML content contains "join-lemmy.org" and if so, add it to the Lemmy list.

hamidtech:
I really enjoyed the video rules section.
I hope it becomes available soon!

thiswillbeyourgithub:
[...]

PR #1474, opened 2025-05-26
feat: add optional `thinking` key to tagging response schema

thiswillbeyourgithub:
In my experience this helps all (both thinking and non thinking) models to provide better answers. I have not seen any negative side effects from that commit so far.

### DIFF
--- a/apps/workers/workers/inference/tagging.ts
+++ b/apps/workers/workers/inference/tagging.ts
@@ -23,6 +23,7 @@ import {
 } from "@karakeep/shared/queues";
 
 const openAIResponseSchema = z.object({
+  thinking: z.string().optional(),
   tags: z.array(z.string()),
 });
 
--- a/packages/shared/prompts.ts
+++ b/packages/shared/prompts.ts
@@ -22,7 +22,7 @@ Please analyze the attached image and suggest relevant tags that describe its ke
 - Aim for 10-15 tags.
 - If there are no good tags, don't emit any.
 ${customPrompts && customPrompts.map((p) => `- ${p}`).join("\n")}
-You must respond in valid JSON with the key "tags" and the value is list of tags. Don't wrap the response in a markdown code.`;
+You must respond in valid JSON with optional "thinking" key for your reasoning and required "tags" key with the value as list of tags. Don't wrap the response in a markdown code.`;
 }
 
 export function buildTextPrompt(
@@ -44,7 +44,7 @@ ${customPrompts && customPrompts.map((p) => `- ${p}`).join("\n")}
 CONTENT START HERE
 ${c}
 CONTENT END HERE
-You must respond in JSON with the key "tags" and the value is an array of string tags.`;
+You must respond in JSON with optional "thinking" key for your reasoning and required "tags" key with the value as an array of string tags.`;
 
   const promptSize = calculateNumTokens(constructPrompt(""));
   const truncatedContent = truncateContent(content, contextLength - promptSize);


thiswillbeyourgithub:
I added 6 more commits, I can easily remove any of them of you disagree, just let me know.

For context, I have some experience with prompt engineering, notably in [wdoc](https://github.com/thiswillbeyourgithub/wdoc/) which is kind of like what's now called deep research but on your own data.

Here is the rationale for each:

### Non controversial

- prompt: fix indent 98e10a7
    - nobrainer: There is a useless indent I removed

- prompt: remove extra 'language' word 98c2e8f
    - minor: The grammar feels incorrect: the `tag is in english` not `tag language is in english`, for example english language is french if `anglais`.

- prompt: use xml as separator 5ffdb47
    - enh: In my experience, using xml as separator yields better prompt adherence

- prompt: propose a tag_error 78c5099
    - enh: instead of telling to not return a tag in case of issue, it's better practice IMO to let the LLM tell us about it. If there is a `tag_error` tag, we can more easily inspect our database later.

- prompt: dont tell bot its a bot 3232704
    - enh: it's always better to tell the bot that it's good than tell it's a bot. This overlaps with #505

### Suggestion

- prompt: don't ask to include website tags 8f698de
- It makes no sense to ask to make tags for websites. For example I read quite a lot of substack.
1. There is zero benefit for me to be able to find substack articles by their tag instead of typing `substack` in the search bar. There are downsides to it though: LLMs are probabilistic so I only have maybe half of my substack bookmarks that are labelled `substack`.
2. The less instructions the better, given we are using small quantized models.
3. The website "information" is already part of the bookmark metadata, there is zero benefit to having it as a tag.
4. If I read mostly substack, this tag would be everywhere so won't have value. If I barely do, I won't care if this tag exists (especially if it's probabilistic), and if I actually need a tag for substack I'll use a rule for it. I rarely read the NYT but I read substack: I will make a rule for substack but not for the NYT. Also related to #1406
5. If you ask to produce 3 to 5 tags, and have 1 taken by the website you are actually asking for 2 to 4 useful tags. To me it's unecessarily low.

- prompt: aim for 5 tags 2c1f2f2
    - If you tell me `aim for 5` that means that `4` is most likely okay. But if you say `aim for 3-5` (i.e. giving a "range") then that means that `2` and `6` are okay. I do think that `2` (or actually `1` if there is a website tag) is not enough, hence that commit.
    - I do think that the number of tags should be a user defined setting that defaults to 5, in the meantime getting rid of the range is an enhancement


### Conclusion


Let me know if you disagree with any and I can

MohamedBassem (changes requested):
First of all, thanks for the PR. I'd really recommend discussing some of the changes suggested in an issue first to make sure we're aligned so that we can avoid any wasted effort.

My thoughts:
1. We're not going to be able to change the prompt output structure. Given that a lot of people use ollama, any slight modification to the prompt structure will break a ton of things to a lot of people.
2. I don't think the "thinking" adds a lot of value anyways. It's going to be logged in the logs, and users don't typically read the logs. If anything, it makes the tagging more expensive because of the increase in output tokens.
3. I'm honestly not a big fan of the `tag_error` issue. Also, I don't really see a reason why the LLM would decide that "something went wrong". Usually it just fails, and the things that go wrong happen in the parsing, etc.

I'm more ok with some of the cosmetic changes, such as removing bot, or removing the word language. Any bigger changes to the prompt is risky.

thiswillbeyourgithub:
Thanks for the reply.

2. The value of the `thinking` is not to be read but to give more reflection to the LLM. It increases the quality of the answers as well as the prompt adherence.
3. Exactly as you said: `the things that go wrong happen in the parsing`. Right now if you have a catastrophic failure of the parsing you would not really know it. I've had this happen many times in the [open-webui](https://github.com/open-webui/open-webui/) project: something somewhere breaks and the RAG is now broken. But it takes weeks for people to notice because LLMs are incredibly apt liars. I'm sure that karakeep tagging with no content and only the title would be pretty hard to spot by judging from the tags alone. Anyway I won't force it upon you if you disagree that's fine.


I just forced push on this PR to remove the `thinking` key as well as the instruction to add a `tag_error` in case of issue.

So this should be mergeable.

PR #1475, opened 2025-05-26
ui: fix background job page width constraint on vertical screen

thiswillbeyourgithub:
It was only using the left half of the screen and with that PR it now uses all width

### DIFF
--- a/apps/web/components/admin/BackgroundJobs.tsx
+++ b/apps/web/components/admin/BackgroundJobs.tsx
@@ -96,7 +96,7 @@ function AdminActions() {
     });
 
   return (
-    <div className="flex flex-col gap-2 sm:w-1/2">
+    <div className="flex flex-col gap-2">
       <ActionButton
         variant="destructive"
         loading={isRecrawlPending}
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Wouldn't it look nicer if we center it instead of expanding it?

thiswillbeyourgithub:
I don't think so. Here is the centered version:

image



And "my" version:
image


I *can* imagine some people with super extra wide screens might have issue but I'd argue that browser windows are not expected to be on super extra wide screens whereas phones and vertical screens are pretty standard. Like, if I had a 2m wide screens, as you can see on the first image, the left sidebar would be super far away anyway. I am the opposite of a reference on the topic though so feel free to tell me your preference.

MohamedBassem:
Honestly, not sure I really like the wide table. It makes the readability harder (by having to track the entire row to find the values). Also the buttons look weirder in that mode.

I'm planning to revamp the entire background jobs page anyways, to make it more informative, and looks nicer. So I think it's better to close this PR for now.

Issue #1413, opened 2025-05-15
Allow secrets to be stored in files locally instead of plaintext


debsidian:
### Describe the feature you'd like

While creating a docker-compose for Karakeep there are a couple of environmental variables that I'm uncomfortable leaving in plaintext.

- `OAUTH_CLIENT_SECRET`
- `NEXTAUTH_SECRET`
- `OPENAI_API_KEY`
- `MEILI_MASTER_KEY`

It would be nice if we could reference a local file for sensitive information such as this. 

### Describe the benefits this would bring to existing Karakeep users

Benefits are security and privacy of sensitive data.

### Can the goal of this request already be achieved via other means?

Not to my knowledge.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

Eragos:
I'm always a security focused guy. But I don't get your point :-/ 

If you follow the basic security rules (secure passwords, SSL encryption, separate networks, latest updates/security fixes, you name it...). You shouldn't have a problem. Security comes from the weakest point of the whole chain. And a plain text file in a secured context is IMHO not the problem.

Some things you can read:
- https://openid.net/developers/how-connect-works/
- https://help.openai.com/en/articles/5112595-best-practices-for-api-key-safety
- https://www.meilisearch.com/docs/learn/security/basic_security

I invite you to further discuss this topic in the discussion section...

_jmy2ct_ 

Best Michael

debsidian:
Maybe I'm miscommunicating here? I'm advocating for basic security practices. The ones that are baked into Docker's own documentation:

> A secret is any piece of data, such as a password, certificate, or API key, that shouldn’t be transmitted over a network or stored unencrypted in a Dockerfile or in your application’s source code.
>
> Docker Compose provides a way for you to use secrets without having to use environment variables to store information. If you’re injecting passwords and API keys as environment variables, you risk unintentional information exposure. Services can only access secrets when explicitly granted by a secrets attribute within the services top-level element.
>
> Environment variables are often available to all processes, and it can be difficult to track access. They can also be printed in logs when debugging errors without your knowledge. Using secrets mitigates these risks.
>
> [Source](https://docs.docker.com/compose/how-tos/use-secrets/)

No third party links necessary. Docker tells you what best practices are. I'm just asking that they be implemented here. It's okay if the answer is 'no', but I figured it couldn't hurt to ask.

MohamedBassem:
[...]

PR #1479, opened 2025-05-27
doc: mention stance on encryption

thiswillbeyourgithub:
As mentioned in discussion #1413 here is a preliminary doc explaining why karakeep is not encrypted (yet?)

### DIFF
--- a/docs/docs/08-security-considerations.md
+++ b/docs/docs/08-security-considerations.md
@@ -1,5 +1,7 @@
 # Security Considerations
 
+## Crawler and networking
+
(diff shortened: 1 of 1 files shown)


MohamedBassem:
I honestly don't think that this is worth mentioning. There are very few self hosted services that does encryption at rest. Also the note about administrators having access to the data is true even if there's encryption at rest because in that world, the server will still need to hold the decryption key to function (e.g. automatic tagging, etc), which means that the administrator will have access to the decryption key.

Only situation where this is not true is if the data is E2E encrypted, and only the user holds the encryption key, but then, the server will merely be a storage/sync backend. I don't think this is in scope for karakeep anyways.

PR #1480, opened 2025-05-27
doc: add guide on dealing with too many tags

thiswillbeyourgithub:
Thought this troubleshooting section was needed


MohamedBassem:
Honestly, not sure if the guide is adding a lot of value. For example, it's not clear what the user needs to change in the prompt, or how the distinction between AI and human tags make a difference. If we want this to be useful, we can probably mention the special prompt tags (`$userTags`), etc. But I'm planning to write a bigger guide about the philosophy of tagging in karakeep. So for now I think I'll close this PR.

PR #1483, opened 2025-05-27
UI fix: move smart list doc link up to prevent misclick

thiswillbeyourgithub:
In the model to create a smart list, I moved the "Learn more about the search query language" link above the "Search Query" field, as otherwise the mouse cursor is exactly above that link when switching from Manual to Smart list. This is mainly a fix for touchscreen users as I misclicked several times when wanting to create new smart lists.

image

Edit: this seems to be maybe a somewhat wider problem, as I see this happen elsewhere too. For example when assigning a list to an already crawled bookmark, the "cancel" button can be clicked on right away when touching the list name:
image
Maybe there is a "general delay" we can set to prevent misclicks on touchscreens?

### DIFF
--- a/apps/web/components/dashboard/lists/EditListModal.tsx
+++ b/apps/web/components/dashboard/lists/EditListModal.tsx
@@ -352,6 +352,14 @@ export function EditListModal({
                   return (
                     <FormItem className="grow pb-4">
                       <FormLabel>{t("lists.search_query")}</FormLabel>
+                      <FormDescription>
+                        <Link
+                          href="https://docs.karakeep.app/Guides/search-query-language"
+                          className="italic"
+                        >
+                          {t("lists.search_query_help")}
+                        </Link>
+                      </FormDescription>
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Honestly, I'm not sure I understand the problem. Can you maybe record a quick screen recording showing it?

thiswillbeyourgithub:
Sure. It seems to only happen on my brave browser phone and not on my computer but I doubt I'm the only one. The images are taken from a zoomed in computer.

1. 
image
2. You want to click but because you have fat fingers you have to touch the middle of the button 
image
3. But when doing so, some browsers seem to think you click twice, the first click selects "smart list" but the second click selects "Learn more about the search query language" right away. 
image


I don't have any problem with brave on my phone but notice that weird clicking issue only on karakeep. It is probably related to the troubles I had with having highlighting work there.

So my suggested fix in the meantime was just to move that link vertically to not be where the clicking is likely to missfire.

PR #1484, opened 2025-05-27
doc: mention the list of example scripts for the community api

thiswillbeyourgithub:
I decided to specifically mention the community scripts otherwise I'm worried people would never see them even though they seem like useful to keep in minds. I think this would dramatically help to foster an contributing ecosystem.

### DIFF
--- a/docs/docs/13-community-projects.md
+++ b/docs/docs/13-community-projects.md
@@ -52,15 +52,13 @@ _By [@thiswillbeyourgithub](https://github.com/thiswillbeyourgithub/)_
 
 A python package to simplify access to the karakeep API. Can be used as a library or from the CLI. Aims for feature completeness and high test coverage but do check its feature matrix before relying too much on it.
 
-Its repository also hosts the [Community Script](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts):
+Its repository also hosts the [Community Script](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts), for example:
 
 | Community Script | Description | Documentation |
 |----------------|-------------|---------------|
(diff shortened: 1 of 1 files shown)


docs/docs/13-community-projects.md
```diff
@@ -52,6 +52,17 @@ _By [@thiswillbeyourgithub](https://github.com/thiswillbeyourgithub/)_
...
+| **Karakeep-List-To-Tag** | Converts a Karakeep list into tags by adding a specified tag to all bookmarks within that list. | [`Link`](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts/karakeep-list-to-tag) |
+| **Omnivore2Karakeep-Highlights** | Imports highlights from Omnivore export data to Karakeep, with intelligent position detection and bookmark matching. Supports dry-run mode for testing. | [`Link`](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts/omnivore2karakeep-highlights) |
+| **Omnivore2Karakeep-Archived** | Fixes the archived status of bookmarks imported from Omnivore by reading export data and updating Karakeep accordingly. | [`Link`](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts/omnivore2karakeep-archived) |
+| **pocket2karakeep-archived** | Fixes the archived status of bookmarks imported from Pocket by reading export data and updating Karakeep accordingly. | [`Link`](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts/pocket2karakeep-archived) |
```
MohamedBassem: Those two are now natively supported in karakeep, we can probably drop them.

thiswillbeyourgithub: Pushed something to address this.

(By the way, I'm not very good with pre commit hooks but if you could set them up so that they don't trigger when editing .md files it would be a bit better)

MohamedBassem: You can bypass them with passing `-n` to `git`, but if you skip them, it means that the lint/format CI actions will fail on the PR.

PR #1489, opened 2025-05-28
feat: Allow PATCH method in CORS headers

spasche:
Hello,

I'm using the [Karakeep Plugin for Obsidian](https://github.com/jhofker/obsidian-hoarder), which [uses the PATCH method of the API](https://github.com/jhofker/obsidian-hoarder/blob/8ec6ef3288d2acf71ab494cc07707c1ee8f73d09/src/main.ts#L295).

Special reverse proxy configuration is currently required for CORS, it would be nice if this could work out of the box by allowing the method directly in the backend.


MohamedBassem:
Thanks for the PR!

PR #1492, opened 2025-05-29
fix: Browser extension tab tracking and tags page performance

### LINKED ISSUE #1404: Bug: scaling issue as number of tags grows
### Describe the Bug

I happen to have thousands of tags and noticed that at least two pages are severely impacted because of this:
The `dashboard/tags` page is almost not even loading. And when it does it still has trouble scrolling:

Image

I suggest at the very least caching the tag list: there is no point in reloading the whole list from the db as long as the last tag db access has not changed

Not sure but maybe the tag list should have pagination with a search bar: when in the tags page, the top search bar should filter among the tags instead of among the bookmarks.

### Steps to Reproduce

1. Have thousands of tags on a powerful computer
2. Noticing struggling load time

### Expected Behaviour

AI Tags have to be scalable.

Mentioned initially in [that comment](https://github.com/karakeep-app/karakeep/issues/111#issuecomment-2864622221)

### Device Details

_No response_

### Exact Karakeep Version

0.24.1

AdrianAcala:
- Fix browser extension tab tracking by using currentWindow instead of lastFocusedWindow
  - Resolves issue where extension would track incorrect tab when popup was opened
  - Fixes issue #111 comment

- Optimize tags page performance for large tag collections
  - Add pagination support to tags.list API endpoint (limit/offset)
  - Limit initial tag load to 100 tags to improve page load time
  - Add search functionality to filter tags on client side
  - Add test coverage
  - Fixes #1404

### DIFF
--- a/apps/browser-extension/src/SavePage.tsx
+++ b/apps/browser-extension/src/SavePage.tsx
@@ -42,7 +42,7 @@ export default function SavePage() {
       if (!newBookmarkRequest) {
         const [currentTab] = await chrome.tabs.query({
           active: true,
-          lastFocusedWindow: true,
+          currentWindow: true,
         });
         if (currentTab?.url) {
           newBookmarkRequest = {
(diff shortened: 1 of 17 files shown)


haappi:
Any reason as to why this was closed? I believe it would fix both of those linked issues

AdrianAcala:
Oops. I think it got closed when I accidentally deleted the repo. I'll restore it shortly.

haappi:
Great, that'll be awesome. Thank you so much!

PR #1498, opened 2025-05-30
fix: Update docker development environment with automatic database migrations

AdrianAcala:
- Add automatic database migration execution in prep service
- Ensure database schema is applied before web/workers services start
- Resolves issues with docker-compose.dev.yml setup for complete development environment
- Database migrations now run automatically via `pnpm db:migrate` in prep container

This provides a fully functional development environment that mirrors production setup where migrations run automatically on container startup.

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -1,4 +1,3 @@
-version: "3.8"
 services:
   web:
     build:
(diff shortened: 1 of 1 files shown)


MohamedBassem:
The directory is already created by the mount, and the migration script already creates the database file. I don't think this PR adds anything that's not already there? If anything, it removes the `pnpm install` part.

PR #1501, opened 2025-05-30
Update 04-hoarder-to-karakeep-migration.md

MadMaximusJB:
Changed incorrect “HOARDER_VERSION” to “KARAKEEP_VERSION”


MohamedBassem (changes requested):
Thanks for the PR, you shouldn't be updating the versioned docs though. You should update the file under `docs` instead.

PR #1505, opened 2025-05-31
feat: add user customisable default archive display behaviour

xuatz:


apps/web/lib/i18n/locales/en/translation.json
```diff
@@ -101,9 +101,16 @@
...
+        "bookmark_click_action": {
+          "title": "Bookmark Click Action",
+          "open_external_url": "Open Original URL",
+          "open_bookmark_details": "Open Bookmark Details"
+        },
+        "archive_display_behaviour": {
+          "title": "Archive Display Mode",
+          "show": "Show up in tags and lists",
+          "hide": "Hide in tags and lists"
+        }
```
xuatz: thought it might be a little nicer to nest related text for each setting instead of having them all flat. but im totally fine to revert this change as well

xuatz:
I only finished the initial user settings implementation, but I think we can already discuss a little about the choice of text to see if it accurately describes what we want to do. Let me know if the text could be improved to more accurately describe the feature. @jl-678

apps/web/lib/i18n/locales/en/translation.json
```diff
@@ -101,9 +101,16 @@
       "interface_lang": "Interface Language",
       "user_settings": {
         "user_settings_updated": "User settings have been updated!",
-        "boomark_click_action": "Bookmark Click Action",
```
xuatz, line 104: typo fix
```diff
-  boomark_click_action
+  bookmark_click_action
```

apps/web/app/dashboard/lists/[listId]/page.tsx
```diff
@@ -23,10 +31,17 @@ export default async function ListPage({
...
+    ? searchParams.includeArchived === "true"
+    : userSettings.archiveDisplayBehaviour === "show"
+
   return (
     <BookmarkListContextProvider list={list}>
       <Bookmarks
-        query={{ listId: list.id }}
+        query={{
+          listId: list.id,
+          includeArchived,
```
Comment (MohamedBassem) [inline apps/web/app/dashboard/lists/[listId]/page.tsx]: here, if we pass `archived: false` when `!includeArchived`, we can avoid all the changes to trpc. If `includeArchived` is `true`, we don't need to pass any filtering by `archive`, so we pass `undefined`

packages/trpc/routers/users.test.ts
```diff
@@ -95,6 +95,7 @@ describe("User Routes", () => {
     await expect(() => user2Caller.users.list()).rejects.toThrow(/FORBIDDEN/);
   });
 
+  // TODO may want to add a test for `archiveDisplayBehaviour` as well
```
xuatz: are you strict with requiring tests? ^^

MohamedBassem: I think it's fine, the tests are mostly testing the edit/get functionality not the specific setting

MohamedBassem (approved):
Looks good to me! Did some small changes (e.g. nuqs), but other than that, this is ready to be merged!

jl-678:
Not sure if I should comment here, but I have been playing with this feature and notice that the setting is not maintained between sessions. Is that expected behavior?

xuatz:
> Not sure if I should comment here, but I have been playing with this feature and notice that the setting is not maintained between sessions. Is that expected behavior?

@jl-678 the global user setting is persisted. The list override is ephemeral as designed. You can set the default global setting to hide as always

jl-678:
Awesome, thank you.

PR #1506, opened 2025-05-31
CORRECTED: Update 04-hoarder-to-karakeep-migration.md

MadMaximusJB:
Corrected location of document update. Updated incorrect migration code.


MohamedBassem:
Thanks for re-opening the PR, actually checking again, people who are migrating will probably still have the `HOARDER_VERSION` env variable. So that's why the next line tells them to change the variable name is needed. As such, the current documentation is the correct one.

Issue #1068, opened 2025-02-24
[Android] Alternative options for viewing links in app


OJ7:
### Describe the feature you'd like

Right now, the app has a custom webview being used to display link content. Other apps I've used allow changing the setting to either use the app webview, custom tab, or default browser.

### Describe the benefits this would bring to existing Hoarder users

The app webview, while has a toolbar the bottom for quick actions, does not support features like forced dark mode or other features/extensions that some web browsers provide.

Example of app that provides similar options:
![options in Feeder app](https://github.com/user-attachments/assets/00a98d23-c9e8-4988-9fe5-0abf2ab97551)

### Can the goal of this request already be achieved via other means?

Not really. When opening a link, the webview opens up. The globe icon at the bottom right can be clicked to open the link with the default browser, but creates an extra step.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### Additional context

#536 is related to this and could also be another option to include for opening links.

Issue #662, opened 2024-11-16
Request: Setting to open content in a "Reader Mode" view by default


fenndev:
### Describe the feature you'd like

Currently, both on the web version and Android app, clicking on a bookmark opens the original article. I would like an option to open a "Reader Mode" style view of articles by default instead, as I would primarily use Hoarder as an Omnivore alternative.

### Describe the benefits this would bring to existing Hoarder users

For those looking to Hoarder as a replacement for Omnivore, this change would allow those users to more closely mimic what they would expect from a Read-It-Later service. 

Issue #536 would be a prerequisite for this feature request.

### Can the goal of this request already be achieved via other means?

Not currently.

### Have you searched for an existing open/closed issue?

- [X] I have searched for existing issues and none cover my fundamental request

MohamedBassem:
Yeah, I think we can support that.

ctschach:
Any update on this "little" feature request?

AlanOC91:
Would love an update on this. I'm not sure if "reader mode" is even available on the android app, I can't seem to find it.

MohamedBassem:
This is coming in the next release, and will be in the nightly build in ~20mins.

digithree:
@MohamedBassem I don't see this in the current code on `main`, is this still in flight for mobile? I have implemented this and can create a PR otherwise

MohamedBassem:
@digithree Thanks for the nice words! yeah, #536 is the better fit. I haven't personally started working on it, so if you already implemented something feel free to send as a PR and we can discuss it there! Also, most of the development discussion happens in the #development channel in discord here: https://discord.com/invite/NrgeYywsFh feel free to join us :)

digithree:
Sweet, I will join and push this up, cheers

Issue #536, opened 2024-10-14
Feature: view cached text/archive/screenshot in mobile app


thallada:
I downloaded the Android app and noticed that the expand button to see the cached text for articles is missing. When I click on an entry it only shows the thumbnail image, a link to the source page, the generated tags, and a text area for notes, but not the actual text of the article. Are there plans to add the ability to view the cached and archived content in the mobile apps?

I prefer to read the extracted text for articles in one interface versus visiting the source pages. This is the workflow I currently have in other read-it-later apps like Pocket or Readwise Reader. So, without this feature I am not going to switch to using Hoarder primarily.

If this is implemented, it would be really awesome if the data was cached on the device for the first N articles in my feed so that I could read articles offline.

fenndev:
Seconding this.

MohamedBassem:
Yeah, this is currently missing. Added it to the roadmap.

MohamedBassem:
This is implemented in #1509 and is coming in the next app release.

Issue #221, opened 2024-06-13
Feature Request: Reader Mode


cochraneserver:
Similar to how Pocket offers a view with just the text

lucius346346:
Isn't that what that "expand" button does?

PR #1509, opened 2025-05-31
feat(mobile): add reader view by default to bookmark detail view

digithree:
Creates a new reader view as the default display for bookmarked articles (links) when list items are tapped, i.e. the bookmark detail view.

## Description

When a link bookmark was successfully scraped from a web page source (likely an article) HTML will be available on request as `bookmark.content.htmlContent`. This PR makes the change firstly to request this by setting request flag `includeContent`, and then if content is present, to display it in-line in the app. If no content is available, the link will fallback to the previous behaviour of using an embedded

Note:
- the usage of `bookmark.content.htmlContent` is assumed safe and depends on any previous efforts to clean this at the time the bookmark was added and processed
- users may still access the original link using the link/globe icon in the app footer nav bar (right-most icon, see screenshot below)

## Testing

This change was tested on Android only. I was unable to get my iOS build environment configured yet so it would be good if someone could test this as an iOS build. I would be very confident it will work just as well on iOS.

## Related tickets

Implements: https://github.com/karakeep-app/karakeep/issues/536 basic implementation. Also mentioned is caching and screenshot view, which is not implemented here. Local DB storage of readable article content is out of scope for this PR but a good idea for another PR

See also:
- partially delivers https://github.com/karakeep-app/karakeep/issues/1068, adding the reader view but not any options to customise behaviour. This PR assumes that the most desired case is for reader view when readable text could be scraped. An additional PR could add customisation to what this PR delivers, if required
- https://github.com/karakeep-app/karakeep/issues/221 and https://github.com/karakeep-app/karakeep/issues/662 where the broad request for reader mode was mentioned (web has already implemented, this PR will now implement for mobile)

## Screenshots and video

Note, all media is from Android testing.

_Link bookmark detail view, reader view_

bookmark_link_reader

_Link bookmark detail view, fallback to WebView_

bookmark_link_webview

_Complete user journey, video_

https://github.com/user-attachments/assets/8f52be77-bedd-42c0-9d77-51276788ad44

### DIFF
--- a/apps/mobile/app/dashboard/bookmarks/[slug]/index.tsx
+++ b/apps/mobile/app/dashboard/bookmarks/[slug]/index.tsx
@@ -155,6 +155,67 @@ function BookmarkLinkView({ bookmark }: { bookmark: ZBookmark }) {
   if (bookmark.content.type !== BookmarkTypes.LINK) {
     throw new Error("Wrong content type rendered");
   }
+
+  if (bookmark.content.htmlContent) {
+    return (
+      <View className="flex-1 bg-background">
+        <WebView
+          originWhitelist={["*"]}
+          source={{
+            html: `
+              <!DOCTYPE html>
+              <html>
+                <head>
+                  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+                  <style>
+                    body {
+                      font-family: -apple-system, BlinkMacSystemFont, 'Segoe UI', system-ui, sans-serif;
+                      line-height: 1.6;
+                      color: #374151;
+                      margin: 0;
+                      padding: 16px;
+                      background: transparent;
+                    }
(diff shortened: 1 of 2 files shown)


digithree:
Dark mode support for mobile reader mode has been added

mobile-reader-dark-mode


https://github.com/user-attachments/assets/d16a8f62-9f40-49c7-ab14-f012a47b5c3d

digithree:
@MohamedBassem ready for review

MohamedBassem:
@digithree I've added implementation for the other views as part of this commit. Thank you for getting this started!

digithree:
@MohamedBassem my pleasure. Good thinking to get the other types covered, and good improvement readings loading handling on the link content, etc.

PR #1515, opened 2025-06-02
feat: preserve unsaved title changes when modifying bookmark tags

### LINKED ISSUE #1339: Unsaved Title Changes Lost When Modifying Tags
### Describe the Bug

 When editing a document, removing or modifying tags causes the dialog to refresh, resulting in the loss of unsaved changes to the document title.

### Steps to Reproduce

- Add any document. 
- Click Edit. 
- Change the document title (do not save yet).
- Remove or modify a tag. 
- Observe that the dialog refreshes and the unsaved title change is lost.

### Expected Behaviour

 Modifying tags should not reset or discard unsaved changes to the title or other fields.

_No response_

### Device Details

firefox 138

### Exact Karakeep Version

latest

AdrianAcala:
Prevents loss of unsaved title edits when users interact with tag selectors or other UI elements. Adds useDialogFormReset hook to maintain form state consistency across component re-renders.

Fixes #1339


apps/browser-extension/src/components/ui/dialog.tsx
```diff
@@ -19,7 +19,7 @@ const DialogOverlay = React.forwardRef<
   <DialogPrimitive.Overlay
     ref={ref}
     className={cn(
-      "fixed inset-0 z-50 bg-black/80  data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0",
+      "data-[state=open]:animate-in data-[state=closed]:animate-out data-[state=closed]:fade-out-0 data-[state=open]:fade-in-0  fixed inset-0 z-50 bg-black/80",
```
xuatz: pretty sure this was reordered by your local setup (maybe e.g. headwind?) and would fail the lint check eventually. wanna give you a heads up ^^

xuatz (approved):
feature works
aside from the tailwind order lint issues, LGTM 👍 
but let's see what @MohamedBassem thinks ^^

apps/browser-extension/src/components/ui/dialog.tsx
AdrianAcala: This is fixed! Good catch! Thanks!

MohamedBassem:
Thanks @AdrianAcala.

PR #1544, opened 2025-06-06
feat: upgrade to pnpm@10.11.1

xuatz:
## Summary

* upgrade key deps of apps/mobile
	* e.g. expo, react-native and associated deps
* upgrade to ~~latest version of~~ pnpm@9.15.9
* ~~revert pnpm@10 default behaviour to prevent deps lifecycle scripts frin running to the legacy behaviour where all deps scripts are allowed to run~~
* ~~**sneakily included a small showcase of pnpm catalogs** fe0ed8090cbe6f85128c2d083f6debb5ff780d17~~

## Details

There is a breaking change for contributors' local environment.
Because of the lockfile changes, you must remove ALL `node_modules` folders (and maybe .turbo, i did it but i never confirmed if it was necessary). Otherwise the application will not work properly.

You can use the following commands for convenience:
```bash
rm -rf node_modules \
	&& pnpm -r exec rm -rf node_modules \
	&& rm -rf .turbo \
	&& pnpm -r exec rm -rf .turbo;
```
Or just do it yourself, don't wipe your disk!

---

I think that maybe, we can do `v0.26.0` release that does not contain any new features, but only the pnpm upgrade?

Or perhaps include this PR for `v0.25.0`, but maybe give it a week or 2 in the nightly channel. Are there nightly builds for the react-native apps as well?

## Notes

A proof that a certain commit had passing ci jobs at least once in the past
* https://github.com/karakeep-app/karakeep/actions/runs/15496406769/job/43634748159?pr=1544


apps/mobile/package.json
xuatz, line 1: we can bring back the `^` and `~` if you prefer that, but i didn't want to disturb the working state ^^

packages/shared/utils/relativeDateUtils.ts
```diff
@@ -5,7 +5,7 @@ interface RelativeDate {
 }
 
 const parseRelativeDate = (date: string): RelativeDate => {
-  const match = date.match(/^([<>])(\d+)([dwmy])$/);
+  const match = /^([<>])(\d+)([dwmy])$/.exec(date);
```
xuatz, line 8: eslint "fixed" this 🤔

pnpm-lock.yaml
xuatz, line 1: this is the secret sauce generated with blood 😅

xuatz:
Mmm why did the trpc test suddenly fail 😅 

~~edit1: working on my machine tho lol~~

image

~~@MohamedBassem maybe can you try to rerun, or we might have to disable the pnpm cache in the ci jobs (wild suggestion)~~

edit2: oh i managed to reproduce it on local after wiping node_modules

edit3: i went back to the unsquashed commits and was able to get it working again on local. since this upgrade is in no rush to be shipped, i'll spend a bit more time to see which unpinned pkg caused the tprc tests to fail.

xuatz:
Turns out that the failing [job](https://github.com/karakeep-app/karakeep/actions/runs/15496995744/job/43636687770?pr=1544) is due to the change in behaviour starting in pnpm@10, where [lifecycle scripts of dependencies are not executed during installation by default](https://github.com/pnpm/pnpm/releases/tag/v10.0.0).

At this moment, I simply allow ALL deps to run the postinstall scripts (which is essentially the current existing behaviour), but in future we can consider to have a separate issue to identify the specific packages that we need to run `postinstall` for karakeep to work (e.g. trpc probably)

package.json
```diff
@@ -33,10 +33,11 @@
...
   },
   "prettier": "@karakeep/prettier-config",
-  "packageManager": "pnpm@9.0.0-alpha.8+sha256.a433a59569b00389a951352956faf25d1fdf43b568213fbde591c36274d4bc30",
+  "packageManager": "pnpm@10.11.1",
   "pnpm": {
     "patchedDependencies": {
       "xcode@3.0.1": "patches/xcode@3.0.1.patch"
-    }
+    },
+    "neverBuiltDependencies": []
```
xuatz: this is needed otherwise `pnpm -F trpc test` would fail.

xuatz: removed, as I ultimately decided to upgrade to v9 instead. 

There is a change in `pnpm deploy` behaviour introduced starting from pnpm@10, that causes some problem with karakeep's current setup.
```
failed to solve: process "/bin/sh -c pnpm deploy --node-linker=isolated --filter @karakeep/workers --prod /prod/workers" did not complete successfully: exit code: 1
 > [web base  9/11] RUN pnpm deploy --node-linker=isolated --filter @karakeep/workers --prod /prod/workers:
0.651  ERR_PNPM_DEPLOY_NONINJECTED_WORKSPACE  By default, starting from pnpm v10, we only deploy from workspaces that have "inject-workspace-packages=true" set
0.651 
0.651 If you want to deploy without using injected dependencies, run "pnpm deploy" with the "--legacy" flag or set "force-legacy-deploy" to true
```

apps/mobile/package.json
```diff
@@ -19,52 +19,52 @@
...
+    "expo-web-browser": "^14.0.2",
     "lucide-react-native": "^0.354.0",
     "nativewind": "^4.1.23",
     "react": "^18.3.1",
-    "react-native": "0.76.3",
+    "react-native": "0.76.9",
     "react-native-awesome-slider": "^2.5.3",
     "react-native-blob-util": "^0.21.2",
-    "react-native-gesture-handler": "~2.21.2",
+    "react-native-gesture-handler": "~2.20.2",
```
xuatz, line 48: any minor version changes in `apps/mobile` were made by migration scripts provided by react-native and expo.

MohamedBassem:
Thanks a lot @xuatz it finally happened! :)

PR #1464, opened 2025-05-24
feat: handle "deleted" webhook events

SalGnt:
Add support to a `deleted` webhook event, that is triggered when a user deletes a bookmark. Fixes #1463.

PR #1575, opened 2025-06-09
fix: webhook job failure due to bookmark id not found

SalGnt:
Fix the `error: [webhook][<jobId>] webhook job failed: Error: [webhook][<jobId>] bookmark with id <bookmarkId> was not found`. Relates to #1464.


MohamedBassem:
I don't think this fixes the issue. There's a bigger problem with the worker itself where it won't be able to find the user because the bookmark is deleted. I filed an issue #1613 to track this instead.

PR #1579, opened 2025-06-10
fix: prevent search list scroll position reset after bookmark interactions

### LINKED ISSUE #1521: Search Query list resets after closing focus pop-up
### Describe the Bug

https://github.com/user-attachments/assets/bdef0d66-fe1a-4104-b8a0-07cc4189aeb2

Using latest version and have had this issue with previous versions as well.

When I view a list of bookmarks through a search query, and the list is large enough to 'load more', any bookmark that I click on and focus on will immediately reset my list back to the bottom of the search query list before the 'load more'.

### Steps to Reproduce

1. Search any query that has enough results to reach the 'load more' 
2. Once new entries are loaded, click on an entry after the previous 'load more'.
3. Close out of the focused pop-up by either clicking outside the window or hitting the back button.
4. List immediately resets its place to before the 'load more' took place.

### Expected Behaviour

List should stay in the same position as I left it after closing out of the focus popup.

_No response_

### Device Details

_No response_

### Exact Karakeep Version

0.24.1 - docker compose latest

AdrianAcala:
Resolves #1521 where the search results list would jump back to its original scroll position after a user clicked on a bookmark entry and then closed the pop-up. This happened because bookmark mutations were invalidating the entire search query cache, causing the list to reset to page 1.

The fix implements proper state isolation between search results and bookmark mutations to preserve the user's scroll position and loaded results during interactions.

### DIFF
--- a/apps/web/app/dashboard/search/page.tsx
+++ b/apps/web/app/dashboard/search/page.tsx
@@ -2,20 +2,35 @@
 
 import { Suspense, useEffect } from "react";
 import BookmarksGrid from "@/components/dashboard/bookmarks/BookmarksGrid";
+import { SearchErrorBoundary } from "@/components/shared/SearchErrorBoundary";
 import { FullPageSpinner } from "@/components/ui/full-page-spinner";
 import { useBookmarkSearch } from "@/lib/hooks/bookmark-search";
 import { useSortOrderStore } from "@/lib/store/useSortOrderStore";
 
 function SearchComp() {
-  const { data, hasNextPage, fetchNextPage, isFetchingNextPage } =
+  const { data, hasNextPage, fetchNextPage, isFetchingNextPage, searchKey } =
     useBookmarkSearch();
 
-  const { setSortOrder } = useSortOrderStore();
+  const { setSortOrder, getSearchState } = useSortOrderStore();
 
   useEffect(() => {
     // also see related cleanup code in SortOrderToggle.tsx
     setSortOrder("relevance");
-  }, []);
+  }, [setSortOrder]);
+
+  // Restore scroll position when component mounts with existing data
+  useEffect(() => {
+    if (data && data.pages.length > 1) {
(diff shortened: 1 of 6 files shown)


apps/web/app/dashboard/search/page.tsx
```diff
@@ -35,8 +52,10 @@ function SearchComp() {
 
 export default function SearchPage() {
   return (
-    <Suspense>
-      <SearchComp />
-    </Suspense>
+    <SearchErrorBoundary>
```
xuatz, line 55: It's not immediately obvious to me, but what's the limitation of `react-error-boundary` that required a custom ErrorBoundary implementation?

If not, could we reuse our existing dependency for consistency and easier maintenance?

apps/web/app/dashboard/search/page.tsx
```diff
@@ -2,20 +2,35 @@
...
 
   useEffect(() => {
     // also see related cleanup code in SortOrderToggle.tsx
     setSortOrder("relevance");
-  }, []);
+  }, [setSortOrder]);
+
+  // Restore scroll position when component mounts with existing data
+  useEffect(() => {
+    if (data && data.pages.length > 1) {
```
xuatz, line 23:
```suggestion
    if (data?.pages.length > 1) {
```

xuatz (changes requested):
I don't think the fix is fully working, however the problem might not be introduced by this pr but already present in main. 
here is a reproduction of a failing case.

scroll restore not working

it seems like we have a bug where upon expanding a bookmark, and closing it, a lot of component rerendering happened. the search input was cleared, however it appears that the search query is still correct, but the sortOrder is wrong 🤔 

we might have to fix that together in this PR, or make a separate issue to fix that problem _first_ before we can return to this PR.

MohamedBassem:
I pushed a simpler fix to this problem here: https://github.com/karakeep-app/karakeep/commit/cf06bbb2b56690ccfe29e36cbde0d8a68ad04bad

PR #1595, opened 2025-06-11
feat: Clear all tags on mobile

Ashok28:
Added a "Clear All" button to the tag manager because AI often generates way too many tags and it's super annoying to remove them one by one. This makes it easier to clean everything up with a single tap.


MohamedBassem:
Thank you!

PR #1596, opened 2025-06-11
UI Fix: Immediately visible search bar in tags list

Ashok28:
Having the search bar hidden in tags list makes no sense, this makes sure it's always visible.


MohamedBassem:
Thank you!

PR #1620, opened 2025-06-16
fix: tweaks to mobile app icon padding + added iOS tinted variant

alexjsp:
Since I started using Karakeep I've found the mobile app icon didn't feel like it totally fit on my iOS homescreen, so I wanted to see if I could improve it. I think the biggest issue is the padding around the logo and icon edge felt too tight, and I didn't want to make any drastic changes, so just made a little tweak to the padding that I think makes things look nicer:

| Existing Icon | New Icon |
| ------- | ------- |
|  |  |

Comparison:
ezgif-74a161754d6e44

I also added a tinted variant that adds a little gradient and I think makes the icon look better in that tinted mode:
| Existing Icon | New Icon |
| ------- | ------- |
|  |  |

Feel free to bin this if you don't like it, or make any suggestions of further tweaks you'd like, happy to help there. But hopefully you like it, I think it looks a bit better on my homescreen!


MohamedBassem:
Thanks, it's a subtle change but I agree that it looks nicer!

PR #1624, opened 2025-06-17
fix(workers): video downloader should log yt-dlp errors

irobot:
In the event that yt-dlp errors out, the error details should be logged.
This can be helpful in deciding if and what extra arguments need to be passed via `CRAWLER_YTDLP_ARGS`.
yt-dlp prints out the error message to stderr.


MohamedBassem (approved):
LGTM as well, will merge after the CI run!

MohamedBassem:
Thanks @irobot!

Issue #1541, opened 2025-06-06
FR: search history


thiswillbeyourgithub:
### Describe the feature you'd like

I think the top search bar should make it easy to redo passed searches. For example via up/down key arrows. On mobile I guess a small "down arrow" icon on the right of the search bar could trigger a drop down to select among the last few searches.

### Describe the benefits this would bring to existing Karakeep users

Better UX

### Can the goal of this request already be achieved via other means?

No

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

lexafaxine:
Great suggestion! I think for both web/mobile app we could leverage local storage to do this. Hi @MohamedBassem  I would be happy to take this on. Please let me know if I can be assigned to this issue :)

MohamedBassem:
@lexafaxine it's all yours! Looking forward for the PR!

PR #1627, opened 2025-06-17
feat: #1541 adding search history

lexafaxine:
#### Add search history for web
Use local storage to store last 5 search history,
The popover showing search history will appear when input is focused.
According to the current search query implementation, history will only be saved to local storage after the input is blur, to prevent unintended saving.

#### Add search history for mobile
Use react-native async-storage to store last 5 search history,
Because we are showing user search history now in the search screen, user need to hit return to submit a search.
(Please refer to the demo video)

### Demo video

web: https://drive.google.com/file/d/1VCfapgyx8nqwxw7JajyC8HlXiJYvjNR7/view?usp=sharing
mobile: https://drive.google.com/file/d/1b5Xj5ekeFxb6IrzoWVUmZpMLWBHpCnBk/view?usp=sharing

### DIFF
--- a/apps/web/components/dashboard/search/SearchAutocomplete.tsx
+++ b/apps/web/components/dashboard/search/SearchAutocomplete.tsx
@@ -1,0 +1,60 @@
+import React from "react";
+import {
+  Command,
+  CommandEmpty,
+  CommandGroup,
+  CommandItem,
+  CommandList,
+} from "@/components/ui/command";
+import { useSearchHistory } from "@/lib/hooks/useSearchHistory";
+import { cn } from "@/lib/utils";
+import { History } from "lucide-react";
+
+interface SearchHistoryAutocompleteProps {
+  searchQuery: string;
+  onSelect: (searchTerm: string) => void;
+  className?: string;
+}
+
+export function SearchAutocomplete({
+  searchQuery,
+  onSelect,
+  className,
+}: SearchHistoryAutocompleteProps) {
+  const { history } = useSearchHistory();
+
+  const filteredHistory = searchQuery
+    ? history.filter((term) =>
+        term.toLowerCase().includes(searchQuery.toLowerCase()),
+      )
+    : history;
+
+  if (history.length === 0) {
+    return (
+      <div className="p-4 text-center text-sm text-muted-foreground">
+        Search your bookmarks...
+      </div>
+    );
+  }
+
+  return (
+    <Command shouldFilter={false} className={cn(className)}>
+      <CommandList>
+        <CommandEmpty>No matching recent searches.</CommandEmpty>
+        <CommandGroup heading="Recent Searches">
+          {filteredHistory.map((term) => (
+            <CommandItem
+              key={term}
+              value={term}
+              onSelect={() => onSelect(term)}
+              className="cursor-pointer"
+            >
+              <History className="mr-2 h-4 w-4" />
+              <span>{term}</span>
+            </CommandItem>
+          ))}
+        </CommandGroup>
+      </CommandList>
+    </Command>
+  );
+}
--- a/apps/web/components/dashboard/search/SearchInput.tsx
+++ b/apps/web/components/dashboard/search/SearchInput.tsx
@@ -3,17 +3,25 @@
 import React, { useEffect, useImperativeHandle, useRef, useState } from "react";
 import { Button } from "@/components/ui/button";
 import { Input } from "@/components/ui/input";
+import {
+  Popover,
+  PopoverContent,
+  PopoverTrigger,
+} from "@/components/ui/popover";
 import { useDoBookmarkSearch } from "@/lib/hooks/bookmark-search";
+import { useSearchHistory } from "@/lib/hooks/useSearchHistory";
 import { useTranslation } from "@/lib/i18n/client";
 import { cn } from "@/lib/utils";
 import { SearchIcon } from "lucide-react";
 
 import { EditListModal } from "../lists/EditListModal";
 import QueryExplainerTooltip from "./QueryExplainerTooltip";
+import { SearchAutocomplete } from "./SearchAutocomplete";
 
 function useFocusSearchOnKeyPress(
   inputRef: React.RefObject<HTMLInputElement>,
   onChange: (e: React.ChangeEvent<HTMLInputElement>) => void,
+  setPopoverOpen: React.Dispatch<React.SetStateAction<boolean>>,
 ) {
   useEffect(() => {
     function handleKeyPress(e: KeyboardEvent) {
@@ -26,6 +34,7 @@ function useFocusSearchOnKeyPress(
         // Move the cursor to the end of the input field, so you can continue typing
         const length = inputRef.current.value.length;
         inputRef.current.setSelectionRange(length, length);
+        setPopoverOpen(true);
       }
       if (
         e.code === "Escape" &&
@@ -53,18 +62,35 @@ const SearchInput = React.forwardRef<
   React.HTMLAttributes<HTMLInputElement> & { loading?: boolean }
 >(({ className, ...props }, ref) => {
   const { t } = useTranslation();
-  const { debounceSearch, searchQuery, parsedSearchQuery, isInSearchPage } =
-    useDoBookmarkSearch();
+  const {
+    debounceSearch,
+    searchQuery,
+    doSearch,
+    parsedSearchQuery,
+    isInSearchPage,
+  } = useDoBookmarkSearch();
+  const { addTerm } = useSearchHistory();
 
   const [value, setValue] = React.useState(searchQuery);
+  const [popoverOpen, setPopoverOpen] = useState(false);
 
   const inputRef = useRef<HTMLInputElement>(null);
+  const popoverRef = useRef(false);
+
   const onChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     setValue(e.target.value);
     debounceSearch(e.target.value);
   };
 
-  useFocusSearchOnKeyPress(inputRef, onChange);
+  const handleHistorySelect = (term: string) => {
+    setValue(term);
+    doSearch(term);
+    addTerm(term);
+    setPopoverOpen(false);
+    inputRef.current?.blur();
+  };
+
+  useFocusSearchOnKeyPress(inputRef, onChange, setPopoverOpen);
   useImperativeHandle(ref, () => inputRef.current!);
   const [newNestedListModalOpen, setNewNestedListModalOpen] = useState(false);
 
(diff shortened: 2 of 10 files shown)


xuatz:
(nit) @lexafaxine sorry, can you update your description to something like
```
closes #1541
```
so that this PR is [autolinked](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) to the issue? ^^

apps/web/components/dashboard/search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,60 @@
+import React from "react";
```
xuatz:
```suggestion
```
not necessary

apps/web/components/dashboard/search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,60 @@
...
+              onSelect={() => onSelect(term)}
+              className="cursor-pointer"
+            >
+              <History className="mr-2 h-4 w-4" />
+              <span>{term}</span>
+            </CommandItem>
+          ))}
+        </CommandGroup>
+      </CommandList>
+    </Command>
```
xuatz: suggestion(ux): might be unrelated to this feature specifically, but I think there is a ux improvement that can be done

1. after firing the search query upon typing, we should dismiss the "Command" dialog
	2. or otherwise at least it should not show "No matching recent searches"
2. we can also maybe consider to dismiss the dialog if the user press "enter" button
	3. seems like an intuitive thing to do, seeing that the dialog doesn't go away

search ux improvement

lexafaxine: Hi @xuatz Thank u for the suggestions! They are really good points. You're right that showing "no results" when there is no result is not necessary in this situation. And regarding the close of the popover, I am considering UX like this:

1. When there are search history that match user's typing, show the history. Close popover if there is no match.
2. After user select one history item the popover will close. (Current behavior)
3. No matter if there is matched history, close the popover if user press Enter, and in the same time blur the input and save the term to history.

Let me know if this sounds OK to you :)

xuatz: yeah sounds good to me 👍

apps/web/components/dashboard/search/SearchInput.tsx
```diff
@@ -53,18 +62,51 @@ const SearchInput = React.forwardRef<
...
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
+    if (e.key === "Enter") {
+      e.preventDefault();
+      if (value) {
+        addTerm(value);
+      }
+      setIsPopoverOpen(false);
+      inputRef.current?.blur();
+    }
+  };
```
xuatz: suggestion(ux): when testing the feature, i found myself instinctively trying to press down on the inputbox when there are search history options, in an attempt to navigate to those items. I think this would be a UX improvement

image

apps/web/components/dashboard/search/SearchAutocomplete.tsx
```diff
@@ -0,0 +1,51 @@
...
+    ? history.filter((term) =>
+        term.toLowerCase().includes(searchQuery.toLowerCase()),
+      )
+    : history;
+
+  return (
+    <Command shouldFilter={false} className={cn(className)}>
+      <CommandList>
+        <CommandEmpty>No matching recent searches.</CommandEmpty>
+        <CommandGroup heading="Recent Searches">
```
xuatz: I think these texts should be... what do you call this... `t("label")`'ed? lol

apps/web/components/dashboard/search/SearchInput.tsx
```diff
@@ -53,18 +62,51 @@ const SearchInput = React.forwardRef<
...
+  const isPopoverVisible = isPopoverOpen && suggestions.length > 0;
+  const handleHistorySelect = (term: string) => {
+    setValue(term);
+    doSearch(term);
+    addTerm(term);
+    setIsPopoverOpen(false);
+    inputRef.current?.blur();
+  };
+
+  const handleKeyDown = (e: React.KeyboardEvent<HTMLInputElement>) => {
```
xuatz:
```suggestion
  const handleKeyDown = (e: KeyboardEvent<HTMLInputElement>) => {
```
just nitpicks ^^

apps/web/components/dashboard/search/SearchInput.tsx
lexafaxine: Thank you for the suggestion! U r right so I tried changing to use the CommandInput which already has good UX out of box :) Therefore I requested a review again, please help review when available! Thanks🙏

apps/web/components/dashboard/search/SearchAutocomplete.tsx
lexafaxine: i18n added ^^
 https://github.com/karakeep-app/karakeep/pull/1627/commits/b33cdb097560871ccb74fbb033c99964ba0e0193

apps/web/components/dashboard/search/SearchInput.tsx
lexafaxine: I guess cmdk has a restriction on keyboard event's type, if only using KeyboardEvent there will be type error :(

apps/web/components/dashboard/search/SearchInput.tsx
xuatz: the ux is feeling great 💪

xuatz (approved):
LGTMv2 👍

apps/web/lib/hooks/bookmark-search.ts
```diff
@@ -45,7 +45,7 @@ export function useDoBookmarkSearch() {
     }
     const id = setTimeout(() => {
       doSearch(val);
-    }, 10);
+    }, 200);
```
MohamedBassem: Any reason why we increased the debounce? It makes the search feels slower. If we're adding the value onBlur, then the debounce value probably doesn't make a difference?

apps/web/lib/searchHistory.tsx
```diff
@@ -0,0 +1,37 @@
+const BOOKMARK_SEARCH_HISTORY_KEY = "karakeep_search_history";
+const MAX_HISTORY_ITEMS = 5;
+
+export function getSearchHistory(): string[] {
+  try {
+    const rawHistory = localStorage.getItem(BOOKMARK_SEARCH_HISTORY_KEY);
```
MohamedBassem: If we're keeping stuff in the user's localstorage, we'll probably need to clean it up otherwise it will leak between accounts on the same device. Which makes me wonder if it might be a good idea to store it in the database instead? If we're storing them in the database, they'll get synced across device, and it'll allow you to use mostly the same logic between the mobile and web app.

We don't need to address the localstorage idea in this PR, but if you agree, we can followup on it with a separate PR.

apps/web/lib/searchHistory.tsx
```diff
@@ -0,0 +1,37 @@
...
+    if (rawHistory) {
+      const parsed = JSON.parse(rawHistory) as string[];
+      if (
+        Array.isArray(parsed) &&
+        parsed.every((item) => typeof item === "string")
+      ) {
+        return parsed;
+      }
```
MohamedBassem: Might be clear to use a simple zod schema here instead?

apps/web/lib/searchHistory.tsx
MohamedBassem, line 1: The logic in this file seems exactly the same as the one in the mobile version (except for the storage utilities). It seems to me that we can move the hooks to the `shared-react` package and pass in the getter and setter as inputs to the hook? This will allow you to share most of the code?

apps/web/lib/searchHistory.tsx
```diff
@@ -0,0 +1,37 @@
+const BOOKMARK_SEARCH_HISTORY_KEY = "karakeep_search_history";
+const MAX_HISTORY_ITEMS = 5;
```
MohamedBassem: How about we store something like 50, but by default render last 5 and only show the others as the user types?

apps/web/components/dashboard/search/SearchInput.tsx
```diff
@@ -26,54 +47,120 @@ function useFocusSearchOnKeyPress(
...
+      addTerm(term);
+      setIsPopoverOpen(false);
+      inputRef.current?.blur();
+    },
+    [doSearch, addTerm],
+  );
+
+  const handleCommandKeyDown = useCallback(
+    (e: React.KeyboardEvent) => {
+      if (e.key === "Enter") {
```
MohamedBassem: Something is wrong with the key down handler. Pressing enter correctly goes to the selected item when the search history first show up. But if you attempt to use `Enter` on any filtered result, it doesn't work.

MohamedBassem (changes requested):
I tired it and it feels great! Left a bunch of comments.

apps/web/components/dashboard/search/SearchInput.tsx
lexafaxine: good catch! there was actually some issue there 🙇 fixed in https://github.com/karakeep-app/karakeep/pull/1627/commits/b5e0acfc843b824be8591b7131456add8e03450e

apps/web/lib/searchHistory.tsx
lexafaxine: changed in https://github.com/karakeep-app/karakeep/pull/1627/commits/4e214bfd44bd209ba2d4f8b62afed3398b0c5bf9

apps/web/lib/searchHistory.tsx
lexafaxine: moved in https://github.com/karakeep-app/karakeep/pull/1627/commits/4e214bfd44bd209ba2d4f8b62afed3398b0c5bf9

apps/web/lib/searchHistory.tsx
lexafaxine: True, fixed in https://github.com/karakeep-app/karakeep/pull/1627/commits/6f83a208f4b1ca7dea32dd37d919e4a584f72f90

apps/web/lib/hooks/bookmark-search.ts
lexafaxine: You're right, that change left because I tried to save search during user's typing, reverted already in https://github.com/karakeep-app/karakeep/pull/1627/commits/6f83a208f4b1ca7dea32dd37d919e4a584f72f90

apps/web/lib/searchHistory.tsx
lexafaxine: Yeah I thought about it these days, and I think it's a good idea to implement it in database so we have sync across device and prevent leak across accounts on same device. And I also realized actually there are a lot of apps are doing something like this XD. 
I would like to implement it as a separate PR as improvement :) Thank you for the idea!

MohamedBassem:
@lexafaxine This is such a great addition, thanks a lot for all the effort you put in this PR and addressing all of our comments. Sorry it took long to merge!

PR #1628, opened 2025-06-17
chore: add start-dev.sh for laziness

xuatz:
## Summary 

* add simple script to start local dev environment

## Details

Actually, i would prefer doing `docker compose up -d`, but it's not working perfectly at the moment. Even when it's perfect, there might be people who prefer not to use docker for dev, so maybe we can keep this even after that.

But I'm not sure if we want this or not yet, but if this is welcomed, I'll update the docs as well to mention this.

### DIFF
--- a/start-dev.sh
+++ b/start-dev.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# Function to check if a command exists
+command_exists() {
+    command -v "$1" >/dev/null 2>&1
+}
+
+# Function to check if a port is in use
+port_in_use() {
+    lsof -i :"$1" >/dev/null 2>&1
+}
+
+# Check if Docker is installed
+if ! command_exists docker; then
+    echo "Error: Docker is not installed. Please install Docker first."
+    exit 1
+fi
+
+# Check if pnpm is installed
+if ! command_exists pnpm; then
+    echo "Error: pnpm is not installed. Please install pnpm first."
+    exit 1
+fi
+
+# Start Meilisearch if not already running
+if ! port_in_use 7700; then
+    echo "Starting Meilisearch..."
+    docker run -d -p 7700:7700 --name karakeep-meilisearch getmeili/meilisearch:v1.13.3
+else
+    echo "Meilisearch is already running on port 7700"
+fi
+
+# Install dependencies if node_modules doesn't exist
+if [ ! -d "node_modules" ]; then
+    echo "Installing dependencies..."
+    pnpm install
+fi
+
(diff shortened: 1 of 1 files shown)


xuatz:
@irobot 

> Shouldn't there also be a check to ensure the Chrome container is up and running?

mmm I haven't had to test anything that required it so i haven't thought about it lol.
i think running the current 3 services represents the bare minimum, so maybe we can keep it as default, and add additional flags to run "everything else" or "full", which tentatively includes the chrome container. What do you think about that?

> Personally I would start workers and web in separate terminals. This way the logs remain distinct and easier to navigate. Would there be a way to somehow achieve the same here? OTOH I could see some value in them being intermingled, so probably not that important.

If i need to dig deeper into the logs, I would run the separately, but usually I just want to start the dev environment to run a PR branch locally.

> Have you tried [docker-compose.dev](https://github.com/karakeep-app/karakeep/blob/main/docker/docker-compose.dev.yml)? If so, I briefly looked at [Dockerfile.dev](https://github.com/karakeep-app/karakeep/blob/main/docker/Dockerfile.dev) and it seems like it's missing the yt-dlp dependency. Could that be (part of) the problem you were having?

I was trying to get the volumes setup "nicely" and also getting the relative paths to the `.env` aligned, but i run into some problem when trying to do the db migration. I can probably find out what my actual problem is with some time, but since there is an open [PR](https://github.com/karakeep-app/karakeep/pull/584) to fix/improve it, I decided I'll just spend my time elsewhere instead

xuatz:
> 
> > > Shouldn't there also be a check to ensure the Chrome container is up and running?
> > 
> > mmm I haven't had to test anything that required it so i haven't thought about it lol. i think running the current 3 services represents the bare minimum, so maybe we can keep it as default, and add additional flags to run "everything else" or "full", which tentatively includes the chrome container. What do you think about that?
> 
> @xuatz Here's my understanding so far:
> * Ingestion - chrome is essential in the process of saving hoarded pages, e.g. reader view, screenshot capture. meilisearch has no role in any of that.
> * Viewing - chrome is not relevant for viewing already stored links, while meilisearch is.
> 
> In summary, depending on what you're working on, you might need only one or the other, or both. If we had to choose only one, it would be difficult for me to pick which one would be more useful to have as a default. Otherwise, for consistency it seems appropriate to have both.

@irobot 
let's have both by default then ^^

irobot (approved):
Awesome! Thank you!

MohamedBassem:
This is pretty cool, thanks @xuatz!

PR #1633, opened 2025-06-19
Updated chrome image / alpine-chrome (fork)

starapo7348:
I have switched to another image which is regular updated. The security is also improved in a more recent version.

image: ghcr.io/zoobdude/alpine-chrome:latest

https://github.com/Zoobdude/alpine-chrome

alpine-chrome (fork)

Chrome running in headless mode in a tiny Alpine image

Thank you to the original alpine-chrome team for all the heavy lifting! This repository simply continues regular builds of the original project (with more recent Chrome versions) as the original is no longer maintained.

It would be nice if you accept the change.

Thank you.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       # If you want to mount a custom directory, change the volume mapping above instead.
       DATA_DIR: /data # DON'T CHANGE THIS
   chrome:
-    image: gcr.io/zenika-hub/alpine-chrome:123
+    image: ghcr.io/zoobdude/alpine-chrome:latest
     restart: unless-stopped
     command:
       - --no-sandbox


docker/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       # If you want to mount a custom directory, change the volume mapping above instead.
       DATA_DIR: /data # DON'T CHANGE THIS
   chrome:
-    image: gcr.io/zenika-hub/alpine-chrome:123
+    image: ghcr.io/zoobdude/alpine-chrome:latest
```
xuatz, line 23:
```suggestion
    image: ghcr.io/zoobdude/alpine-chrome:100
```
haven decide if we can/want to change to this image, but if we were, probably want to pin the version as well?

Eragos: `:100` did't work for me. Get a `missing Manifest error`. Haven't check if it exists. I'm on `latest` actually for more testing.

Eragos: `:136` from yesterday is working → [Recent tagged image versions](https://github.com/zoobdude/alpine-chrome/pkgs/container/alpine-chrome/versions?filters%5Bversion_type%5D=tagged)

xuatz: i picked 100 because that's what the readme says. i've been scammed lol
image

thanks for testing it out!

xuatz: but, who is zoobdude and why can we trust zoobdude.

I hope it doesn't sound like I'm like "who the hell is zoobdude", more like why should we switch over and the justification for it.

fun fact, gcr.io/zenika-hub/alpine-chrome:124 was literally published yesterday 🤪 
https://hub.docker.com/r/zenika/alpine-chrome/tags
image

xuatz (changes requested):
not sure if we should change image when the original image is still maintained

docker/docker-compose.yml
Eragos: That’s the point. It should be trustful and actually I don’t see that.

MohamedBassem:
As mentioned in the comment, there's no way we'd accept a random image from a repo with zero stars.

PR #1636, opened 2025-06-20
feat: Add cookie support for browser page access

liCells:
Cookies are now managed through a JSON file specified by the BROWSER_COOKIE_PATH configuration.

### Changes

- Introduced BROWSER_COOKIE_PATH configuration to specify the path to the cookies JSON file.
- Added cookie functionality, JSON file parsing, and path validation.

### Example Cookies JSON File
The cookies JSON file should be an array of cookie objects with the following properties:
```json
[
  {
    "name": "auth_token",
    "value": "xxx",
    "domain": "xxx.com",
    "path": "/"
  }
]
```
Additional properties like secure, httpOnly, or expires can be included to configure cookie behavior as needed.
Specific reference can be made to [CookieParam](https://pptr.dev/api/puppeteer.cookieparam).

### Misc

For example adding the auth_token for x/twitter gives access to some of the protected tweets.

Of course, it can also be applied to other similar websites.

### DIFF
--- a/apps/workers/workers/crawlerWorker.ts
+++ b/apps/workers/workers/crawlerWorker.ts
@@ -19,7 +19,6 @@ import metascraperDescription from "metascraper-description";
 import metascraperImage from "metascraper-image";
 import metascraperLogo from "metascraper-logo-favicon";
 import metascraperPublisher from "metascraper-publisher";
-import metascraperReadability from "metascraper-readability";
 import metascraperTitle from "metascraper-title";
 import metascraperTwitter from "metascraper-twitter";
 import metascraperUrl from "metascraper-url";
@@ -72,7 +71,6 @@ const metascraperParser = metascraper([
   }),
   metascraperAmazon(),
   metascraperReddit(),
-  metascraperReadability(),
   metascraperAuthor(),
   metascraperPublisher(),
   metascraperTitle(),


apps/workers/workers/crawlerWorker.ts
```diff
@@ -212,10 +213,30 @@ export class CrawlerWorker {
...
+    logger.info("[Crawler] No cookie path provided, will not set cookies");
+    return [];
+  }
+
+  if (!existsSync(cookiePath)) {
+    logger.warn(`[Crawler] Cookie file does not exist at path: ${cookiePath}`);
+    return [];
+  }
+  try {
+    return JSON.parse(readFileSync(cookiePath, "utf8")) as CookieParam[];
```
MohamedBassem: Can you please use the readFile api instead?

MohamedBassem: Can you provide a zod schema and parse the file with it to ensure that the file is in the expected schema?

packages/shared/config.ts
```diff
@@ -54,6 +54,7 @@ const allEnv = z.object({
   BROWSER_WEB_URL: z.string().optional(),
   BROWSER_WEBSOCKET_URL: z.string().optional(),
   BROWSER_CONNECT_ONDEMAND: stringBool("false"),
+  BROWSER_COOKIE_PATH: z.string().optional(),
```
MohamedBassem: Can you please update the documentation in `docs/docs/03-configuration.md` to mention this new variable? Also mention to people the expected schema?

PR #1642, opened 2025-06-20
deps: switch from eslint to oxlint

xuatz:
This is just an early preview, with only apps/browser-extension migrated as an example to confirm if we can/want to proceed with the migration this way.

WIP not ready to review yet, drafting PR description


MohamedBassem:
@xuatz Yeah, this looks good to me, I think we can proceed with the rest of the packages. Did you have to turn off any rules btw? Or was this all spitted up by the config migration code?

xuatz:
> @xuatz Yeah, this looks good to me, I think we can proceed with the rest of the packages. Did you have to turn off any rules btw? Or was this all spitted up by the config migration code?

@MohamedBassem 
actually the oxlint migration script reported and removed tons of unsupported rules. I saved them on a separate [commit](https://github.com/xuatz/karakeep/commit/966fe64553c9a231c2af9792625cba527dcab5e9).

image

I was gonna write a little report about that, but i got lazy halfway through yesterday, and I've been taking break from it since lol. I'm still not sure what this means for karakeep, if we can give up not having these rules or, attempt to implement some of them by hand as custom oxlint rules 🤔

tooling/oxlint/oxlint-base.json
xuatz, line 1: @MohamedBassem while im here, another annoying thing is that, "reuseable"/extendable configs only supports `plugins`,`rules` and `overrides` for now.
* https://oxc.rs/docs/guide/usage/linter/nested-config.html#extending-configuration-files

it means that we need duplicate fields like, `globals`, and `ignorePatterns` on every subrepo's `.oxlintrc.json`. a little annoying, but i personally can accept it, but what do you think?

MohamedBassem:
@xuatz Regarding the missing rules, it's fine, I'm not too attached to them. I'm fine dropping whatever they're currently not supporting. As for the subrepo problem, I'm ok with what you suggested as well.

package.json
```diff
@@ -16,18 +16,17 @@
...
     "format:fix": "turbo --no-daemon format --continue -- --write --cache-location node_modules/.cache/.prettiercache",
-    "lint": "turbo --no-daemon lint --continue -- --cache-location node_modules/.cache/.eslintcache",
-    "lint:fix": "turbo --no-daemon lint --continue -- --fix --cache-location node_modules/.cache/.eslintcache",
+    "lint": "turbo --no-daemon oxlint --continue --",
+    "lint:fix": "turbo --no-daemon oxlint --continue -- --fix",
     "typecheck": "turbo --no-daemon typecheck"
   },
   "devDependencies": {
     "@karakeep/prettier-config": "workspace:^0.1.0",
-    "@tanstack/eslint-plugin-query": "^5.20.1",
```
xuatz, line 25: This was removed without replacement 😭 
@tanstack/eslint-plugin-query not supported yet
* https://github.com/oxc-project/oxc/issues/481

package.json
```diff
@@ -16,18 +16,17 @@
     "prepare": "husky",
     "format": "turbo --no-daemon format --continue -- --cache-location node_modules/.cache/.prettiercache",
     "format:fix": "turbo --no-daemon format --continue -- --write --cache-location node_modules/.cache/.prettiercache",
-    "lint": "turbo --no-daemon lint --continue -- --cache-location node_modules/.cache/.eslintcache",
-    "lint:fix": "turbo --no-daemon lint --continue -- --fix --cache-location node_modules/.cache/.eslintcache",
+    "lint": "turbo --no-daemon oxlint --continue --",
+    "lint:fix": "turbo --no-daemon oxlint --continue -- --fix",
```
xuatz: not super sure if i'm doing it right with turbopack, not that familiar with it. let me know if something needs to be fixed/improved

apps/browser-extension/.oxlintrc.json
```diff
@@ -0,0 +1,744 @@
...
+    "correctness": "warn"
+  },
+  "env": {
+    "builtin": true,
+    "commonjs": true,
+    "browser": true,
+    "es2022": true,
+    "node": true
+  },
+  "globals": {
```
MohamedBassem, line 17: Those to me seem like defaults? Do we need to actually list them here and in other files? I assume if we remove this section it'll just work?

apps/web/app/dashboard/cleanups/page.tsx
```diff
@@ -4,6 +4,7 @@ import { useTranslation } from "@/lib/i18n/server";
 import { Paintbrush, Tags } from "lucide-react";
 
 export default async function Cleanups() {
+  // oxlint-disable-next-line rules-of-hooks
```
MohamedBassem, line 7: It's not clear to me how this is violating the rules of hooks?

package.json
MohamedBassem: I think it's fine to keep the rule name in turbo as `lint` and revert the change in this line

package.json
MohamedBassem: It's unfortunate, but the speed wins I assume are worth it.

tooling/tailwind/package.json
```diff
@@ -11,8 +11,10 @@
   "license": "MIT",
   "scripts": {
     "clean": "rm -rf .turbo node_modules",
-    "format": "prettier --check . --ignore-path ../../.gitignore",
-    "lint": "eslint .",
+    "format": "prettier .",
```
MohamedBassem, line 14: Is the change of removing the gitignore here intended?

turbo.json
```diff
@@ -32,12 +32,9 @@
       ],
       "outputLogs": "new-only"
     },
-    "lint": {
+    "oxlint": {
```
MohamedBassem: As I mentioned above, I don't think you need to change this.

tooling/tailwind/package.json
xuatz: without removing it, it seems to ignore the root .prettierignore.
i saw that most other `"format"` were a lot more simpler so i assumed this was an old pattern that can be updated.

if this was very intentional, i can revert this and see if there is any additional follow up required.

package.json
xuatz: reverted 👍

apps/browser-extension/.oxlintrc.json
xuatz: I removed them and there were no output changes 👍 
applied the suggestion.

apps/web/app/dashboard/cleanups/page.tsx
xuatz: yeah the code works, but oxlint seem to (maybe mistakenly) flag this as a violation of rules-of-hooks. 

but I don't think we want to turn it off just because of this incident, hence I went with inline skip instead.

MohamedBassem: Ah I think the issue is that it's technically not a hook because this is a server component. It's fine, we can keep it for now.

Issue #1572, opened 2025-06-09
Configuring MEILI_MASTER_KEY for karakeep and meilisearch results in 403 authentication errors


kdwils:
### Describe the Bug

Installation configuration was done based on https://docs.karakeep.app/configuration

On a clean install of karakeep and meilisearch in a kubernetes environment on a fresh pvc, configuring the MEILI_MASTER_KEY for both karakeep and meilisearch results in 403s when loading the search view in the karakeep app. 

Choosing to not configure the master key results in no errors 

### Steps to Reproduce

1. Generate a master key using `openssl rand -base64 36`
2. kubectl apply -f example-manifest.yaml with master key secret
```yaml
# example-manifest.yaml
---
apiVersion: v1
kind: Namespace
metadata:
  name: hoarder
---
apiVersion: v1
data:
  BROWSER_WEB_URL: http://chrome:9222
  DATA_DIR: /data
  DISABLE_SIGNUPS: "false"
  MEILI_ADDR: http://meilisearch:7700
  NEXTAUTH_URL: http://fake
kind: ConfigMap
metadata:
  name: hoarder
---
apiVersion: v1
data:
  MEILI_NO_ANALYTICS: "true"
kind: ConfigMap
metadata:
  name: meilisearch
---
apiVersion: v1
kind: Service
metadata:
  name: chrome
spec:
  ports:
    - port: 9222
      protocol: TCP
      targetPort: 9222
  selector:
    app: chrome
  type: ClusterIP
---
apiVersion: v1
kind: Service
metadata:
  labels:
    app: hoarder
  name: hoarder
spec:
  ports:
    - name: http
      port: 3000
      targetPort: 3000
  selector:
    app: hoarder
---
apiVersion: v1
kind: Service
metadata:
  name: meilisearch
spec:
  ports:
    - port: 7700
      protocol: TCP
      targetPort: 7700
  selector:
    app: meilisearch
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: data
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 500Mi
---
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  name: meilisearch
spec:
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 500Mi
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: chrome
spec:
  replicas: 1
  selector:
    matchLabels:
[...]

PR #1651, opened 2025-06-21
Feature/meili master key updated example

kdwils:
Characters such as `+` seem to cause authentication issues with meilisearch if they are part of the `MEILI_MASTER_KEY` environment variable. This PR is to update the example command to generate a master key to only use alphanumeric characters.

Related issue: https://github.com/karakeep-app/karakeep/issues/1572

Screenshot of the updated table
table


MohamedBassem:
Thank you!

PR #1664, opened 2025-06-23
feat: Add cookie support for browser page access

liCells:
Cookies are now managed through a JSON file specified by the BROWSER_COOKIE_PATH configuration.

### Changes

- Added type definition for BROWSER_COOKIE_PATH in packages/shared/config.ts.
- Implemented asynchronous cookie loading in apps/workers/workers/crawlerWorker.ts using fs.promises.readFile to read the JSON file and Playwright's context.addCookies to load cookies.
- Updated docs/docs/03-configuration.md to include a description of BROWSER_COOKIE_PATH and an example JSON file format.

### Misc

For example adding the auth_token for x/twitter gives access to some of the protected tweets.

Of course, it can also be applied to other similar websites.

[Source PR](https://github.com/karakeep-app/karakeep/pull/1636)


MohamedBassem:
Extremely sorry for the long delay in reviewing this PR. I fixed the merge conflicts and merged the PR in https://github.com/karakeep-app/karakeep/commit/c57fd5137cc29870667777a371a4d1fcdf69436b

maelp:
Really cool! So my understanding is that we would need to have some kind of local script that we regularly run on our laptop to copy that JSON file to the backend somewhere, so it can be loaded?

When we change it, do we need to "restart" Karakeep so it picks up the latest version, or is it "refetched" each time Karakeep does a crawl (would be even better!)

It would be nice to share with the community a smal typescript / python script someone could run to create such a JSON file automatically from its own browser 

eg

```sh
./create_shared_cookies.sh --site nytimes.com --site bluesky.com --output cookies.json
scp cookies.json my-server:/karakeep/data
```

liCells:
> Really cool! So my understanding is that we would need to have some kind of local script that we regularly run on our laptop to copy that JSON file to the backend somewhere, so it can be loaded?
> 
> When we change it, do we need to "restart" Karakeep so it picks up the latest version, or is it "refetched" each time Karakeep does a crawl (would be even better!)
> 
> It would be nice to share with the community a smal typescript / python script someone could run to create such a JSON file automatically from its own browser
> 
> eg
> 
> ```shell
> ./create_shared_cookies.sh --site nytimes.com --site bluesky.com --output cookies.json
> scp cookies.json my-server:/karakeep/data
> ```

If you need to export cookies from a browser, you can use a browser extension like Cookie-Editor. The exported JSON format is nearly identical to the structure required for this functionality, requiring only minor adjustments.

Regarding the issue of reloading the cookie file, I recall from the previous development process that the loading occurs during project startup, meaning that any modifications to the cookie file would require a service restart. However, you can test this in two different scenarios to confirm the behavior.

PR #1682, opened 2025-06-27
fix: switch import / export icons

HarryPeach:
In my opinion, the icons for import and export are reversed from what a user would expect them to be.

image

All this PR does is switch the "Download" and "Upload" symbols in the Import/Export settings menu!

| Current  | My Changes |
| ------------- | ------------- |
| image  | image  |

(A world of difference, I know)

### DIFF
--- a/apps/web/components/settings/ImportExport.tsx
+++ b/apps/web/components/settings/ImportExport.tsx
@@ -55,7 +55,7 @@ function ImportCard({
     <Card className="transition-all hover:shadow-md">
       <CardContent className="flex items-center gap-3 p-4">
         <div className="rounded-full bg-primary/10 p-2">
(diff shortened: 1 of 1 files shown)


xuatz:
Hmmm despite the lucide-icon name, I kinda agree that the image somehow seem to convey the opposite meaning. I agree with the change, but I think we should also make it consistent and apply it across the whole application for other instances (I briefly searched and I think there are only just a handful other instances)

Eragos:
Hey!

From the view of a karakeep user, it's IMOH better to switch the icons. It's better than creating confusions ;-)

- lucide-upload
- lucide-download

Both namings makes sense from a different point of view.

Best, Michael

PR #1706, opened 2025-07-05
fix: Ensure email is handled in a case-insensitive manner across auth…

haroondilshad:
…entication and user creation

## Fix: Case-Insensitive Email Authentication

**Problem:** Users couldn't login with different email cases (e.g., `user@example.com` vs `USER@EXAMPLE.COM`).

**Solution:** Normalize emails to lowercase across all platforms.

### Changes
- **Mobile App**: Email input normalizes to lowercase on typing and login
- **Web App**: SignIn/SignUp forms normalize email to lowercase  
- **Server Auth**: Database queries use lowercase emails
- **User Creation**: Store emails as lowercase in database
- **Testing**: Added test for email case insensitive uniqueness

### Impact
- Users can login with any email case variation
- Prevents duplicate accounts with different email cases
- Consistent behavior across mobile, web, and server

**Tests:** All 137 tests pass, no regressions.

### DIFF
--- a/apps/mobile/app/signin.tsx
+++ b/apps/mobile/app/signin.tsx
@@ -100,7 +100,7 @@ export default function Signin() {
     if (loginType === LoginType.Password) {
       const randStr = (Math.random() + 1).toString(36).substring(5);
       login({
-        email: formState.email,
+        email: formState.email.toLowerCase(),
         password: formState.password,
         keyName: `Mobile App: (${randStr})`,
       });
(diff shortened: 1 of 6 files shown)


MohamedBassem:
Sorry for the delay on getting back to this PR but unfortunately this change is not safe as there might be some deployments already with duplicate emails (that differ in case sensitivity). We can solve it by just naive case-insensitive checks. And a migration also will not work because it'll violate the uniqueness constraint. I think we're kinda stuck with case sensitivity in email unfortunately.

haroondilshad:
Bummer. Thanks for the review. Maybe if in the future, it's discovered that there aren't many cases in the wild of same email with differing case sensitivity, we can rethink this one. Cheers!

Issue #1418, opened 2025-05-16
Show item count and search result count


benwaffle:
### Describe the feature you'd like

I'd like to see on the home page the total number of items i have saved, and when searching i'd like to see the number of results found.

### Describe the benefits this would bring to existing Karakeep users

I guess it's nice to see it in order to motivate me to go through my backlog and lower it.

### Can the goal of this request already be achieved via other means?

Not that I can see.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

Eragos:
Had some "kind of" self-made with the rule engine:

![Image](https://github.com/user-attachments/assets/22894282-6b1d-4f51-80b6-3e55097be8cd)

Best Michael

officiallymarky:
I'd love a similar bubble on "Home" that is like what you see with lists to show how many items in your "inbox" but excluding Archived.

rodsnts:
@MohamedBassem @xuatz Can you assign me this one please?

xuatz:
> @MohamedBassem @xuatz Can you assign me this one please?

@rodsnts it's now yours! Go crazy!

PR #1738, opened 2025-07-15
feat: Add bookmarks/tags counter to sidebar and search results (#1418)

rodsnts:


MohamedBassem (changes requested):
Thanks for the PR, overall, the change looks good to me, but I have a couple of thoughts:

1. Let's try to keep the colors the same as that of the lists, and I also don't feel like the motion is very visible to be justified.
2. I think it looks a bit odd when you have three digit numbers.
3. I think using `user.stats` just for those three numbers is very expensive as the stats endpoint does a lot of computations. Maybe we can pass an optional input to selectively select which stat we want returned?


Screenshot 2025-07-17 at 9  58 39@2x

rodsnts:
Thanks for the feedback, yeah I agree with your points I will make the changes later Today

PR #1740, opened 2025-07-16
feat: Force human tags before AI tags

### LINKED ISSUE #1630: FR: human tags should appear first in the bookmark interface

The tags appear in the creation order I believe, but I think the human tags should always appear first.

For example, I want to see directly if a given bookmark is short to read, using [the Karakeep-Time-Tagger script](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts/karakeep-time-tagger) I made (using the [community scripts](https://github.com/thiswillbeyourgithub/karakeep_python_api/tree/main/community_scripts)). But the `0-5m` is only visible if I click on the bookmark.

Image

Image

Better organization

No

### Additional context

_No response_

haappi:
We're more special.

er that didn't work- closes #1630

### DIFF
--- a/apps/browser-extension/src/components/TagList.tsx
+++ b/apps/browser-extension/src/components/TagList.tsx
@@ -14,16 +14,20 @@ export default function TagList({ bookmarkId }: { bookmarkId: string }) {
       {bookmark.tags.length === 0 && !isBookmarkStillTagging(bookmark) && (
         <Badge variant="secondary">No tags</Badge>
       )}
-      {bookmark.tags.map((tag) => (
-        <Badge
-          key={tag.id}
-          className={
-            tag.attachedBy == "ai" ? "bg-purple-500 text-white" : undefined
-          }
-        >
-          {tag.name}
-        </Badge>
-      ))}
+      {[...bookmark.tags]
+        .sort((a, b) =>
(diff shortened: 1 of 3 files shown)


xuatz:
@haappi I think to make sure your efforts are not in vain, you should perhaps apply this change after we address https://github.com/karakeep-app/karakeep/issues/1404.

Because, we need to decide how we plan to progressively return more tags. Infinite scroll? Manually clicking to fetch more tags? And we might need a different UX for `/dashboard/tags` and `Edit Bookmark` dialog tags input field.

More specifically why it affects your PR, depending on the solution to #1404, the sorting might be already done on server side, maybe.

btw also, would be helpful to include screenshots or demo if you don't mind 🙇

thiswillbeyourgithub:
Hi, I'm both the author of issue #1630 and #1404 and can report that the current PR works fine. I don't see any reason to merg it as adressing #1404 can take some more time no?

In any case thank you very much @haappi

xuatz:
> Hi, I'm both the author of issue #1630 and #1404 and can report that the current PR works fine. I don't see any reason to merg it as adressing #1404 can take some more time no?
> 
> In any case thank you very much @haappi

@thiswillbeyourgithub yeah its not a blocker for me. this PR can be merged after it gets reviewed and approved by @ MohamedBassem

MohamedBassem:
Thank you!

Issue #486, opened 2024-10-05
[Feature Request][Browser Extension] Show "hoarded" status on addon icon


bwachtendorf:
Currently you cannot quickly determine if the page you are currently on is already hoarded.
I think it would be nice to show a small status indicator like a gold star or a green circle in the extension icon to show the current page is hoarded.

Small mockup:
![image](https://github.com/user-attachments/assets/4551c9eb-d594-46f9-b37e-efa4a315cb56)

MohamedBassem:
Thanks for the feature request! To make this happen, we'll have to query hoarder for every single page that you visit in your browser, which I'm kinda not that big of a fan off to be honest. I'll leave the feature request open to see if other people would want it, but I don't think we'll implement this soon to be honest :)

bwachtendorf:
Yeah you are right that might be a bit much.
Maybe as an option that has be enabled manually or in combination with #478 it would be appropriate to create a local cache, which could be checked instead of querying hoarder directly.

huyz:
A local cache would go a very long way. For me, the most common use case is: I have a lot of tabs and jump between articles all the time and I can't keep track of what I've already hoarded (sometimes I'm still not done reading the article and keep the tab open; when I finish the article, I'm not sure if I've already hoarded when I partially read it)

kamtschatka:
So it would just be a small quality of life improvement right?
Because what would be the worst that happens? The dialog would open again and hoarder would not add it a second time, because we have deduplication of links.
Am I missing something?

bwachtendorf:
The current proposed status would just be a small QoL feature in my opinion.

The scope could be widened by signaling, that another page of this domain was hoarded or the count of hoarded pages on this domain, like bitwarden does with saved credentials for a domain.

jalrok:
This would be a great option to have in the extension settings. Raindrop does this and seeing the check mark over the icon makes it very quick to tell if it's already been retrieved.

NLZ:
I would find it useful as well.

Are the concerns performance related?

MohamedBassem:
Given that multiple people are asking for it, I've marked it as approved, and I'd be happy if someone would pick it up and implement it. Whoever implements it though will need to make sure that it's optional in the extension settings.

adripo:
Thank you for considering this @MohamedBassem. It would be a helpful quality-of-life improvement, especially when working with many tabs and trying to avoid duplicate hoards.

PR #1745, opened 2025-07-17
feat: add tab bookmark count badge indicator

qixing-jk:
This PR introduces a new feature that displays a badge on the icon of the browser extension to show the number of times the current page's URL has been added as a bookmark.
 
## Main functions
**Bookmark quantity indicator:** The expansion icon now displays a symbol that contains the number of bookmarks in the active tab. If the current URL has been accurately bookmarked, the symbol is green; otherwise, it is red.
Real-time update: The count of the indicator will be automatically updated when bookmarks are added or removed for the current page.
**Configurable cache:** To optimize performance and reduce API calls, the indicator count is cached. The expiration time of the cache can be configured in the extension's options. We adopted the stale-while-revalidate (SWR) strategy to ensure the final consistency of the indicator count.
**Settings switch:** This indicator feature can be enabled or disabled in the extension's options page.
Cache management: A new menu item has been added to the right-click menu of the extension icon, allowing users to clear the cache for the current page or all pages.
**Improved URL verification:** More strict verification has been implemented, allowing bookmarks to be added only for URLs with the http and https protocols.

## Related issue
https://github.com/karakeep-app/karakeep/issues/486

## Results show
image
image

You can now easily check if the page you're viewing has already been hoarded and see the total count of similar URLs.
Badge Numbers and Colors: Their Meanings
Bookmark Count Badge System

**Smart Color Coding:**
🟢 Green: Exact URL match found
🔴 Red: No bookmarks found
🔢 Numbers: 0 = No bookmarks, 1+ = Count of bookmarks for current page

**Real-time Updates**  
- The Badge updates instantly after adding a bookmark, no need to refresh the page manually.  
- When a bookmark is removed, the Badge status synchronizes immediately as well.

### DIFF
--- a/apps/browser-extension/src/background/background.ts
+++ b/apps/browser-extension/src/background/background.ts
@@ -286,7 +286,7 @@ chrome.commands.onCommand.addListener(handleCommand);
  * @param tabId The ID of the tab to update.
  */
 export async function setBadge(badgeStatus: string | null, tabId?: number) {
-  if (!tabId || !badgeStatus) return;
+  if (!tabId) return;
 
   if (badgeStatus) {
     return await Promise.all([


apps/browser-extension/src/background/background.ts
```diff
@@ -1,124 +1,262 @@
...
+} from "../utils/settings.ts";
+import { cleanupApiClient, getApiClient } from "../utils/trpc.ts";
+import { MessageType } from "../utils/type.ts";
+import { NEW_BOOKMARK_REQUEST_KEY_NAME } from "./protocol.ts";
+
+const OPEN_KARAKEEP_ID = "open-karakeep";
+const ADD_LINK_TO_KARAKEEP_ID = "add-link";
+
+// Initialize the cache system, which creates a timer for periodic cache cleanup.
+initializeCache();
```
MohamedBassem: Is the alarm really needed? Can we just store the last time the purge was done in the in memory cache and just trigger a purge async if an hour has passed in the `onUpdated` / `onActivated` callbacks?

apps/browser-extension/src/background/background.ts
```diff
@@ -1,124 +1,262 @@
...
+
+// Listen for timer events and execute cache cleanup tasks when triggered.
+chrome.alarms.onAlarm.addListener((alarm) => {
+  if (alarm.name === "badgeCachePurgeAlarm") {
```
MohamedBassem: PURGE_ALARM_NAME?

apps/browser-extension/src/background/background.ts
```diff
@@ -1,124 +1,262 @@
...
+    return;
+  }
+  console.log("Tab activated", tabId, tabInfo);
+
+  try {
+    const cachedInfo = await getBadgeStatusSWR(tabInfo.url);
+    if (cachedInfo) {
+      const { count: cachedBadgeCount, isExisted: cachedIsExisted } =
+        cachedInfo;
+      await setBadge(cachedBadgeCount, cachedIsExisted, tabId);
```
MohamedBassem: Setting a count text is a bit confusing, can we just set a boolean indicator of whether the page existed or not? I don't think setting a count adds a value?

apps/browser-extension/src/background/background.ts
```diff
@@ -1,124 +1,262 @@
...
+  await checkAndUpdateIcon(changeInfo);
+});
+
+// Listen for REFRESH_BADGE messages from popup and update badge accordingly
+chrome.runtime.onMessage.addListener(async (msg) => {
+  if (msg && msg.type) {
+    if (
+      msg.currentTab &&
+      (msg.type === MessageType.BOOKMARK_CREATED_REFRESH_BADGE ||
+        msg.type === MessageType.BOOKMARK_DELETED_REFRESH_BADGE)
```
MohamedBassem: If we're handling them the same, do we need two different message types? Why not just `BOOKMARK_REFRESH_BADGE`?

apps/browser-extension/src/OptionsPage.tsx
```diff
@@ -55,6 +60,24 @@ export default function OptionsPage() {
       <Logo />
       <span className="text-lg">Settings</span>
       <hr />
+      <div className="flex flex-col gap-2">
+        <span className="mb-1">Show count badge</span>
+        <Button
```
MohamedBassem: Can we use a toggle instead of a button?

apps/browser-extension/src/background/background.ts
qixing-jk: image
image
This is the effect I want to achieve. The color indicates whether the links are exactly the same, and the number represents the total number of matching items. 
If possible, as in this case, present specific results

MohamedBassem:
@qixing-jk the coderabbit stuff are experimental, feel free to ignore them if they don’t make sense. This it the first PR it runs on :)

qixing-jk:
> @qixing-jk the coderabbit stuff are experimental, feel free to ignore them if they don’t make sense. This it the first PR it runs on :)

Some of the suggestions are indeed meaningful, but I have just finished my thoughts. Now I will start addressing the issues raised by these suggestions.

qixing-jk:
Sorry, after multiple rounds of SWR caching attempts, I found that the iterative process was completed only to find out that tanstack already has a ready-made swr mechanism, I should have looked more into the project package dependencies and package-specific features.

There may still be a lot of problems with the code, but I can't optimize it for the time being. Functionality is not a problem, but the logic processing of multiple state changes is a headache. I've run it through AI review, so the problems should be minimal.

MohamedBassem:
@qixing-jk thanks for all the effort you've put into that, I'll review it soon :)

MohamedBassem:
@codex review

MohamedBassem:
Sorry it took too long to merge this PR!

PR #1762, opened 2025-07-22
fix(mobile): android share intent stuck on hoarding screen

### LINKED ISSUE #513: Unreliable Item Addition via Share Menu on Android
**Description:**
When using the Android share menu to add items to Hoarder, the process is unreliable. Links are either:

- Not added at all, or
- Added multiple times.

For example:
1. I add a link to Hoarder via the share menu.
2. I see the "Hoarded" screen as expected, then exit the app.
3. Upon returning to the app, I see the same screen again, and Hoarder adds the item again, resulting in duplicate entries.

**Steps to Reproduce:**
Use the share menu outside of the Hoarder app (e.g., from a browser) to add a link.
Observe if the link is added correctly, or if it is added multiple times.
Exit and re-enter the app to check for duplicates.

**Expected Behavior:**
The shared link should be added once, with no duplicates or missing entries.

**Actual Behavior:**
Links may not be added at all or may be added multiple times.

**Device/Environment:**
Platform: Android
Version: 12
App Version: 1.6.3

xuatz:
## Summary 

- Fix race condition by moving resetShareIntent() to success/error callbacks
- Add validation to ensure share intent data exists before processing
- Upgrade expo-share-intent from 3.0.0 to 3.2.3 to fix Android timing issues
- Add Android-specific debug logging for share intent troubleshooting
- Replace "Hoarding/Hoarded" terminology with "Saving/Saved"
- Fix TypeScript errors with proper null checking and state management

Resolves #513 


xuatz:
> I'm installing the dev variant on my device and trying it out for maybe a week, to be sure that the problem is completely fixed. So far so good.

(android) So far, after first day of usage, 100% success in saving bookmarks via share intent.

edit2: after 5 days, saving bookmark via android share intent have had 100% success rate.

xuatz:
hmmm unfortunately, while the problem of launching the app via share content is more reliable about, I've noticed a couple of times when i made a new share, the app says "already saved".

it's gonna be pretty hard to reproduce in my opinion, but i guess i can make some assumptions about what happened, but not sure why that happened 🤔 

tough nut to crack.

xuatz:
It seems like if I diligently click on dismiss, the app will continue to behave properly over time. 

Maybe we can look into making sure all exit paths (go back, standby in bg, killed in bg) should already execute whatever side effects triggered by clicking dismiss. 

Or do without the dismiss and go into the app directly anyways

Or build the share intent from scratch so we don't even need to open the app (probably for the best)

MohamedBassem:
@xuatz I like the APP_VARIANT idea, would you be interested in sending this in a separate PR?

xuatz:
> @xuatz I like the APP_VARIANT idea, would you be interested in sending this in a separate PR?

@MohamedBassem sure, no problem 👌

xuatz:
> @xuatz I like the APP_VARIANT idea, would you be interested in sending this in a separate PR?

@MohamedBassem sure, no problem 👌

xuatz:
> > @xuatz I like the APP_VARIANT idea, would you be interested in sending this in a separate PR?
> 
> @MohamedBassem sure, no problem 👌

PR made, please let me know if the docs instruction is confusing and you need some clarification.
Or otherwise feel free to update the docs if you have a better idea :bowing_man: 

* https://github.com/karakeep-app/karakeep/pull/2422

PR #1780, opened 2025-07-25
fix: prometheus add karakeep prefix to metrics

CrazyWolf13:
Hi,

Thanks for the prometheus metrics.

Many of my apps all list a prefix of their app name in prometheus, karakeep does not so, if another app would have the same metrics name, I guess there may be a conflict?

So to be sure this PR adds a prefix for karakeep too.

### DIFF
--- a/packages/trpc/stats.ts
+++ b/packages/trpc/stats.ts
@@ -15,6 +15,7 @@ import {
   WebhookQueue,
 } from "@karakeep/shared/queues";
 
+// Queue metrics
 const queuePendingJobsGauge = new Gauge({
   name: "karakeep_queue_jobs",
   help: "Number of jobs in each background queue",
@@ -55,6 +56,7 @@ const queuePendingJobsGauge = new Gauge({
   },
 });
 
+// User metrics
 const totalUsersGauge = new Gauge({
   name: "karakeep_total_users",
   help: "Total number of users in the system",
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thank you!

CrazyWolf13:
@MohamedBassem 

Sorry to ping you on this PR, but I've just seen there are more stats inside of karakeep, from the grafana config linked in the issue above.

Namely the following:

```bash
scrape_series_added
scrape_samples_scraped
scrape_duration_seconds
scrape_samples_post_metric_relabeling
```

Can you point me to where this scrape bucket is configured so I can make another PR.
This seems to have not been renamed to karakeep_prefix, though they don't seem to be in the prometheus config I edited above?

MohamedBassem:
@CrazyWolf13 those are apparently automatically generated by prometheus on the server side https://prometheus.io/docs/concepts/jobs_instances/

CrazyWolf13:
@MohamedBassem aahh gotcha, thanks I was already wondering where the hell they were implemented in karakeep :D

I have updated the dashboard, 

[Karakeep-1757275022457.json](https://github.com/user-attachments/files/22198350/Karakeep-1757275022457.json)


Shall I upload the updated Board to [Grafana Dashboards](https://grafana.com/dashboards), which was originally done by @simono41 do you want to do that or does @simono41 want to do that?

MohamedBassem:
@CrazyWolf13 I'll leave that decision between you and @simono41 :)

Issue #1779, opened 2025-07-24
Add Android Share button to single bookmark


Lightjohn:
### Describe the feature you'd like

In Karakeep android for each saved link it would be nice to be able to Share them directly from here.
Instead I have to open them until I can reach Firefox and from there Share the link to another App.

<img width="720" height="1560" alt="Image" src="https://github.com/user-attachments/assets/ab644f34-c908-4f04-9f57-bb4e2d5800aa" />
<img width="720" height="1560" alt="Image" src="https://github.com/user-attachments/assets/d0bb410f-7dbc-4f80-a256-d11ad829e45f" />

### Describe the benefits this would bring to existing Karakeep users

Share button is a common feature in Android to share anything from one app to another.
It would make Sharing easier

### Can the goal of this request already be achieved via other means?

Open link in integrated browser then open in Normal browser then share from there

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

MohamedBassem:
It's planned in #1188.

Issue #1616, opened 2025-06-16
[mobile] on mobile, there should be an option to quickly share the raw content of the bookmark


xuatz:
### Describe the feature you'd like

For example, if it was an url, then, we should have a url icon to copy to clipboard.
Even if they get rendered as an image or youtube video etc, we should still have the ability to quick copy the url.

We have something like this in the web via "Copy Link"

### Describe the benefits this would bring to existing Karakeep users

It might not be the main use case of karakeep, but it allows us to quickly share links/resources across devices via karakeep.

### Can the goal of this request already be achieved via other means?

I have to create a text bookmark with backticks e.g. `www.google.com`, and then open the bookmark, and edit it, to copy the text.

In the case of youtube video. not even clicking the browser icon helps, since it opens up the native youtube app.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

MohamedBassem:
That's exactly my plan for #1188 which is already in the 0.26 milestone. Basically a new menu item called `Share`. Let's track it there.

Issue #1343, opened 2025-05-01
Share button in mobile app


chreniuc:
### Describe the feature you'd like

I've noticed that there is no button to share links from the mobile app. It would be great if there was a button to share the link.

### Describe the benefits this would bring to existing Karakeep users

I encountered a lot of scenarios where I wanted to send a link to someone and didn't know how to do it, I had to open the url on the browser and then share it from there.

### Can the goal of this request already be achieved via other means?

Open in browser and then share it

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

MohamedBassem:
thin

Issue #1054, opened 2025-02-19
add share or copy bookmark link and landscape support on android


xxjoe2:
### Describe the feature you'd like

would be great if we can share the bookmark link or copy the url of a bookmark on android app in the bookmarks list just like pocket does

also on android only portrait mode is supported. please make it supports landscape mode too

### Describe the benefits this would bring to existing Hoarder users

enable users to copy/share url and browse the url with the browser users preferred. eg copy link and browse in incognito mode with another browser, or share the link with others via whatsapp

and as a tablet user, it's quite frustrating to keep rotating the tablet after adding bookmarks to hoarder and then rotate it back to landscape mode

### Can the goal of this request already be achieved via other means?

not on android app afaik

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

MohamedBassem:
Tracking this in #1188

PR #1785, opened 2025-07-26
feat(mobile): add context aware sharing option in mobile app

### LINKED ISSUE #1188: Outbound sharing from mobile app

Currently, the mobile app (at least on Android) doesn't have any obvious means to get hoarded images out of the app into other apps. It would be good to enable the long press context menu to access sharing so the image could be shared to other apps. As a side note the preview page for viewing a hoarded image shouldn't crop the image if it already fits entirely on screen

More usability from the apps - being able to readily access hoarded images in other apps

Technically the PWA can do this since it's just rendering the image in the browser anyway, but it can't receive OS shares so it's much less usable to collect stuff. To get full functionality you currently need both the PWA and the native app.

### Additional context

_No response_

xuatz:
## Summary

* add context aware sharing option in mobile app

fixes #1188 
Also mentioned in #1054 #1343 #1616 #1779 lol

## Details

Originally, @MohamedBassem mentioned that this should be an action in the menu, but maybe it's better UX to have a share icon, although that may eventually clutter up the bookmark bottom row 🤔 

For now I attached the screenshots for both versions, and maybe see if anyone have any opinions to chip in on which should we go with?

## Notes

For now, only tested on android (physical device).
And only for `text`, `image` and `url`.

Would be nice if someone can help to test the feature on ios (physical device) and also other types of bookmarks ^^

### Menu Version

image

### Share Icon Version

image

### Text to Clipboard

image

### Image to native share intent

image

### Url to native share intent

image


xuatz:
cc @xxjoe2 @Clancy508 @chreniuc @Lightjohn

let me know if you have any preference for the UX/feature 👍

Lightjohn:
Seems pretty good to me :) Exactly what I was thinking

MohamedBassem:
Thanks @xuatz!

PR #1796, opened 2025-07-28
feat: rediscovery feed

### LINKED ISSUE #435: Add "Random Bookmarks View" to rediscover forgotten links
In large bookmark collections, many valuable links become forgotten as they get buried under newer additions. To help users rediscover these overlooked bookmarks, I propose adding a "Random Bookmarks View" feature.

Proposed Solution:

 - Random Bookmarks View: Introduce a new view that displays bookmarks in a random order.
 - Persistent Randomness: Use a constant seed for the randomization so that the order remains the same across page refreshes. The seed could automatically change daily or be manually reset via a "Shuffle" button.

Benefits:

 - Rediscovery: Encourages users to revisit and engage with older bookmarks they might have forgotten.
 - Enhanced Engagement: Adds an element of surprise and exploration, making the bookmark manager more interactive.
 - Personalization: Gives users control over how they view and interact with their bookmark collection.

This feature is akin to how platforms like Google Photos highlight past memories, prompting users to revisit content from a year ago or more. Implementing a similar concept in. Hoader can enhance user experience by bringing forgotten but potentially valuable bookmarks back into view.

xuatz:
fixes #435

### DIFF
--- a/apps/mobile/app/dashboard/(tabs)/index.tsx
+++ b/apps/mobile/app/dashboard/(tabs)/index.tsx
@@ -11,7 +11,7 @@ import { useToast } from "@/components/ui/Toast";
 import useAppSettings from "@/lib/settings";
 import { useUploadAsset } from "@/lib/upload";
 import { MenuView } from "@react-native-menu/menu";
-import { Plus, Search } from "lucide-react-native";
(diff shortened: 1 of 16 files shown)


mashraf92:
@xuatz looking forward to rediscovery feature

Issue #1793, opened 2025-07-28
Incorrect export name


playtime-backrest:
### Describe the Bug

It's less about being "incorrect" and more about needing to be updated, as exporting a JSON list will give us the original name of `hoarder-export-date-string.json`. "Hoarder" should be changed to the new name of Karakeep.

### Steps to Reproduce

1. Settings
2. User Settings
3. Export
4. JSON
5. See name

### Expected Behaviour

`karakeep-export-date-string.json`

### Exact Karakeep Version

0.26.0

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

dvdpearson:
Quickly created https://github.com/karakeep-app/karakeep/pull/1829

PR #1829, opened 2025-08-09
Renamed export filename to karakeep

dvdpearson:
Quick fix resolving https://github.com/karakeep-app/karakeep/issues/1793


MohamedBassem:
Thanks @dvdpearson!

Issue #1834, opened 2025-08-11
WebView scrolling on iOS doesn't match rest of app or Safari


landonepps:
### Describe the Bug

The mobile app uses a WebView to display the bookmark preview.
WebView on iOS has a `"fast"` deceleration rate by default, which doesn't match the rest of the app and makes scrolling through long articles tedious.
We should update it to `"normal"` to match the non-WebView screens in the app as well as Safari.

https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#decelerationrate

<table>
<tr>
<td>fast (current behavior)</td>
<td>normal (expected behavior)</td
</tr>
<tr>
<td>
<video src=https://github.com/user-attachments/assets/7f594d06-2543-40ab-8e98-3b7a8e168ac9>
</td>
<td>
<video src=https://github.com/user-attachments/assets/e63bf131-fda9-498e-bd2d-bd458c987628>
</td>
</tr>
</table>

### Steps to Reproduce

1. Open any bookmark or archive preview in the iOS app.
2. Attempt to scroll quickly down the page.
3. Note that the scroll stops abruptly.
4. Tap the "browser" icon to open the bookmark in Safari.
5. Attempt to scroll quickly down the page.
6. Note that the scroll decelerates much more slowly.

### Expected Behaviour

Scrolling should have the same behavior on the bookmark/archive preview as the rest of the Karakeep app and Safari.

### Device Details

iOS

### Exact Karakeep Version

1.7.0

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

landonepps:
PR got merged. Thanks!

PR #1835, opened 2025-08-11
ui(mobile): Fix WebView deceleration rate on iOS

landonepps:
The mobile app uses a WebView to display the bookmark preview.
WebView on iOS has a `"fast"` deceleration rate by default, which doesn't match the rest of the app and makes scrolling through long articles tedious.
We should update it to `"normal"` to match the non-WebView screens in the app as well as Safari.

https://github.com/react-native-webview/react-native-webview/blob/master/docs/Reference.md#decelerationrate

Issue: #1834

<table>
<tr>
<td>fast (old behavior)</td>
<td>normal (new behavior)</td
</tr>
<tr>
<td>
<video src=https://github.com/user-attachments/assets/7f594d06-2543-40ab-8e98-3b7a8e168ac9>
</td>
<td>
<video src=https://github.com/user-attachments/assets/e63bf131-fda9-498e-bd2d-bd458c987628>
</td>
</tr>
</table>

### DIFF
--- a/apps/mobile/components/bookmarks/BookmarkLinkPreview.tsx
+++ b/apps/mobile/components/bookmarks/BookmarkLinkPreview.tsx
@@ -112,6 +112,7 @@ export function BookmarkLinkReaderPreview({
         }}
         showsVerticalScrollIndicator={false}
         showsHorizontalScrollIndicator={false}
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thank you!

PR #1847, opened 2025-08-17
feat: Support video uploads and attachments

packetmonkey:
This commit allows the following mime types to be uploaded and attached as video assets on bookmarks.

- video/mp4
- video/webm
- video/x-matroska

This seems to be the minimal diff that adds the functionality in my dev environment. This change only adds the functionality to the API and makes no UI facing changes. Videos uploaded and attached using the API appear on the related bookmark similar to a video added using the native yt-dlp functionality.

I'm happy to make any changes needed to help land this.

Thanks for the amazing app!

### DIFF
--- a/packages/shared/assetdb.ts
+++ b/packages/shared/assetdb.ts
@@ -26,8 +26,17 @@ export const enum ASSET_TYPES {
   IMAGE_WEBP = "image/webp",
   APPLICATION_PDF = "application/pdf",
   TEXT_HTML = "text/html",
+
   VIDEO_MP4 = "video/mp4",
+  VIDEO_WEBM = "video/webm",
+  VIDEO_MKV = "video/x-matroska",
 }
+
+export const VIDEO_ASSET_TYPES: Set<string> = new Set<string>([
+  ASSET_TYPES.VIDEO_MP4,
+  ASSET_TYPES.VIDEO_WEBM,
+  ASSET_TYPES.VIDEO_MKV,
+]);
 
 export const IMAGE_ASSET_TYPES: Set<string> = new Set<string>([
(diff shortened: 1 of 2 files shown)


Drashi:
Thank you, this is the functionality I miss the most, along with the support for GIF upload.

packages/shared/assetdb.ts
```diff
@@ -38,13 +47,15 @@ export const IMAGE_ASSET_TYPES: Set<string> = new Set<string>([
...
   ...IMAGE_ASSET_TYPES,
+  ...VIDEO_ASSET_TYPES,
   ASSET_TYPES.TEXT_HTML,
   ASSET_TYPES.APPLICATION_PDF,
 ]);
 
 // The assets that we allow as a bookmark of type asset
 export const SUPPORTED_BOOKMARK_ASSET_TYPES: Set<string> = new Set<string>([
   ...IMAGE_ASSET_TYPES,
+  ...VIDEO_ASSET_TYPES,
```
MohamedBassem: To add full support for a new bookmark type, it'll require more effort than just that. For example, a new asset type must be defined (e.g. similar to image and pdf) and lots of places needs to be updated to support it. If all what you need is to attach "it" to a bookmark via API, then I'd remove it from here.

MohamedBassem:
Thanks for the PR, I think if all what you need is video attachments via APIs, then let's keep the PR's scope to that. Adding a new bookmark asset type is a bit more work and might deserve its own PR.

packetmonkey:
Thanks for the info! My goal is to add support only using the API to enable workarounds. I have updated the pull request.

MohamedBassem:
Thank you!

Issue #1787, opened 2025-07-26
Export Links and Notes JSON file doesn't work (15k bookmarks, 32k tags)


sprior:
### Describe the Bug

I love Karakeep's functionality!  I've got 15k bookmarks (mostly imported from Pocket) and 32k AI generated tags.  I've done extensive bookmark and tagging cleanup and I'd like to do a JSON export for safekeeping.  When I click on export file in Karakeep JSON format I get no visual acknowledgement of the request at all so I don't know whether it's started but is going to take a while, or maybe it aborted - no way to know.  Doing a  docker logs -f karakeep-web-1 doesn't show any indication of the request at all though I see logs of health check calls after that.

### Steps to Reproduce

1. Start with 15k bookmarks, 32k tags
2. Go to User Settings -> Import/Export
3. Click on the black Export button
4. Nothing at all happens

### Expected Behaviour

I think if this is going to be an extended operation (which it probably is) I should at least get an immediate acknowledgement that the process has started.  It would make sense for this to be kicked off as a background activity and then I get a notification when the export file is ready for download, this approach would also address any web browser timeout issues.  But for all I know the export process just silently crashed.

### Device Details

Firefox 141, Windows 11

### Exact Karakeep Version

v0.26.0

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

MohamedBassem:
@sprior anything in the logs when you trigger the export?

sprior:
The only thing that appears in the logs is:


<-- GET /api/trpc/users.settings,users.whoami?batch=1&input=%7B%220%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%2C%221%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%7D
--> GET /api/trpc/users.settings,users.whoami?batch=1&input=%7B%220%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%2C%221%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%7D 200 71ms
<-- GET /api/health
--> GET /api/health 200 1ms

sprior:
[...]

PR #1870, opened 2025-08-24
feat: Show loading indicator while export file is being generated #1787

Ahmed-Abdel-karim:
## Summary
Add a loading indicator to the export button #1787 

---

### Changes
- Use `button` instead of `Link` for starting the export  
- Use **Fetch API** to create a function that starts the download process  
- Wrap this function with `useQuery` to track the loading state  
- Disable the export button and show a spinner while fetching the export data  
- Create a download link dynamically and start the download
- show a toast if an error occurred

https://github.com/user-attachments/assets/53c14212-36d9-4361-802a-5deaa97d20db

### DIFF
--- a/apps/web/components/settings/ImportExport.tsx
+++ b/apps/web/components/settings/ImportExport.tsx
@@ -1,8 +1,7 @@
 "use client";
 
-import { useState } from "react";
-import Link from "next/link";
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thanks @Ahmed-Abdel-karim!

PR #1876, opened 2025-08-27
feat: add gif asset type support

Drashi:
This commit allows "image/gif" mime type to be uploaded and attached to bookmarks.


MohamedBassem:
Thank you!

PR #1891, opened 2025-09-05
Fatal typo in docker-compose.yml for meilisearch container

dierochade:
the blueprint fort he compose contains

    environment:
      - MEILI_NO_ANALYTICS="true"

This should be a boolean, but is formatted as string. 

Leads to: 

error: invalid value '"true"' for '--no-analytics'
  [possible values: true, false]

and breaks the container at startup

## Summary by CodeRabbit

* **Bug Fixes**
  * Ensured the search service’s telemetry is reliably disabled in container deployments by correcting a configuration value to the expected boolean.
  * Prevents intermittent analytics pings that could occur due to a mis-typed configuration, improving user privacy and compliance.
  * Aligns runtime configuration with the service’s expected types to avoid environment-specific inconsistencies (local, CI, staging).

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -35,7 +35,7 @@ services:
     env_file:
       - .env
     environment:
-      MEILI_NO_ANALYTICS: "true"
+      MEILI_NO_ANALYTICS:true
     volumes:
       - meilisearch:/meili_data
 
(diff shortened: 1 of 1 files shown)


MohamedBassem:
E2E tests have been running with this docker compose file as is without issues (https://github.com/karakeep-app/karakeep/blob/main/packages/e2e_tests/docker-compose.yml#L22). Something is probably wrong in your environment. Your template is also different than what's in the repo. You're passing a list of envs to docker compose while the repo is passing a map. Maybe that's the difference .

PR #1907, opened 2025-09-07
feat: Replace Chat Completions with Responses API for all OpenAI models

camjac251:
## Summary
- **Complete migration** from Chat Completions API to Responses API  
- Add support for GPT-5 features (reasoning effort, verbosity control)
- Model-aware parameter handling (GPT-4 vs GPT-5 capabilities)
- Upgrade OpenAI SDK from v4.104.0 to v5.19.1
- Update default models to GPT-5-mini for future readiness
- No changes to InferenceClient public interface

## New Configuration Options
- `INFERENCE_REASONING_EFFORT`: Control reasoning depth (`minimal`/`low`/`medium`/`high`, default: `low`)
- `INFERENCE_VERBOSITY`: Output verbosity for GPT-5 models (`low`/`medium`/`high`, default: `medium`)
- Default models updated to `gpt-5-mini` (configurable via `INFERENCE_TEXT_MODEL` and `INFERENCE_IMAGE_MODEL`)

## Why this change?
OpenAI's Responses API is the future-facing interface for all their models. Chat Completions is being phased out. This migration is necessary for GPT-5 compatibility and new features like reasoning control.

## Key Implementation Details
- GPT-5 models receive verbosity and full reasoning effort parameters
- GPT-4 models get empty reasoning object and `max_output_tokens` parameter
- Image inference restructured to use `input_image` format
- Response extraction handles multiple API response structures
- Structured output uses direct Zod schema conversion
- Removed dependency on `zodResponseFormat` helper

## Known Limitations
- The OpenAI SDK v5 doesn't export TypeScript types for the Responses API yet, requiring use of `any` types in 4 places (with eslint-disable comments)
- Once OpenAI adds proper type definitions, these can be replaced with proper types

## Testing
- `pnpm typecheck --filter=@karakeep/shared --filter=@karakeep/trpc`
- `pnpm lint --filter=@karakeep/shared --filter=@karakeep/trpc`
- Manual testing with real API keys:
  - ✅ GPT-4 and GPT-5 model compatibility verified
  - ✅ Text inference (bookmark summaries, general queries)
  - ✅ Structured output (auto-tagging with Zod schemas)
  - ✅ Image analysis (screenshot processing, base64 images)
  - ✅ All inference modes tested with production-like payloads

### DIFF
--- a/packages/shared/config.ts
+++ b/packages/shared/config.ts
@@ -60,8 +60,8 @@ const allEnv = z.object({
   OLLAMA_KEEP_ALIVE: z.string().optional(),
   INFERENCE_JOB_TIMEOUT_SEC: z.coerce.number().default(30),
   INFERENCE_FETCH_TIMEOUT_SEC: z.coerce.number().default(300),
-  INFERENCE_TEXT_MODEL: z.string().default("gpt-4.1-mini"),
-  INFERENCE_IMAGE_MODEL: z.string().default("gpt-4o-mini"),
+  INFERENCE_TEXT_MODEL: z.string().default("gpt-5-mini"),
+  INFERENCE_IMAGE_MODEL: z.string().default("gpt-5-mini"),
   EMBEDDING_TEXT_MODEL: z.string().default("text-embedding-3-small"),
   INFERENCE_CONTEXT_LENGTH: z.coerce.number().default(2048),
   INFERENCE_MAX_OUTPUT_TOKENS: z.coerce.number().default(2048),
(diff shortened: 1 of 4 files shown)


mratsim:
That doesn't seem to work with OpenAI compatible servers that do not provide the Responses API, for example https://github.com/ggml-org/llama.cpp/issues/14702

This will significantly reduce the ability of Karakeep to be fully self-hosted with no data leak. From a versioning perspective it's also a major breakage.

camjac251:
Ah yeah, I was thinking this could be a first step before a full refactor (working on it locally), moving it to a provider agnostic (OpenAI/Anthropic/Google), but this as it is would break many other compatibility services currently if not using OpenAI. I'll try to roll it back to maybe offer it as a flag maybe to choose between the Responses API and Chat Completions API (legacy)

mratsim:
Alternatively you can ask for the full path `.../v1/chat/completions` or `.../v1/responses` and disambiguate from here, while keeping non-full path defaulting to `/v1/chat/completions` to avoid breaking existing setup on update.

A deprecation warning can be added in an admin panel that non-full path will be phased out in "some version".

PR #1923, opened 2025-09-09
Update 13-community-projects.md

fivestones:
Updated the Raycast Extension link


MohamedBassem:
Thanks for the fix! Can you also update `docs/docs/13-community-projects.md` so that it doesn't get reverted in the next release.

PR #1927, opened 2025-09-10
Update sameSite options to match currently found values

a10kiloham:
errors out on valid capital cases and a couple other enum types.


MohamedBassem (changes requested):
Sorry for reviewing this PR so late (better late than never though).

I think coderabbit here is correct and this wouldn't pass the typecheck. Those values are indeed case insensitive, but you'll need to covert them to the values accepted by playwright. I would have done it myself but because this PR is coming from your main branch, I wouldn't be able to modify the branch myself.

MohamedBassem:
Closing the PR for inactivity, feel free to reopen it once the comments are addressed.

PR #1928, opened 2025-09-10
Add clarity on cookie path

a10kiloham:


MohamedBassem:
Hey, thanks for the PR, though I think the clarification is a bit redundant. It currently already say `**Path to a JSON file** containing cookies`.

PR #1930, opened 2025-09-11
add current tab title while saving from extension

Abel-Liu:
Add current tab title while saving from extension.

Benefits:

1. Tab title shows the website's title which is meaningful and good to read.
2. Avoid empty title when sync bookmark to browser using Floccus.


MohamedBassem:
Thank you!

PR #1723, opened 2025-07-11
feat(mobile): Add user setting for default bookmark view mode

xuatz:
## Summary
- Adds a new user setting for default bookmark view mode on mobile devices
- Allows users to choose between "Reader" and "Browser" as their preferred default view when opening bookmarks
- Setting is stored locally on the mobile device for instant access and offline capability

## Changes Made
- Added `mobileBookmarkClickDefaultViewMode` setting to mobile app's local settings
- Implemented settings UI in mobile Settings > Default Bookmark View
- Updated bookmark view component to use the user's preferred default view mode
- Default value is "Reader" for new users

## User Experience
- Users can now set their preferred default view mode (Reader or Browser) for bookmarks
- The setting is applied immediately when opening bookmarks on mobile
- No network calls required - setting is stored locally for better performance
- Setting persists across app sessions

### Tested on ios simulator

I've also tested on the android app on device, connected to a karakeep server on v0.25.0 to make sure that the latest version of the app works even on older server instances.

Resolves #1603

🤖 Generated with [Claude Code](https://claude.ai/code)

Co-Authored-By: Claude <noreply@anthropic.com>

PR #1946, opened 2025-09-14
fix(dev): worker not started properly in helper start script

xuatz:
## Summary

Fixed start-dev.sh script not starting workers, resulting in incomplete migration (queue.db tables not created).

Good news, seems like it hasn't affected anyone else besides me ^^

### DIFF
--- a/start-dev.sh
+++ b/start-dev.sh
@@ -61,16 +61,12 @@ if [ -n "$DATA_DIR" ] && [ ! -d "$DATA_DIR" ]; then
     mkdir -p "$DATA_DIR"
 fi
 
+echo "Running database migrations..."
+pnpm run db:migrate
+
 echo "Starting web app and workers..."
 pnpm web & WEB_PID=$!
 pnpm workers & WORKERS_PID=$!
-
-# Give services a moment to start
-sleep 4
-
-# Run database migrations after services have started
-echo "Running database migrations..."
-pnpm run db:migrate || echo "Note: Database might already be migrated"
 
 # Function to handle script termination
 cleanup() {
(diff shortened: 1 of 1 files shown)


MohamedBassem:
hmm, not sure I understand what's the problem. I think if anything, the db migration should run "before" the services are started. Also, `db:migrate` doesn't create the worker tables (the worker creates it on startup).

xuatz:
> hmm, not sure I understand what's the problem. I think if anything, the db migration should run "before" the services are started. Also, `db:migrate` doesn't create the worker tables (the worker creates it on startup).

@MohamedBassem i might not have truly identified the root problem, but with the previous version of the script, when i wipe the data and start from scratch, i will get this error

Screenshot 2025-09-14 at 22 00 29

I git bisect and found the commit that introduced the problem was from the last change in this file in https://github.com/karakeep-app/karakeep/pull/1723/files#diff-2a8f4ba8ca6b7526d6e5e2613237a4bbc525398d7f8c18d7997bf577aba4752c

> Also, `db:migrate` doesn't create the worker tables (the worker creates it on startup).

If this is the case, then its simply because this line ended up blocking the script execution, which cause the workers to simply not start

```
# Wait for web app to be ready
echo "Waiting for web app to start..."
until curl -s http://localhost:3000 > /dev/null 2>&1; do
    sleep 1
done
```

MohamedBassem:
This usually means that the worker failed to start or wasn't running when you opened the web app. Can you try:
1. Run the migrations
2. Start the web and workers
3. Wait until web is reay?

xuatz:
> This usually means that the worker failed to start or wasn't running when you opened the web app. Can you try:
> 
> 1. Run the migrations
> 2. Start the web and workers
> 3. Wait until web is reay?

@MohamedBassem seems good 👍 i applied the suggestions.

MohamedBassem:
Thanks!

Issue #1915, opened 2025-09-08
Metrics endpoint "Bad Request"


Emastermind:
### Describe the Bug

After upgrading to 0.27.0, metrics endpoint doesn't work anymore. The following request throw a 400 "Bad Resquest" error:

```bash
curl --location 'https://<MY_DOMAIN>/api/metrics' \
--header 'Authorization: Bearer <MY_TOKEN>'
```

```bash
HTTP/2 400 
server: nginx
content-type: text/plain;charset=UTF-8
access-control-allow-origin: *
access-control-allow-methods: GET, POST, PUT, PATCH, DELETE, OPTIONS
access-control-allow-headers: Content-Type, Authorization
access-control-allow-credentials: true
vary: RSC, Next-Router-State-Tree, Next-Router-Prefetch, Next-Router-Segment-Prefetch
www-authenticate: Bearer error="invalid_request"
x-powered-by: Hono

Bad Request  
```

### Steps to Reproduce

Use curl to request endpoint with a valid metrics token:

```bash
curl --location 'https://<MY_DOMAIN>/api/metrics' \
--header 'Authorization: Bearer <MY_TOKEN>'
```

See that endpoint return 400 bad request. No 

### Expected Behaviour

prometheus metrics results

### Screenshots or Additional Context

- Karakeep docker image version "release"
- Behind nginx reverse proxy
- Rollback to v0.26.0 fix the issue

### Exact Karakeep Version

v0.27.0

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

MohamedBassem:
Anything in the logs? Also, I assume that token that you're passing matches `PROMETHEUS_AUTH_TOKEN`? Can you trying changing the token to one without special chars temporarily and see if this helps?

I'm using the metrics endpoint on 0.27 and seems to be working just fine.

Emastermind:
Indeed ... special chars in token is the root cause. Token with specicial char like `@` or `#` return a 400.
I swicth to alphanumeric token for now. Thanks for the support and project, I use it daily!

MohamedBassem:
@Emastermind glad it worked!

PR #1957, opened 2025-09-16
Add doc updates for prometheus metrics

claytono:
This adds documentation for `WORKERS_HOST` and clarifies token requirements for `PROMETHEUS_AUTH_TOKEN`.

When setting up the prometheus metrics, I ran into a problem where the token I chose originally had special characters and would fail mysteriously with `400 Bad Request`.  This was reported in https://github.com/karakeep-app/karakeep/issues/1915, but not documented at that time.  In addition, if you are using `WORKERS_PORT` to scrape metrics inside of a container, it won't be reachable since the default is `127.0.0.1`.  I had to override this undocumented config item to `0.0.0.0` so Prometheus could reach it.


MohamedBassem:
Thank you

PR #1985, opened 2025-09-27
Tagclean

grimwiz:


grimwiz:
First speedup test - windowing the gui and delete queues

grimwiz:
ok

grimwiz:
opened against the karakeep-app distro instead of my grimwiz one, I want to test it first before submitting.

PR #2019, opened 2025-10-07
fix: fix dev script shebang for better compatibility

maya-doshi:
this will work more consistently across distros. the karakeep-linux.sh script already uses this shebang


MohamedBassem:
thank you!

PR #2031, opened 2025-10-12
feat(feeds): Add import tags from RSS feed categories

### LINKED ISSUE #1996: Import tags from feed

Would it be possible when importing from an RSS Feed to (optionally, configured for each feed) add the tags for each item as tags for that bookmark?

It's a feature that could be useful for multiple people i guess

Not that I could find with the current feed import functionality

### Additional context

My personal use case for this:
I use Readeck to read articles. Some of them I want in Karakeep as well as bookmarks. To do so I:

* Mark the article as a favorite in Readeck. Optionally I add some labels (tags) in Readeck
* In FreshRSS I import the Readeck favorites using the json api(++)
* In FreshRSS I created a saved query for all the Readeck favorites which makes it available as a feed again, including the tags
* In Karakeep I subscribe to that feed.
* When I favorite an article in Readeck it gets added to Karakeep, but without the labels I assigned in Readeck.

(++) Why passing everything through FreshRSS?
Two reasons:
1. The RSS feed that Readeck makes available requires login, which Karakeep does not support. 
2. The links in the Readeck RSS point to the readeck instance instead of the original articles

So in FreshRSS I created a 'feed' which uses the Readeck json api.
* FreshRSS has support to use login credentials so that already fixes problem 1
* By using the json api and transforming that into a feed, then making it available again to the outside world using a saved query, fixes problem 2. Feed is set up as follows:
    * https://readeck.local/api/bookmarks?limit=10&sort=-created&is_marked=1
    * type of feed source: lson (dot notation)
    * dot notation for finding news items: .
    * dot notation for title: title
    * dot notation for content: description
    * dot notation for link: url (this is the link to the original article, while in the original readeck feed it points to the readeck instance itself)
    * dot notation for item tags: labels
    * dot notation for unique ID: id

MohamedBassem:
- Add importTags boolean field to rssFeedsTable schema (default: false)
- Create database migration 0063_add_import_tags_to_feeds.sql
- Update zod schemas (zFeedSchema, zNewFeedSchema, zUpdateFeedSchema) to include importTags
- Update Feed model to handle importTags in create and update methods
- Update feedWorker to:
  - Read title and categories from RSS parser
  - Attach categories as tags to bookmarks when importTags is enabled
  - Log warnings if tag attachment fails

Resolves #1996

### DIFF
--- a/packages/db/drizzle/0063_add_import_tags_to_feeds.sql
+++ b/packages/db/drizzle/0063_add_import_tags_to_feeds.sql
@@ -1,1 +1,0 @@
-ALTER TABLE `rssFeeds` ADD `importTags` integer DEFAULT false NOT NULL;


MohamedBassem:
@claude Can you expose this new option to the feed options UI?

MohamedBassem:
@codex review

teranex:
upgraded to Karakeep 0.28, work perfectly :)
Thank you.

KingOfSpades:
> upgraded to Karakeep 0.28, work perfectly :)
> 
> Thank you.

Can you explain me how this works? I thought this meant we would be able to add custom tags to each RSS import

MohamedBassem:
@KingOfSpades if you want to add custom tags on RSS imports, you can do that with the rule engine. This change uses the tags exposed by the RSS feed as tags to the bookmark.

PR #2055, opened 2025-10-18
Fix list: filter to include smart lists in search queries

### LINKED ISSUE #1669: Searching query "list:" doesn't work on Smart Lists
### Describe the Bug

It looks like Search Query syntax doesn't work on Smart Lists.

### Steps to Reproduce

1. I've Smart List called "NOT ASSIGNED"
2. When I use search syntax list:"NOT ASSIGNED" it doesn't seem to show any articles from this list
3. When I search for other lists it works fine e.g. list:"Toys" etc.
4. I'm on 0.25.0 Docker based version

### Expected Behaviour

Search syntax for lists works on both standard and smart lists. 

_No response_

### Device Details

_No response_

### Exact Karakeep Version

0.25.0

AdrianAcala:
- Update getIds function to handle both manual and smart lists when filtering by list name
- Smart lists are resolved by evaluating their stored query recursively
- Add cycle detection to prevent infinite recursion in nested smart list queries
- Maintain backward compatibility with existing manual list filtering

Fixes #1669

### DIFF
--- a/apps/web/components/ui/markdown/markdown-readonly.tsx
+++ b/apps/web/components/ui/markdown/markdown-readonly.tsx
@@ -85,11 +85,13 @@ export function MarkdownReadonly({
         },
         code({ className, children, ...props }) {
           const match = /language-(\w+)/.exec(className ?? "");
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
+          const { ref, ...rest } = props as any;
           return match ? (
             <SyntaxHighlighter
               PreTag="div"
               language={match[1]}
-              {...props}
+              {...rest}
               style={dracula}
             >
               {String(children).replace(/\n$/, "")}
--- a/packages/db/schema.ts
+++ b/packages/db/schema.ts
@@ -395,6 +395,8 @@ export const bookmarkLists = sqliteTable(
   },
   (bl) => [
     index("bookmarkLists_userId_idx").on(bl.userId),
+    // Helps queries that lookup lists by (userId, name)
+    index("bookmarkLists_userId_name_idx").on(bl.userId, bl.name),
     unique("bookmarkLists_userId_id_idx").on(bl.userId, bl.id),
   ],
 );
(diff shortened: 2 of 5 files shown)


xuatz:
I want this feature too! I'll try to help review this PR, hopefully soon™️ lol

xuatz:
@AdrianAcala I think we are missing (at least) 1 (more) scenario.

## How to reproduce

1. create a bookmark
2. give the bookmark a tag (e.g. #cooking)
4. create a smart list called `cooking` with the following query
    * `#cooking`
5. open the list and expect to see that the bookmark shows up in the smart list
7. archive the bookmark
5. create a smart list called "archived but not in list" with the following query
    * `is:archived and -is:inlist`
8. open the list

## Expected

Should NOT see bookmark created in 1 in this list

## Actual

bookmark can be found in this list

---

This means that the karakeep search query language's `is:inlist` still doesn't consider bookmarks in smart lists as `inlist`

Do you think it makes sense to also include to handle for this scenario in your PR?

If that's too much to ask, I'll separately make a new issue later.

xuatz:
Otherwise, as for the intended scope of the PR, the feature seems to be working ^^

I also tried messing around with having multiple lists with the same name, smart or manual lol.

xuatz (approved):
@MohamedBassem you're probably more familiar with potential edge cases, but LGTM 👍

AdrianAcala:
@xuatz , added the fix and test for the `is:inlist` error. Please review @xuatz and @MohamedBassem .

apps/web/components/ui/markdown/markdown-readonly.tsx
```diff
@@ -85,11 +85,13 @@ export function MarkdownReadonly({
...
+          // eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
+          const { ref, ...rest } = props as any;
           return match ? (
             <SyntaxHighlighter
               PreTag="div"
               language={match[1]}
-              {...props}
+              {...rest}
               style={dracula}
             >
```
xuatz, line 96:
```suggestion
          code({ className, children, ref, ...props }) {
            const match = /language-(\w+)/.exec(className ?? "");
            return match ? (
              <SyntaxHighlighter
                PreTag="div"
                language={match[1]}
                {...props}
                style={dracula}
              >
```

how about this instead?

MohamedBassem:
This is an old PR that also already got fixed in main. Closing for now.

Issue #462, opened 2024-10-03
Add support for more AI providers in a generic way


kamtschatka:
Different people have already asked for "their" AI provider to be supported.
It is unlikely that we add support for all of them, but we could switch to a library that allows us to connect to at least the more well known ones.
In the discussion https://github.com/hoarder-app/hoarder/discussions/453, there was a list of possible solutions already:
* https://github.com/themaximalist/llm.js
* https://github.com/samestrin/llm-interface
* https://github.com/Portkey-AI/gateway/

kamtschatka:
Requested providers so far:
* OpenRouter (seems to work with OpenAI support already)
* OpenAI Azure (seems to have some issues: #137)
* 1min.ai (#451)
* Gemini (https://github.com/hoarder-app/hoarder/discussions/453)
* Perplexity (#262)

bhupesh-sf:
What about local LLM's using ollama. All we need is to give an option to configure OpenAI Base url and there are many providers which 
 given OpenAI compatible endpoint for the use of local  LLMs.

MohamedBassem:
@bhupesh-sf hoarder already supports local llms natively. Check the inference section in the configuration documentation

MohamedBassem:
Gemini now has an OpenAI compatible API as well: https://developers.googleblog.com/en/gemini-is-now-accessible-from-the-openai-library/

bebound:
I've used https://github.com/stulzq/azure-openai-proxy to simulate Azure OpenAI as OpenAI. It works well for [chenzhaoyu94/chatgpt-web](https://github.com/Chanzhaoyu/chatgpt-web).

When use the same config in hoarder, it shows "something went wrong", and I can't find any useful message in the log.

dinnouti:
+1 for Amazon Bedrock LLMs like Claude, Meta, Cohere, so-on

jbohnslav:
If you use a LiteLLM proxy, you can already connect to all of these LLMs via the OpenAI api.

xiaoduo:
If can support OpenAI compatible API, would be enough.

dinnouti:
Just to close the loop on the Bedrock, AWS has a sample OpenAI-compatible RESTful APIs for Amazon Bedrock code.

https://github.com/aws-samples/bedrock-access-gateway

bradhawkins85:
For those wanting to use Gemini here is the section from my Docker .env that works perfectly.

[...]

Issue #137, opened 2024-05-04
Azure OpenAI:  Can't seem to get it to work


markheaps:
I have added the
OPENAI_BASE_URL: https:///xxxx.openai.azure.com
OPENAI_API_KEY: API Key from Azure Portal
INFERENCE_TEXT_MODEL: GPT4
INFERENCE_IMAGE_MODEL: GPT4

The deployment is setup as:

<img width="283" alt="image" src="https://github.com/MohamedBassem/hoarder-app/assets/140619805/6b12b81f-45b4-4d12-af55-272547f1fdd9">

The name is GPT4:

<img width="247" alt="image" src="https://github.com/MohamedBassem/hoarder-app/assets/140619805/421fe672-8a90-41a4-bd89-ef3de3d530dd">

I go in and add a bookmark and one gets added, but no tags are added to it.  If I understand this correctly, isn;t it supposed to use GPT to auto tag, etc?

MohamedBassem:
Can you share the logs of the workers container? Also in the example you shared, do you have an extra slash after "https" or is that just a typo?

markheaps:
Sorry was just a typo in my post:

<img width="357" alt="image" src="https://github.com/MohamedBassem/hoarder-app/assets/140619805/befa1a3d-fa17-4b11-ab4b-476437995ffe">

Logs:

Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-9.0.6.tgz.
> @hoarder/workers@0.1.0 start:prod /app/apps/workers
> tsx index.ts
2024-05-04T16:47:32.687Z info: Workers version: nightly
2024-05-04T16:47:32.720Z info: [Crawler] Connecting to existing browser instance: http://chrome:9222
(node:34) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
2024-05-04T16:47:32.737Z info: [Crawler] Successfully resolved IP address, new address: http://172.20.0.4:9222/
2024-05-04T16:47:35.101Z info: Starting crawler worker ...
2024-05-04T16:47:35.103Z info: Starting inference worker ...
2024-05-04T16:47:35.104Z info: Starting search indexing worker ...
2024-05-04T16:50:31.725Z info: [search][3] Attempting to index bookmark with id kmehwk3wl5wnitrdbi39kqhi ...
[...]

PR #2072, opened 2025-10-26
docs: Add Azure configuration details for OpenAI-compatible API

cloudchristoph:
This PR adds documentation for using Azure as an OpenAI-compatible API provider. The update clarifies how to configure the API key, base URL, and model deployment name when integrating with Azure services.

Helps with issues like #137 and #462


MohamedBassem:
Thanks a lot @cloudchristoph!

Issue #988, opened 2025-02-06
Organize imported bookmarks based on existing folder structure


jbaker0428:
## Background
I just tried importing 2500 bookmarks from an HTML file for the first time and was dismayed to find that it put them all in one list without any tags. I would prefer to preserve my existing organization rather than spin up AI to create a different set of tags. The AI would just not have enough of the original context to be able to do a good job. 

## New Functionality
I would like for the Hoarder import tool to organize the imported data based on the folder structure from my Chrome bookmarks that's encoded in the HTML file. 

For example, my Chrome bookmarks bar contains a folder structure of:

- Projects
  - NAS
    - Hardware
    - Software

That structure could be converted to tags, nested lists, or both. I don't have a great sense of which would be more useful, but I suspect that will vary from person to person.
### Tags 
A bookmark in that Hardware folder would get three tags [Projects, NAS, Hardware]. If I happened to have a different "Hardware" folder elsewhere in my bookmarks, that can reuse the same Hardware tag. 

In the case of Chrome, the "Bookmarks Bar" and "All Bookmarks" folders should probably not get tags named after them. I imagine something similar would apply to Firefox.

### Lists
The HTML bookmark folder structure could be directly converted to a set of nested lists. This could admittedly get messy in the lists UI for complex cases. I just did a quick test on my desktop (1440p 27" monitor) and the UI only started to get weird about nesting lists once I got 9 levels deep. It created a horizontal scroll bar inside the left toolbar where the list of lists is, so it's handling that case gracefully at least.

![Image](https://github.com/user-attachments/assets/320b281b-0fa1-4538-b82a-35ad0ac9fbc8)

I'm not familiar with how this feature request plays into the non-HTML import options.

### Describe the benefits this would bring to existing Hoarder users

This would greatly accelerate how quickly new users can get up and running with their existing data. In my case, not having a feature like this may be a showstopper. I personally can't see myself sorting through that one big list to manually create equivalent tags. 

### Can the goal of this request already be achieved via other means?

Not that I've found. The Hoarder documentation doesn't indicate the existence of any additional options for importing. Perhaps it could be laundered through another program (Linkwarden etc) that already has this functionality, but I don't have access to any of those to test. 

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

MohamedBassem:
Hey, thanks for the detailed feature request. This is currently planned and is tracked in #538

Issue #538, opened 2024-10-15
Retain directory / folder structure while importing HTML export from Chrome (nested bookmarks)


samholton:
I did a test of exporting bookmarks from Chrome and importing that HTML file. The HTML file includes headers for the folder names in Chrome, Is it possible to create lists and add the bookmarks to the lists? Currently they all end up under a single imported list.

kamtschatka:
It is not possible to have the same structure, as Chrome allows folders in folders, Hoarder only allows 1 level of nesting.

kamtschatka:
ah indeed. I wanted to do that once before and it did not work. Guess we can do that then.

SeanChengN:
I have nearly 2000 bookmarks to import and it would kill me if their structure changed😫

cbingb666:
This is very much needed to make it easier to migrate large numbers into Hoarder at once!

PR #2084, opened 2025-11-03
Fix HTML import reusing existing lists

umutkeltek:
Summary

The previous upstream fix ([1bae66f7785](https://github.com/karakeep-app/karakeep/commit/1bae66f7785289818eba3249f651a320f497c6a4)￼) made the Netscape importer build a proper list tree, but it still recreated a new “Imported Bookmarks” hierarchy on every import.
This patch adds the missing reuse logic so that if a user re-imports, the existing root and its nested lists are reused instead of duplicated.

What’s changed
	•	packages/shared/import-export/importer.ts:
– fetches the current list tree before import
– reuses the existing Imported Bookmarks hierarchy
– only creates folders that don’t exist yet
	•	apps/web/lib/hooks/useBookmarkImport.ts: forwards cached list inventory to shared importer
	•	packages/shared/import-export/importer.test.ts: adds coverage for reuse behaviour
	
	Manual: import and re-import a large Raindrop/Netscape HTML export — verify that folders are reused, not duplicated.

Motivation

Fixes [#538](https://github.com/karakeep-app/karakeep/issues/538)￼
Relates to [#988](https://github.com/karakeep-app/karakeep/issues/988)￼

### DIFF
--- a/apps/web/lib/hooks/useBookmarkImport.ts
+++ b/apps/web/lib/hooks/useBookmarkImport.ts
@@ -13,6 +13,7 @@ import {
   useAddBookmarkToList,
   useCreateBookmarkList,
 } from "@karakeep/shared-react/hooks/lists";
+import { api } from "@karakeep/shared-react/trpc";
 import {
   importBookmarksFromFile,
   ImportSource,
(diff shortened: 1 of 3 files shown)


MohamedBassem:
Thanks for the PR and sorry it took too long to get to it. Unfortunately, I think the current behavior is the intended one. Every import is isolated, and then the user can merge lists as needed.

Issue #1186, opened 2025-04-02
Archive git repositories via `git clone`


PowerUser64:
### Describe the feature you'd like

It would be really cool if Hoarder could download a copy of a git repository as a way of archiving the code and its history.

### Describe the benefits this would bring to existing Hoarder users

This would make it possible to store source code archives as part of Hoarder's "personal archive" goals.

### Can the goal of this request already be achieved via other means?

As far as I'm aware, this is not possible through other means (within Hoarder, that is).

It might be possible to archive a specific version of a repository by saving a direct link to a tarball (or similar) of a release that sites like github have, but that won't get anything like the commit history of the repository.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### Additional context

Maybe a way to distinguish whether the site should be archived via `git clone` or not could be to look for the `.git` extension that is frequently at the end of urls that lead to git source repositories.

haappi:
libgit2 seems like a decent pick, which also has WASM binaries for JS (https://github.com/petersalomonsen/wasm-git). However I'm not exactly sure if the repos active or not

maya-doshi:
I have a first draft implementation on my branch (https://github.com/maya-doshi/karakeep/tree/feature/git-archive) which works mostly well enough for me. although, it uses the system git (like yt-dlp is used)

@haappi does it have a change of getting merged, ill polish it up if so

PR #2091, opened 2025-11-05
feat: archive git repositories

maya-doshi:
issue where mentioned: https://github.com/karakeep-app/karakeep/issues/1186

adds dependencies for system git and tar.

new config options:
- CRAWLER_GIT_DOWNLOAD: enable archiving (false default)
- CRAWLER_GIT_TIMEOUT: timeout for jon (60s default)
- CRAWLER_GIT_MIRROR: use --mirror flag (false default)
- CRAWLER_GIT_CLONE_DEPTH: depth of repo clone (-1, or full depth default)
- CRAWLER_GIT_CLONE_ALL_BRANCHES: get all branches (false default), currently not implemented (will finish it if approved)

this is my first time contributing to karakeep so lmk what i should change, and how i can add docuemntation

also the filename for the download only has '.gz' (no '.tar') but that bug is also present with the video downloads all coming as mp4s


xuatz:
@maya-doshi you can just add it directly to `docs/docs`. You probably want to edit [this page](https://github.com/karakeep-app/karakeep/blob/main/docs/docs/03-configuration.md) to add the new config options you are introducing.

However you should probably discuss with @MohamedBassem and make sure that everyone is one the same page about how git repo hoarding should look like for karakeep, so that there is a higher chance of the PR getting merged, especially since the issue was tagged in `icebox`. Otherwise it would be a pity if the PR was closed due to mismatch in expectation.

Should be easier to catch him in discord lol

maya-doshi:
@xuatz thank you, i did not know there was a discord. i will reach out to him there

MohamedBassem:
Hey there, thanks for the PR, but as @xuatz mentioned (and according to our [contribution guidelines](https://github.com/karakeep-app/karakeep/blob/main/CONTRIBUTING.md)), we should aim to work only on approved features that are aligned with the product's direction to avoid wasting efforts. Although archiving git repos sounds interesting, I find that it's a bit too specific and slightly not aligned with Karakeep's main goals. Karakeep's main feature is not archival but instead bookmarking. As such, I'm unfortunately leaning toward not accepting this feature for now.

PR #2149, opened 2025-11-17
feat: freshen up docker-compose.yml for local dev purposes

xuatz:
## Summary

duplicate/related to: 
* #584 
* #580

---

* opinionatedly renamed and moved `docker-compose.dev.yml` to root as `docker-compose.yml`
* applied the principles from `start-dev.sh` on `docker-compose.yml`
* defined naive defaults, so that anybody can do `docker compose up` without even setting up a .env file to have a simple dev instance running
  * still supports out of the box root `.env` overwrites due to the side-effect of having the `.env` in the the same dir as docker-compose.yml
* handles `prep` container execution and `DATA_DIR` init
* ensured that data is persisted across subsequent `docker compose up`
* ensured that fast reload works for dev server

## Details

First and foremost, why all the refactoring when there is already a `docker-compose.dev.yml`?

The "improvement" is proposed for 2 reasons:

1. i naively try to use `docker compose -f docker/docker-compose.dev.yml up` but there was an error on the latest main

```
! Corepack is about to download https://registry.npmjs.org/pnpm/-/pnpm-9.15.9.tgz
Scope: all 24 workspace projects

   ╭───────────────────────────────────────────────────────────────────╮
   │                                                                   │
   │                Update available! 9.15.9 → 10.22.0.                │
   │   Changelog: https://github.com/pnpm/pnpm/releases/tag/v10.22.0   │
   │            Run "corepack use pnpm@10.22.0" to update.             │
   │                                                                   │
   ╰───────────────────────────────────────────────────────────────────╯


> karakeep@0.1.0 db:migrate /app
> pnpm --filter @karakeep/db run migrate


> @karakeep/db@0.1.0 migrate /app/packages/db
> tsx migrate.ts


node:internal/modules/run_main:123
    triggerUncaughtException(
    ^
Error [TransformError]: 
You installed esbuild for another platform than the one you're currently using.
This won't work because esbuild is written with native code and needs to
install a platform-specific binary executable.

Specifically the "@esbuild/darwin-arm64" package is present but this platform
needs the "@esbuild/linux-arm64" package instead. People often get into this
situation by installing esbuild on Windows or macOS and copying "node_modules"
into a Docker image that runs Linux, or by copying "node_modules" between
Windows and WSL environments.

If you are installing with npm, you can try not copying the "node_modules"
directory when you copy the files over, and running "npm ci" or "npm install"
on the destination platform after the copy. Or you could consider using yarn
instead of npm which has built-in support for installing a package on multiple
platforms simultaneously.

If you are installing with yarn, you can try listing both this platform and the
other platform in your ".yarnrc.yml" file using the "supportedArchitectures"
feature: https://yarnpkg.com/configuration/yarnrc/#supportedArchitectures
Keep in mind that this means multiple copies of esbuild will be present.

Another alternative is to use the "esbuild-wasm" package instead, which works
the same way on all platforms. But it comes with a heavy performance cost and
can sometimes be 10x slower than the "esbuild" package, so you may also not
want to do that.

    at generateBinPath (/app/node_modules/esbuild/lib/main.js:1737:17)
    at esbuildCommandAndArgs (/app/node_modules/esbuild/lib/main.js:1818:33)
    at ensureServiceIsRunning (/app/node_modules/esbuild/lib/main.js:1975:25)
    at transform (/app/node_modules/esbuild/lib/main.js:1876:37)
    at file:///app/node_modules/tsx/dist/index-7AaEi15b.mjs:14:2865
    at applyTransformers (file:///app/node_modules/tsx/dist/index-7AaEi15b.mjs:14:1266)
    at transform (file:///app/node_modules/tsx/dist/index-7AaEi15b.mjs:14:2812)
    at load (file:///app/node_modules/tsx/dist/esm/index.mjs?1763396486478:2:2415)
    at async nextLoad (node:internal/modules/esm/hooks:748:22)
    at async Hooks.load (node:internal/modules/esm/hooks:385:20)

Node.js v22.21.1
/app/packages/db:
 ERR_PNPM_RECURSIVE_RUN_FIRST_FAIL  @karakeep/db@0.1.0 migrate: `tsx migrate.ts`
Exit status 1
 ELIFECYCLE  Command failed with exit code 1.
``` 

2. because @MohamedBassem himself said that "This setup wasn't super reliable for me though."

and a bonus 3rd point, I wanted to deploy a staging version of karakeep to run my own nightly, but built from my own source code (unmerged changes), so I needed a docker-compose.yml that I was comfortable and familiar with, and hence this PR.

---

So with that out of the way, I will explain a little why I moved and renamed the docker-compose.yml.

The purpose of `docker compose up` in a repo, in my opinion, is to have a fast way to get dev started.

As a result, it should not use a custom name `docker-compose.dev.yml`, or be in a non-root path, because that would go against the idea of "getting started fast".

Having the docker-compose.yml in the root also allow us to share `.env` that would be used if the developer chooses to run the application on nodejs directly.

But of course, having a `docker` folder still serves it's purpose, for organisational purposes, to put the Dockerfile variants and or docker-compose variants such as more importantly a `docker-compose.prod.yml` example for people who want to run prod instances with a docker-compose file, which is common for selfhosters. But for the purposes for making the life easier for contributors and maintainers, I think it makes sense to have a root docker-compose.yml for that sake.

## Notes

Btw, the docs update is entirely AI tweaked, and it's not my intention to keep it that way.
If/When this PR is approved, I plan to revisit the docs and give it a human touch lol.
- [ ] docs/docs/07-development/01-setup.md
- [ ] docs/docs/02-installation/01-docker.md

### Duplicate PR?

btw, I am aware of https://github.com/karakeep-app/karakeep/pull/584, but it's opened a year ago and activity seems to have died down due to various reasons.

I took a stab at it and I believe I should have address the PR change requests and concerns

### DIFF
--- a/docker/docker-compose.dev.yml
+++ b/docker/docker-compose.dev.yml
@@ -44,6 +44,8 @@ services:
   chrome:
     image: gcr.io/zenika-hub/alpine-chrome:124
     restart: unless-stopped
+    ports:
+      - 9222:9222
     command:
       - --no-sandbox
       - --disable-gpu
--- a/docs/docs/08-development/01-setup.md
+++ b/docs/docs/08-development/01-setup.md
@@ -2,60 +2,44 @@
 
 ## Quick Start
 
-### Docker Compose
-
-1. (Optional) Copy the sample env file whenever you need to override the defaults or commit real secrets:
-
-```bash
-cp .env.sample .env
-```
-
-   - Even if you skip this step, the compose file injects sensible defaults: `DATA_DIR=/data`, `MEILI_ADDR=http://meilisearch:7700`, `NEXTAUTH_URL=http://localhost:3000`, and `NEXTAUTH_SECRET=dev-nextauth-secret`.
-   - When you're ready, set at least `NEXTAUTH_SECRET` (use `openssl rand -base64 36`) and any optional keys like `OPENAI_API_KEY` inside `.env` so they override the defaults.
-   - The stack automatically mounts `/data` inside the containers, so you only need to change `DATA_DIR` if you prefer a host path.
-2. Start every dependency, install packages, and run migrations in one go:
-
-```bash
-docker compose -f docker/docker-compose.dev.yml up
-```
-
-   - The `prep` service creates `DATA_DIR` (if missing), runs `pnpm install --frozen-lockfile`, and then `pnpm run db:migrate` the first time (and whenever you restart the stack) so the rest of the services always have dependencies ready.
-   - The `web` and `workers` services run `pnpm web` and `pnpm workers` respectively using the exact same code that lives on your host machine, so hot reload works out of the box.
-3. Tail the dev logs when needed:
-
-```bash
-docker compose logs -f web workers
-```
-
-4. Stop everything with `docker compose down`. Re-run with `--build` if you change Node dependencies or the Dockerfile.
-
-This setup exposes the usual endpoints:
-- Web app: http://localhost:3000
-- Meilisearch: http://localhost:7700
-- Chrome debugger: http://localhost:9222
-
-### Host script alternative
-
-If you prefer to keep pnpm running directly on your machine, you can still use the legacy helper script:
+For the fastest way to get started with development, use the one-command setup script:
 
 ```bash
 ./start-dev.sh
 ```
 
-It starts Meilisearch + Chrome in Docker, installs dependencies if needed, and then runs the web app and workers locally until you hit Ctrl+C.
+This script will automatically:
+- Start Meilisearch in Docker (on port 7700)
+- Start headless Chrome in Docker (on port 9222)
+- Install dependencies with `pnpm install` if needed
+- Start both the web app and workers in parallel
+- Provide cleanup when you stop with Ctrl+C
+
+**Prerequisites:**
+- Docker installed and running
+- pnpm installed (see manual setup below for installation instructions)
+
+The script will output the running services:
+- Web app: http://localhost:3000
+- Meilisearch: http://localhost:7700  
+- Chrome debugger: http://localhost:9222
+
+Press Ctrl+C to stop all services and clean up Docker containers.
+
+Alternatively, you can use Docker Compose to run the full stack in containers — see [Docker Compose Details](#docker-compose-details) below.
 
 ## Manual Setup
 
-Karakeep uses `node` version 22. To install it, you can use `nvm` [^1]
+Karakeep uses `node` version 24. To install it, you can use `nvm` [^1]
 
 ```
-$ nvm install  22
+$ nvm install 24
 ```
 
 Verify node version using this command:
 ```
 $ node --version
-v22.14.0
+v24.0.0
 ```
 
 Karakeep also makes use of `corepack`[^2]. If you have `node` installed, then `corepack` should already be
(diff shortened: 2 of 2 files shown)


MohamedBassem:
Ok, so I finally get to review this :) I love everything that this PR does except moving the docker-compose file to the root and removing the `.dev` suffix. I think this adds a bit of confusion on what the main docker compose is (and IMO, the main one is the prod env not the dev one).
I feel like we can get the entire experience as perfect as you've described it by just documenting the `-f docker/docker-compose.dev.yml` flag in the dev docs. If you're ok with that, I'm happy to do those changes.

xuatz:
> Ok, so I finally get to review this :) I love everything that this PR does except moving the docker-compose file to the root and removing the `.dev` suffix. I think this adds a bit of confusion on what the main docker compose is (and IMO, the main one is the prod env not the dev one).
> I feel like we can get the entire experience as perfect as you've described it by just documenting the `-f docker/docker-compose.dev.yml` flag in the dev docs. If you're ok with that, I'm happy to do those changes.

@MohamedBassem yeah I'm cool with your suggestions. if you're gonna do the changes, can I leave it you to make sure the docs changes makes sense too? I haven't vetted them so I don't sign off that they make sense lol

otherwise I'm okay to apply the changes instead and you can review again later

docker/docker-compose.dev.yml
```diff
@@ -30,45 +52,62 @@ services:
...
     working_dir: /app
     environment:
-      MEILI_ADDR: http://meilisearch:7700
-      BROWSER_WEB_URL: http://chrome:9222
-      DATA_DIR: /data
+      DATA_DIR: ${DATA_DIR:-/data}
+      MEILI_ADDR: ${MEILI_ADDR:-http://meilisearch:7700}
+      BROWSER_WEB_URL: ${BROWSER_WEB_URL:-http://chrome:9222}
+      NEXTAUTH_URL: ${NEXTAUTH_URL:-http://localhost:3000}
+      NEXTAUTH_SECRET: ${NEXTAUTH_SECRET:-super-secure-nextauth-secret}
```
MohamedBassem, line 79: technically, those shouldn't be needed here.

docker/docker-compose.dev.yml
```diff
@@ -30,45 +52,62 @@ services:
...
-    env_file:
-      - .env
+    restart: "no"
     working_dir: /app
     environment:
-      DATA_DIR: /data
+      DATA_DIR: ${DATA_DIR:-/data}
+      MEILI_ADDR: ${MEILI_ADDR:-http://meilisearch:7700}
+      NEXTAUTH_URL: ${NEXTAUTH_URL:-http://localhost:3000}
+      NEXTAUTH_SECRET: ${NEXTAUTH_SECRET:-super-secure-nextauth-secret}
```
MohamedBassem, line 100: same here

docs/docs/08-development/01-setup.md
```diff
@@ -2,50 +2,68 @@
...
+docker compose logs -f web workers
+```
+
+4. Stop everything with `docker compose down`. Re-run with `--build` if you change Node dependencies or the Dockerfile.
+
+This setup exposes the usual endpoints:
 - Web app: http://localhost:3000
-- Meilisearch: http://localhost:7700  
+- Meilisearch: http://localhost:7700
 - Chrome debugger: http://localhost:9222
```
MohamedBassem: This is kinda not true for the docker setup. It seems like you removed exposing the chrome port and meilisearch was never exposed in the first place, so those are slightly misleading.

docs/docs/08-development/01-setup.md
```diff
@@ -2,50 +2,68 @@
...
+```bash
+./start-dev.sh
+```
+
+It starts Meilisearch + Chrome in Docker, installs dependencies if needed, and then runs the web app and workers locally until you hit Ctrl+C.
 
 ## Manual Setup
 
-Karakeep uses `node` version 24. To install it, you can use `nvm` [^1]
+Karakeep uses `node` version 22. To install it, you can use `nvm` [^1]
```
MohamedBassem: is this a bad merge? we use node 24

docs/docs/08-development/01-setup.md
```diff
@@ -2,50 +2,68 @@
...
 ```
-$ nvm install  24
+$ nvm install  22
 ```
 
 Verify node version using this command:
 ```
 $ node --version
-v24.14.0
+v22.14.0
```
MohamedBassem: Same here

docs/docs/08-development/01-setup.md
```diff
@@ -2,50 +2,68 @@
 
 ## Quick Start
 
-For the fastest way to get started with development, use the one-command setup script:
+### Docker Compose
```
MohamedBassem: I'd leave the dev script as the quick start one, and move those details to the docker compose section?

docs/docs/08-development/01-setup.md
xuatz: yes, reverted!

docs/docs/08-development/01-setup.md
xuatz: yes, reverted!

docker/docker-compose.dev.yml
xuatz: i think the reason why i added NEXTAUTH_SECRET here, is so we can have a default `-super-secure-nextauth-secret`, where using the .env, we are forced to `openssl rand -base64 36` or put a dummy file

and since i did it for NEXTAUTH_SECRET, i did it for NEXTAUTH_URL as well

if you still don't think we should do it, i'll remove in the next commit

docs/docs/08-development/01-setup.md
xuatz: reexposed chrome port and removed meilisearch port from docs

docs/docs/08-development/01-setup.md
xuatz: applied comments

MohamedBassem:
Perfect, thanks for addressing the changes. Can you rebase (to resolve the conflicts) so that we can merge it?

MohamedBassem:
Thank you!

PR #2176, opened 2025-11-27
fix: lazy load js-tiktoken in prompts module

MohamedBassem:
The js-tiktoken module loads a large encoding dictionary into memory immediately on import. This change defers the loading of the encoding until it's actually needed by using a lazy getter pattern.

This reduces memory usage for processes that import this module but don't actually use the token encoding functions.


MohamedBassem:
@codex review

Issue #1578, opened 2025-06-09
FR: Disallow case sensitivity in tags


sprior:
### Describe the feature you'd like

When importing my bookmarks it appears that the AI will commonly suggest a tag in both all lowercase and with the first letter capitalized and karakeep seems to be creating tags for both.  While I'm sure there's someone out there who wants case sensitive tags it probably isn't what most people want.

I'd suggest making bookmark tagging use a case insensitive check to make sure that tag doesn't already exist before creating a new one and if it does use the existing one instead.

### Describe the benefits this would bring to existing Karakeep users

It'll avoid LOTs of tag clutter.

### Can the goal of this request already be achieved via other means?

not that I'm aware of

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

MohamedBassem:
This has been there since April last year, implemented in https://github.com/karakeep-app/karakeep/commit/b4c7de2ab00fb43a70c984138b79577a65c95191

Is it not working as expected? Do you have an example? Feel free to re-open the issue when you have a repro.

sprior:
Right now I have in my cleanup suggestions one to consolidate "S3", "s3", and "1960s".  Obviously the third one is a totally different thing but the first two differ only in the case of the S.  Note that "s3" is in the manual tags section and "S3" is in the AI tags section in case you're not checking across the two.
Ideally I'd like to migrate manual tags in favor of AI tags if possible.  I've only been using Karakeep for less than 2 weeks so no chance I've ever used a version before the change you mentioned.  I'm in the process of importing 20,000 bookmarks from Pocket using its CSV export.

It says I don't have permissions to reopen the issue.

sprior:
@MohamedBassem You mentioned reopening if I had an example and I do (see previous comment) but I don't have permissions to reopen.

xuatz:
helped to reopened for visibility, but still subjected to @MohamedBassem to triage 🙇

giacomocerquone:
So, by using ollama and generating tags through llama3.1:8b tags gets generated with several different casing (Accessory, accessory, etc).

This wouldn't be a problem for me unless the smart lists query languages would be case insensitive.
That for me is the deal braker when it comes to tags case sensitivity.

This is also to say that the problem should be tackled from a querying perspective and not from the ai generation which should always be considered unpredictable.

Btw I also tried adding the following custom prompt which did not do anything at all: `all tags must be lowercase`

[...]

PR #2187, opened 2025-11-29
fix: update tag mapping to return normalized tags instead of original tags in connectTags function

giacomocerquone:
See #1578 

UPDATE:

Tested on my machine, probably it's just better to do `t.originalTag.toLowerCase()` than getting the normalized one otherwise spaces will be removed etc. etc.

### DIFF
--- a/apps/workers/workers/inference/tagging.ts
+++ b/apps/workers/workers/inference/tagging.ts
@@ -346,7 +346,7 @@ async function connectTags(
               (mt) => normalizeTag(mt.name) === t.normalizedTag,
             ),
         )
-        .map((t) => t.normalizedTag);
+        .map((t) => t.originalTag.toLowerCase());
 
       return { matchedTagIds, notFoundTagNames };
     })();


giacomocerquone:
hey there @MohamedBassem this should be easy to merge right?

MohamedBassem:
Hey, thanks for opening a PR, I'll need to read the code around it to understand the bug a bit more clearly because last I tried, I couldn't repro it.

MohamedBassem:
Ok, I went and read the code. I don't think what this PR is doing, is correct. The change you introduced will force any new tag to be lower case basically and that might not be what everyone wants. New tags should be saved as they are generated from the LLM. The normalization is just to try to see if what the LLM generated is close to something that already existed. But it was never the intention to persist in the database the normalized format.

giacomocerquone:
> Ok, I went and read the code. I don't think what this PR is doing, is correct. The change you introduced will force any new tag to be lower case basically and that might not be what everyone wants. New tags should be saved as they are generated from the LLM. The normalization is just to try to see if what the LLM generated is close to something that already existed. But it was never the intention to persist in the database the normalized format.

got it, re-reading the first issue I linked I have to say I thought the objective was a different one... well this is a bit of a mess at is right now because slight variation of the same tag is still created and the llm aren't really as good as I'd like at keeping tags consistent

Thanks for the check on the pr  btw 🙏

PR #2239, opened 2025-12-10
Update 01-docker.md

kkrieger85:
Update command to improve installation step


MohamedBassem:
Thanks for the PR, though tbh, I'm more inclined toward the current way it's currently written as it makes it easier to just copy and paste.

PR #2245, opened 2025-12-10
feat: add Matter import

Moondragon85:
I've implemented an import module for [Matter](https://getmatter.com) exports.

I've tested it with my own Matter export of about 7000 items.

Unfortunately the Matter export doesn't include a “creation date” so I used the “last interaction date” instead.

Matter Import 1

Matter Import 2


MohamedBassem:
Thanks for the PR!

PR #2280, opened 2025-12-20
feat: add "URL Does Not Contain" condition to rule engine

### LINKED ISSUE #2259: Rule Engine: Add "URL DOES NOT Contain" Condition

Rule engine is a great feature. Allows me to automatically move saves with reddit.com or youtube.com URLs into their own specific lists. But what about everything else? I also have a "read later" list that I use just for actual articles. So these URLs could be from anywhere. The idea here is I could automatically move saves that don't fit some lists, into another list. So if I have 3 lists, say Read Later, Reddit and YouTube.. I could create a rule where if any of the following are true: URL does not contain reddit.com + URL does not contain youtube.com --> action: Add to Read Later list. That way if I save an article from New York Times, Gizmodo, whatever, it saves to my Read Later list. 

I think this would allow for even greater customization of lists and how one can save items, and would eliminate the need to create a bunch of domain specific rules. 

Maybe through webhooks? But seems like that would be way more complicated than a dropdown selection. 

### Additional context

_No response_

MohamedBassem:
Add a new condition type `urlDoesNotContain` that allows users to create rules based on URLs that do NOT contain specific strings.

This enables more flexible rule configurations, such as automatically adding bookmarks to a "Read Later" list if the URL does not contain "reddit.com" or "youtube.com".

Fixes #2259

### DIFF
--- a/packages/trpc/lib/ruleEngine.ts
+++ b/packages/trpc/lib/ruleEngine.ts
@@ -3,6 +3,7 @@ import { and, eq } from "drizzle-orm";
 
 import { bookmarks, tagsOnBookmarks } from "@karakeep/db/schema";
 import { LinkCrawlerQueue } from "@karakeep/shared-server";
+import { BookmarkTypes } from "@karakeep/shared/types/bookmarks";
 import {
   RuleEngineAction,
   RuleEngineCondition,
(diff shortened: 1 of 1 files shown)


MohamedBassem:
@codex review

PR #2302, opened 2025-12-25
feat: add synchronized reading progress for bookmarks

### LINKED ISSUE #1423: FR: Keeping track of reading progress of bookmarks

esimkowitz:
## Summary

Track and restore reading position in reader mode across web and mobile. Uses text character offsets with anchor text verification to find the first visible paragraph and scroll back to it on return.

Closes #1423

## Note on unsupported views

This PR only covers the reader view for now.

### Archived views

Adding support for archived views would require enabling JavaScript execution in the archive view iframes. While SingleFile, the new Karakeep extension, and the parsers all default to sanitizing JavaScript when saving archived HTML, there is still a chance someone archives something with malicious JS code in it. I didn't want to be the one to make this call, however I'm eager to figure out how we can add support for reading progress to the archived views while minimizing the risk to users.

### Live webpages - mobile

Adding support for live webpages in the mobile app should be very easy. Like with the reader view on mobile, this is encapsulated in a mobile webview so we can just inject a contentScript and the logic should be pretty much the same.

### Live webpages - web

For live webpage on the web, I think this would be a great feature for the Karakeep browser extension! It can host the same contentScript, with a lookup to see if a page is in your bookmarks. If so, it can autoscroll to your last known position!

## Architectural Decisions

**Three-Layer Code Organization**: Core utilities (`reading-progress-core.ts`) contain platform-agnostic logic shared between web and mobile. Platform-specific modules (`reading-progress-dom.ts`, `reading-progress-webview-src.ts`) extend the core with environment-specific implementations. Adds a generator script for transpiling the core logic into stringified JavaScript to inject into the Mobile WebView (and into iframes in the future).

**Dual-Strategy Position Restoration**: Uses anchor text (~50 chars of paragraph content) as primary lookup, with character offset as fallback. This makes restoration resilient to minor content changes while still working when anchor text isn't available.

**Text Normalization**: Collapses whitespace to single spaces before calculating offsets, ensuring consistent character counting regardless of HTML formatting variations.

**Content-Ready Signal with RAF Fallback**: Parent components signal when content is ready for restoration via the `contentReady` prop. The hook attempts restoration immediately when signaled, then uses a single `requestAnimationFrame` as a fallback if the layout hasn't been calculated yet. This avoids polling while ensuring restoration works even when the scroll parent needs one more paint cycle to have scrollable height.

**Persist on Unload**: Only persists changes to the DB when the view is unloaded or blurred (or when the app closes).

**Separate Reading Progress Table**: Reading progress is stored in a dedicated `userReadingProgress` table rather than as columns on the bookmarks table. This allows each user to track their own reading progress independently when viewing shared bookmarks.

**Platform-Specific Scrolling**: Web version handles nested scrolling (including Radix ScrollArea detection), while mobile assumes window-level scrolling only.

### Web - Full-Screen Reader
1. Open a long article bookmark
2. Click "Open full page" to enter reader mode
3. Scroll partway through the article
4. Close the tab or navigate away
5. Re-open the same bookmark in reader mode
6. Verify it scrolls to approximately where you left off

### Web - Preview Panel
1. Open sidebar preview for a long article
2. Scroll partway through in the cached content view
3. Close the preview
4. Re-open the same bookmark's preview
5. Verify position is restored

### Mobile
1. Open a long article in mobile reader view
2. Scroll partway through
3. Background the app (switch to another app)
4. Return to Karakeep
5. Verify position is maintained

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1379,9 +1379,6 @@ importers:
       '@types/nodemailer':
         specifier: ^6.4.17
         version: 6.4.17
-      esbuild:
-        specifier: ^0.27.2
-        version: 0.27.2
       vite-tsconfig-paths:
         specifier: ^4.3.1
         version: 4.3.2(typescript@5.9.3)(vite@7.0.6(@types/node@24.10.3)(jiti@2.4.2)(lightningcss@1.30.1)(sass@1.89.1)(terser@5.41.0)(tsx@4.20.3)(yaml@2.8.0))
@@ -3285,12 +3282,6 @@ packages:
     cpu: [ppc64]
     os: [aix]
 
-  '@esbuild/aix-ppc64@0.27.2':
-    resolution: {integrity: sha512-GZMB+a0mOMZs4MpDbj8RJp4cw+w1WV5NYD6xzgvzUJ5Ek2jerwfO2eADyI6ExDSUED+1X8aMbegahsJi+8mgpw==}
-    engines: {node: '>=18'}
-    cpu: [ppc64]
-    os: [aix]
-
   '@esbuild/android-arm64@0.18.20':
     resolution: {integrity: sha512-Nz4rJcchGDtENV0eMKUNa6L12zz2zBDXuhj/Vjh18zGqB44Bi7MBMSXjgunJgjRhCmKOjnPuZp4Mb6OKqtMHLQ==}
     engines: {node: '>=12'}
@@ -3299,12 +3290,6 @@ packages:
 
   '@esbuild/android-arm64@0.25.6':
     resolution: {integrity: sha512-hd5zdUarsK6strW+3Wxi5qWws+rJhCCbMiC9QZyzoxfk5uHRIE8T287giQxzVpEvCwuJ9Qjg6bEjcRJcgfLqoA==}
-    engines: {node: '>=18'}
-    cpu: [arm64]
-    os: [android]
-
-  '@esbuild/android-arm64@0.27.2':
-    resolution: {integrity: sha512-pvz8ZZ7ot/RBphf8fv60ljmaoydPU12VuXHImtAs0XhLLw+EXBi2BLe3OYSBslR4rryHvweW5gmkKFwTiFy6KA==}
     engines: {node: '>=18'}
     cpu: [arm64]
     os: [android]
@@ -3321,12 +3306,6 @@ packages:
     cpu: [arm]
     os: [android]
 
-  '@esbuild/android-arm@0.27.2':
-    resolution: {integrity: sha512-DVNI8jlPa7Ujbr1yjU2PfUSRtAUZPG9I1RwW4F4xFB1Imiu2on0ADiI/c3td+KmDtVKNbi+nffGDQMfcIMkwIA==}
-    engines: {node: '>=18'}
-    cpu: [arm]
-    os: [android]
-
   '@esbuild/android-x64@0.18.20':
     resolution: {integrity: sha512-8GDdlePJA8D6zlZYJV/jnrRAi6rOiNaCC/JclcXpB+KIuvfBN4owLtgzY2bsxnx666XjJx2kDPUmnTtR8qKQUg==}
     engines: {node: '>=12'}
@@ -3335,12 +3314,6 @@ packages:
 
   '@esbuild/android-x64@0.25.6':
     resolution: {integrity: sha512-0Z7KpHSr3VBIO9A/1wcT3NTy7EB4oNC4upJ5ye3R7taCc2GUdeynSLArnon5G8scPwaU866d3H4BCrE5xLW25A==}
-    engines: {node: '>=18'}
-    cpu: [x64]
-    os: [android]
-
-  '@esbuild/android-x64@0.27.2':
-    resolution: {integrity: sha512-z8Ank4Byh4TJJOh4wpz8g2vDy75zFL0TlZlkUkEwYXuPSgX8yzep596n6mT7905kA9uHZsf/o2OJZubl2l3M7A==}
     engines: {node: '>=18'}
     cpu: [x64]
     os: [android]
@@ -3357,12 +3330,6 @@ packages:
     cpu: [arm64]
     os: [darwin]
 
-  '@esbuild/darwin-arm64@0.27.2':
-    resolution: {integrity: sha512-davCD2Zc80nzDVRwXTcQP/28fiJbcOwvdolL0sOiOsbwBa72kegmVU0Wrh1MYrbuCL98Omp5dVhQFWRKR2ZAlg==}
-    engines: {node: '>=18'}
-    cpu: [arm64]
-    os: [darwin]
-
   '@esbuild/darwin-x64@0.18.20':
     resolution: {integrity: sha512-pc5gxlMDxzm513qPGbCbDukOdsGtKhfxD1zJKXjCCcU7ju50O7MeAZ8c4krSJcOIJGFR+qx21yMMVYwiQvyTyQ==}
     engines: {node: '>=12'}
@@ -3371,12 +3338,6 @@ packages:
 
   '@esbuild/darwin-x64@0.25.6':
     resolution: {integrity: sha512-GfXs5kry/TkGM2vKqK2oyiLFygJRqKVhawu3+DOCk7OxLy/6jYkWXhlHwOoTb0WqGnWGAS7sooxbZowy+pK9Yg==}
-    engines: {node: '>=18'}
-    cpu: [x64]
-    os: [darwin]
-
-  '@esbuild/darwin-x64@0.27.2':
-    resolution: {integrity: sha512-ZxtijOmlQCBWGwbVmwOF/UCzuGIbUkqB1faQRf5akQmxRJ1ujusWsb3CVfk/9iZKr2L5SMU5wPBi1UWbvL+VQA==}
     engines: {node: '>=18'}
     cpu: [x64]
     os: [darwin]
@@ -3393,12 +3354,6 @@ packages:
     cpu: [arm64]
     os: [freebsd]
 
-  '@esbuild/freebsd-arm64@0.27.2':
-    resolution: {integrity: sha512-lS/9CN+rgqQ9czogxlMcBMGd+l8Q3Nj1MFQwBZJyoEKI50XGxwuzznYdwcav6lpOGv5BqaZXqvBSiB/kJ5op+g==}
-    engines: {node: '>=18'}
-    cpu: [arm64]
-    os: [freebsd]
-
   '@esbuild/freebsd-x64@0.18.20':
     resolution: {integrity: sha512-tgWRPPuQsd3RmBZwarGVHZQvtzfEBOreNuxEMKFcd5DaDn2PbBxfwLcj4+aenoh7ctXcbXmOQIn8HI6mCSw5MQ==}
     engines: {node: '>=12'}
@@ -3407,12 +3362,6 @@ packages:
 
   '@esbuild/freebsd-x64@0.25.6':
     resolution: {integrity: sha512-2SkqTjTSo2dYi/jzFbU9Plt1vk0+nNg8YC8rOXXea+iA3hfNJWebKYPs3xnOUf9+ZWhKAaxnQNUf2X9LOpeiMQ==}
-    engines: {node: '>=18'}
-    cpu: [x64]
-    os: [freebsd]
-
-  '@esbuild/freebsd-x64@0.27.2':
-    resolution: {integrity: sha512-tAfqtNYb4YgPnJlEFu4c212HYjQWSO/w/h/lQaBK7RbwGIkBOuNKQI9tqWzx7Wtp7bTPaGC6MJvWI608P3wXYA==}
     engines: {node: '>=18'}
     cpu: [x64]
     os: [freebsd]
@@ -3429,12 +3378,6 @@ packages:
     cpu: [arm64]
     os: [linux]
 
-  '@esbuild/linux-arm64@0.27.2':
-    resolution: {integrity: sha512-hYxN8pr66NsCCiRFkHUAsxylNOcAQaxSSkHMMjcpx0si13t1LHFphxJZUiGwojB1a/Hd5OiPIqDdXONia6bhTw==}
-    engines: {node: '>=18'}
-    cpu: [arm64]
-    os: [linux]
-
   '@esbuild/linux-arm@0.18.20':
     resolution: {integrity: sha512-/5bHkMWnq1EgKr1V+Ybz3s1hWXok7mDFUMQ4cG10AfW3wL02PSZi5kFpYKrptDsgb2WAJIvRcDm+qIvXf/apvg==}
     engines: {node: '>=12'}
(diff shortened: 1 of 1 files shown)


packages/trpc/routers/bookmarks.ts
```diff
@@ -583,6 +584,24 @@ export const bookmarksAppRouter = router({
...
+    .input(
+      z.object({
+        bookmarkId: z.string(),
+        readingProgressOffset: z.number().int().nonnegative(),
+        readingProgressAnchor: z.string().max(ANCHOR_TEXT_MAX_LENGTH).nullish(),
+      }),
+    )
+    .use(ensureBookmarkOwnership)
+    .mutation(async ({ input, ctx }) => {
+      await ctx.db
```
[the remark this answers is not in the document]
esimkowitz: this is fine and expected

packages/shared-react/hooks/reading-progress.ts
```diff
@@ -0,0 +1,449 @@
...
+  // Update reading position on scroll
+  // Store reference to scroll parent for cleanup
+  const scrollParentRef = useRef<HTMLElement | Window | null>(null);
+
+  useEffect(() => {
+    if (!enabled || typeof window === "undefined") {
+      return;
+    }
+
+    const handleScroll = () => {
```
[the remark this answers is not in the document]
esimkowitz: addressing

packages/shared/scripts/build-webview-js.ts
```diff
@@ -0,0 +1,124 @@
...
+}
+
+async function watch() {
+  console.log(`Watching ${SRC_FILE} for changes...`);
+
+  // Initial build
+  await build();
+
+  // Watch for changes
+  fs.watch(SRC_FILE, async (eventType) => {
```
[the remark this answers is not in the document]
esimkowitz: addressing

apps/mobile/components/bookmarks/BookmarkLinkPreview.tsx
```diff
@@ -51,6 +122,67 @@ export function BookmarkLinkReaderPreview({
...
+        apiUtils.bookmarks.getBookmark.invalidate({ bookmarkId: bookmark.id });
+      },
+      onError: (error) => {
+        console.error("[ReadingProgress] Failed to save progress:", error);
+      },
+    });
+
+  // Save progress function
+  const saveProgress = useCallback(() => {
+    if (currentPosition.current === null) return;
```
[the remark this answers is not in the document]
esimkowitz: description updated

apps/web/components/dashboard/preview/LinkContentSection.tsx
```diff
@@ -124,6 +129,19 @@ export default function LinkContentSection({
...
+  const initialOffset = bookmark.content.readingProgressOffset;
+  const initialAnchor = bookmark.content.readingProgressAnchor;
+
+  // Auto-save reading progress on visibility change/section change
+  useReadingProgressAutoSave({
+    bookmarkId: bookmark.id,
+    initialOffset,
+    initialAnchor,
+    containerRef: contentRef,
+    enabled: section === "cached", // Only track in cached/reader view
```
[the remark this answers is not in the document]
esimkowitz: description updated

esimkowitz:
Moving back to draft mode, I found some ways to improve the performance and reliability of this feature

esimkowitz:
I ended up deciding not to over-optimize things too early, I have some follow-up PRs I want to add (like live webpage progress tracking/restoring) but I'll address any changes to the reading progress tracking I need to make then. I did add percentage tracking, though, in advance of adding rules engine support.

j4dk:
@esimkowitz hi! was looking for similar functionality of karakeep and found your PR :) so just wanted to drop a word of encouragement and say thank you and great job. Looking forward to this feature!
Thanks to everyone involved with karakeep in general - its absolutely great.

packages/db/schema.ts
```diff
@@ -357,6 +357,33 @@ export const highlights = sqliteTable(
...
+  {
+    id: text("id")
+      .notNull()
+      .primaryKey()
+      .$defaultFn(() => createId()),
+    bookmarkId: text("bookmarkId")
+      .notNull()
+      .references(() => bookmarks.id, {
+        onDelete: "cascade",
+      }),
```
MohamedBassem, line 372: If we're storing reading progress for shared bookmarks, it means that those records will dangle if the user loses access to those bookmarks. Should we also add a foreign key here on the list membership (with onCascade delete) such that when a user loses access to a list, their reading progress is wiped?

packages/trpc/routers/bookmarks.ts
```diff
@@ -593,9 +636,30 @@ export const bookmarksAppRouter = router({
...
+        });
+        zBookmark.content.readingProgressOffset =
+          progress?.readingProgressOffset ?? null;
+        zBookmark.content.readingProgressAnchor =
+          progress?.readingProgressAnchor ?? null;
+        zBookmark.content.readingProgressPercent =
+          progress?.readingProgressPercent ?? null;
+      }
+
+      return zBookmark;
```
MohamedBassem: I'd prefer if this is a separate route and not part of the main `ZBookmark`. `getBookmark` is a hot endpoint that gets called for every element in the bookmark grid. So don't want to increase its latency by doing another round trip to the db for some data that might not end up being used. It's also aligned with having the `updateReadingProgress` route separate.

packages/trpc/routers/bookmarks.test.ts
```diff
@@ -1001,4 +1001,307 @@ describe("Bookmark Routes", () => {
...
+      assert(retrieved.content.type === BookmarkTypes.LINK);
+      expect(retrieved.content.readingProgressOffset).toBe(500);
+
+      // Delete the bookmark
+      await api.deleteBookmark({ bookmarkId: bookmark.id });
+
+      // Verify bookmark is gone (and implicitly, the progress cascade deleted)
+      await expect(() =>
+        api.getBookmark({ bookmarkId: bookmark.id }),
+      ).rejects.toThrow(/Bookmark not found/);
```
MohamedBassem, line 1237: That fact that we're asserting that it's deleted by calling `getBookmark` makes this a not so useful test. However, it's relevant if you end up splitting the route as I mentioned in another comment.

packages/trpc/models/bookmarks.ts
```diff
@@ -591,6 +595,10 @@ export class Bookmark extends BareBookmark {
               publisher: row.bookmarkLinks.publisher,
               datePublished: row.bookmarkLinks.datePublished,
               dateModified: row.bookmarkLinks.dateModified,
+              // Reading progress is fetched separately per-user in getBookmark
+              readingProgressOffset: null,
+              readingProgressAnchor: null,
+              readingProgressPercent: null,
```
MohamedBassem: That's another reason why they probably shouldn't be part of `ZBookmark` and have their own route.

packages/shared/package.json
```diff
@@ -22,17 +22,20 @@
...
     "@types/html-to-text": "^9.0.4",
     "@types/nodemailer": "^6.4.17",
+    "esbuild": "^0.27.2",
     "vite-tsconfig-paths": "^4.3.1",
     "vitest": "^3.2.4"
   },
   "scripts": {
+    "generate:webview-js": "tsx scripts/build-webview-js.ts",
+    "generate:webview-js:watch": "tsx scripts/build-webview-js.ts --watch",
+    "check:webview-js": "tsx scripts/build-webview-js.ts --check",
```
MohamedBassem: How about we move the script to the `tools/` directory instead of the `shared` package?

packages/shared/package.json
```diff
@@ -22,17 +22,20 @@
...
     "typecheck": "tsc --noEmit",
     "format": "prettier . --cache --ignore-path ../../.prettierignore --check",
     "format:fix": "prettier . --cache --write --ignore-path ../../.prettierignore",
     "lint": "oxlint .",
     "lint:fix": "oxlint . --fix",
     "test": "vitest"
   },
-  "main": "index.ts",
```
MohamedBassem, line 36: Is this accidental?

packages/shared/utils/reading-progress-core.ts
```diff
@@ -0,0 +1,263 @@
...
+  let topParagraph: Element | null = null;
+
+  for (const paragraph of paragraphs) {
+    const rect = paragraph.getBoundingClientRect();
+
+    // If this paragraph's top is at or below the viewport top, it's our target
+    if (rect.top >= viewportTop) {
+      topParagraph = paragraph;
+      break;
+    }
```
MohamedBassem: Is there an assumption here that the paragraphs are going to be sorted in asc order by their top position? Is this a safe assumption? Would it be safer if we sort them first?

packages/shared/utils/reading-progress-core.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  test("returns 0 for whitespace-only string", () => {
+    expect(normalizeTextLength("   \n\t")).toBe(0);
+  });
+});
+
+describe("ANCHOR_TEXT_MAX_LENGTH", () => {
+  test("is 50 characters", () => {
+    expect(ANCHOR_TEXT_MAX_LENGTH).toBe(50);
+  });
+});
```
MohamedBassem: Is this test really useful?

packages/shared-react/hooks/reading-progress.ts
```diff
@@ -0,0 +1,230 @@
+"use client";
+
+import { useCallback, useEffect, useRef, useState } from "react";
+
+import type { ReadingPosition } from "@karakeep/shared/utils/reading-progress-dom";
+import { SCROLL_THROTTLE_MS } from "@karakeep/shared/utils/reading-progress-core";
```
MohamedBassem: Should both `reading-progress-core` and `zreading-progress-dom` get moved to `shared-react`?

packages/shared-react/hooks/reading-progress.ts
```diff
@@ -0,0 +1,230 @@
...
+  // Also checks element visibility (may be hidden via CSS in inactive tab)
+  useEffect(() => {
+    if (!enabled || !contentReady || typeof window === "undefined") return;
+
+    const container = containerRef.current;
+    if (!container || !isElementVisible(container)) return;
+
+    const saveCurrentProgress = () => {
+      let positionToSave = lastKnownPositionRef.current;
+      const freshPosition = getReadingPosition(container);
```
MohamedBassem: It's not clear to me what's the point of `lastKnownPositionRef` if you're always fetching a fresh position on save? Maybe add a comment?

packages/shared-react/hooks/reading-progress.ts
```diff
@@ -0,0 +1,230 @@
...
+      if (freshPosition !== null && freshPosition.offset > 0) {
+        positionToSave = freshPosition;
+      }
+      if (positionToSave !== null && positionToSave.offset > 0) {
+        savePositionRef.current(positionToSave);
```
MohamedBassem: noob question: is there a difference between `current(...)` and `current = ...`? You seem to be mixing both.

packages/db/schema.ts
esimkowitz: I was following the pattern in the highlights table when setting this one up, would you like me to add a similar dangling entry deletion for that table if a user loses access to a highlighted bookmark?

MohamedBassem: Looking at the highlights router, the middleware enforces bookmark ownership. So users can't highlight content on bookmarks they don't own so it shouldn't suffer from the same problem. I think this is first metadata that we're attaching to bookmarks by collaborators not owners

esimkowitz: A foreign key for the list id could get messy if the user has the same bookmark shared with them in multiple shared lists. It seems to me like this could be handled purely on the backend by adding checks in `removeCollaborator` and `leaveList` to check if the user still has access to the bookmark via other means before deleting the entry. What do you think?

packages/shared/package.json
esimkowitz: ah, this was a linting warning, shared package didn't have an index.ts entrypoint so my linter was complaining to me. I guess Claude decided to delete the entry as a remediation, I can add this back

packages/shared/utils/reading-progress-core.ts
esimkowitz: According to [StackOverflow](https://stackoverflow.com/questions/8203770/using-queryselectorall-is-the-result-returned-by-the-method-ordered), the result from querySelectorAll is ordered in document order so sorting shouldn't be necessary

packages/shared/utils/reading-progress-core.test.ts
esimkowitz: lol, no claude must have added that, I'll remove

packages/shared-react/hooks/reading-progress.ts
esimkowitz: I had put them in shared utils because they didn't have any react code, I can move them though if you'd prefer them in shared-react

packages/shared-react/hooks/reading-progress.ts
esimkowitz: I don't fully remember why I added this, I think it was an attempt at caching the position so I didn't need to fetch it every time but it appears unnecessary now. I'll remove it

packages/shared/package.json
esimkowitz: should i move the script definitions to the main package.json too? Otherwise they'll be referencing `../../tools/build-webview-js.ts`

packages/shared-react/hooks/reading-progress.ts
esimkowitz: ah, seems like a claudeism, this is a ref wrapping a useCallback. that's silly, I'll fix it

esimkowitz: Oh wait never mind I remember why i did this, we need a stable ref so that the event handlers don't need to tear down and rebuild every time savePosition changes. In this case, .current is not being updated, it's a function being referenced. This was intentional, though a bit wonky. I can update the comment to clarify why we're doing this

MohamedBassem:
@esimkowitz I hope you don't mind but I did a big refactoring to this PR motivated by the fact that I unified the `ReaderView` between the mobile and the web using expo's `use dom`. This helped deliver quite a bit of simplification. The changes I introduced are:
1. Dropped esbuild and script injection and everything specific to the `Webview` now that it's being taken care of by expo's `use dom`.
2. Instead of hiding content until position is restored, I found it better UX to show a banner giving the user the option to scroll back to position instead of automatically scrolling them ourseleves.
3. I'm now showing a progress bar at the top of reading views so that the user are aware of what progress are we tracking.
4. Split the reader progress into its own trpc handler.

MohamedBassem:
@greptile review

MohamedBassem:
Thanks a lot @esimkowitz, you've put a great amount of effort in this PR and I love some of the ideas in here. Sorry this took so long!

Issue #2271, opened 2025-12-19
Allow list editing on mobile


colado:
### Describe the feature you'd like

User should be able to edit list name after list has been created.

### Describe the benefits this would bring to existing Karakeep users

Feature is already available on web.

### Can the goal of this request already be achieved via other means?

By using web app.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

### Additional context

Could be done inside the ListView of List screen, above Delete List.

![Image](https://github.com/user-attachments/assets/30d7ddc2-c2e1-469b-9e63-6a7cf1ff5e9d)

PR #2310, opened 2025-12-27
feat(mobile): create new edit screen and path

colado:
Adds edit list feature to mobile.

Closes https://github.com/karakeep-app/karakeep/issues/2271

### DIFF
--- a/apps/mobile/app/dashboard/lists/[slug]/edit.tsx
+++ b/apps/mobile/app/dashboard/lists/[slug]/edit.tsx
@@ -88,20 +88,22 @@ const EditListPage = () => {
           <View className="gap-2">
             <Text className="text-sm text-muted-foreground">List Type</Text>
             <View className="flex flex-row gap-2">
-              <Button
-                variant={list?.type === "manual" ? "primary" : "secondary"}
-                className="flex-1"
-                disabled
-              >
-                <Text>Manual</Text>
-              </Button>
-              <Button
-                variant={list?.type === "smart" ? "primary" : "secondary"}
-                className="flex-1"
-                disabled
-              >
-                <Text>Smart</Text>
-              </Button>
+              <View className="flex-1">
+                <Button
+                  variant={list?.type === "manual" ? "primary" : "secondary"}
+                  disabled
+                >
+                  <Text>Manual</Text>
+                </Button>
+              </View>
+              <View className="flex-1">
+                <Button
+                  variant={list?.type === "smart" ? "primary" : "secondary"}
+                  disabled
+                >
+                  <Text>Smart</Text>
+                </Button>
+              </View>
             </View>
           </View>
 


apps/mobile/app/dashboard/lists/[slug]/edit.tsx
```diff
@@ -0,0 +1,113 @@
...
+import CustomSafeAreaView from "@/components/ui/CustomSafeAreaView";
+import { Input } from "@/components/ui/Input";
+import { Text } from "@/components/ui/Text";
+import { useToast } from "@/components/ui/Toast";
+import { api } from "@/lib/trpc";
+
+import { useEditBookmarkList } from "@karakeep/shared-react/hooks/lists";
+
+// TODO Duplicated from new -- import from somewhere
+type ListType = "manual" | "smart";
```
Comment (MohamedBassem) [inline apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: I think this can be
```suggestion
type ListType = ZBookmarkList["type"];
```

apps/mobile/app/dashboard/lists/[slug]/edit.tsx
```diff
@@ -0,0 +1,113 @@
+import { useState } from "react";
+import { View } from "react-native";
+import { useLocalSearchParams } from "expo-router";
+import { dismiss } from "expo-router/build/global-state/routing";
```
Comment (MohamedBassem) [inline apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: This seems like a slightly weird import. We probably should use `useRouter` (https://docs.expo.dev/router/advanced/stack/#dismiss-action)?

apps/mobile/app/dashboard/lists/[slug]/edit.tsx
```diff
@@ -0,0 +1,113 @@
...
+import { api } from "@/lib/trpc";
+
+import { useEditBookmarkList } from "@karakeep/shared-react/hooks/lists";
+
+// TODO Duplicated from new -- import from somewhere
+type ListType = "manual" | "smart";
+
+const EditListPage = () => {
+  const [text, setText] = useState("");
+  const [listType, setListType] = useState<ListType>("manual");
```
Comment (MohamedBassem) [inline apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: List types are typically not editable. So this probably shouldn't be a state but instead read directly form `list.type`

apps/mobile/app/dashboard/lists/[slug]/edit.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const [query, setQuery] = useState("");
+  const { listId } = useLocalSearchParams();
+  const { data: list } = api.lists.get.useQuery({ listId: listId as string });
+  const { mutate, isPending } = useEditBookmarkList({
```
Comment (MohamedBassem) [inline apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: We might want to show a `FullPageSpinner` if `isPending` is true

apps/mobile/app/dashboard/lists/[slug]/edit.tsx
Comment (colado) [reply apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: True, auto-imported by mistake. Changed

apps/mobile/app/dashboard/lists/[slug]/edit.tsx
Comment (colado) [reply apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: Removed, that input is disabled so no state.

apps/mobile/app/dashboard/lists/[slug]/edit.tsx
Comment (colado) [reply apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: No longer needed. See below.

apps/mobile/app/dashboard/lists/[slug]/edit.tsx
Comment (colado) [reply apps/mobile/app/dashboard/lists/[slug]/edit.tsx]: Thanks, added this too. Spinner will be visible while waiting for response, i.e. between submit and server errors/success. Wondering if new.tsx should have the same behaviour?

colado:
Note for testing: toast notifications are currently not visible inside modals on iOS. I opened a separate issue for this: https://github.com/karakeep-app/karakeep/issues/2314

MohamedBassem:
Thanks @colado!

PR #2324, opened 2025-12-30
Use the Ollama generate endpoint instead of chat

eriktews:
Ollama has two API endpoints for text generation. There is a chat endpoint for interactive and interactive chat like generation of text and there is a generate endpoint that is used one one-shot prompts, such as summarization tasks and similar things.

Karakeep used the chat endpoint that resulted in odd summaries. This commit makes karakeep use the generate endpoint instead, which results in better and more compact summaries.


MohamedBassem:
@codex review

PR #2332, opened 2026-01-02
fix(mobile): fix Save button not working in Server Address screen

lucamartinetti:
## Summary
- Fix Save button in Edit Server Address screen not responding to taps on Android
- The issue was caused by `headerTransparent: true` being inherited from the default screen options, which breaks touch handling for header buttons
- Setting `headerTransparent: false` explicitly fixes the issue

## Test plan
- [x] Open the mobile app and navigate to Settings > Server Address
- [x] Verify the Save button in the header now responds to taps
- [x] Verify changes are saved correctly when pressing Save

### DIFF
--- a/apps/mobile/app/server-address.tsx
+++ b/apps/mobile/app/server-address.tsx
@@ -94,6 +94,7 @@ export default function ServerAddress() {
       <Stack.Screen
         options={{
           title: "Server Address",
+          headerTransparent: false,
           headerRight: () => (
             <Pressable onPress={handleSave}>
               <Text className="text-base font-semibold text-blue-500">


MohamedBassem:
Thanks for reporting the issue and suggesting a fix, however this seems like it adds a large padding above the header as well.

MohamedBassem:
This turned out to be more problematic than expected as it also breaks the looks on ios. I've fixed it in a different way: https://github.com/karakeep-app/karakeep/commit/0e1d28a5951f6ae55f05e87b161f54be55cb71d2 thanks a lot for the PR though.

PR #2340, opened 2026-01-03
fix: stop theme flashes with cloudflare rocket loader

sweepies:
Fixes light mode flashes on page load when using Cloudflare Rocket Loader as per the next-theme docs: https://github.com/pacocoursey/next-themes#using-with-cloudflare-rocket-loader


MohamedBassem:
Thanks a lot for taking the time to investigate and fix it!

PR #2353, opened 2026-01-06
fix:  Accept more permissive RSS feed content types and Fix User-Agent key

ElectricTea:
Some feeds are Content-Type application/xml only and will respond with a 406 error to responses with a header of content type application/rss+xml. This change allows for the more permissive content types application/xml and text/xml to be accepted.

Also fixes "UserAgent" header key with correct "User-Agent".


MohamedBassem:
Thanks for the PR! Good catch with the `User-Agent` header as well!

Issue #1670, opened 2025-06-24
Additional condition in Rule Engine "If title contains"


piotrkustal:
### Describe the feature you'd like

I'd like to be able to assign tags to bookmarks in case when specific keyword in "Title" is identified.

### Describe the benefits this would bring to existing Karakeep users

Add operator "Title has specific keyword" in IF conditions for Rule Engine Module

### Can the goal of this request already be achieved via other means?

Partially by URL detection but doesn't work with bookmarks saved from Facebook or YouTube as URLS have generic IDs

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

mokhovyk:
@MohamedBassem could you review the solution https://github.com/karakeep-app/karakeep/pull/2354?

MohamedBassem:
Merged in #2354 thanks to @mokhovyk

PR #2354, opened 2026-01-06
feat(rules): add "Title Contains" condition to Rule Engine (#1670)

mokhovyk:
Adds two new conditions to the Rule Engine:
-  **"Title Contains"**
-  **"Title Does Not Contain"**

This allows users to create automation rules based on bookmark titles. Solved #1670

### DIFF
--- a/packages/trpc/lib/ruleEngine.ts
+++ b/packages/trpc/lib/ruleEngine.ts
@@ -22,6 +22,7 @@ async function fetchBookmark(db: AuthedContext["db"], bookmarkId: string) {
       link: {
         columns: {
           url: true,
+          title: true,
         },
       },
       text: true,


MohamedBassem:
Hey, thanks for opening the PR! Although the logic itself is correct, `title contains` is tricky though, because the rule engine is evaluated when the bookmark is "created" but it usually gets a title when its "crawled". So this rule will only work for bookmarks created with the API and explicitly setting a title. Or with other triggers like "tagging", or lists, etc. So we might need to hide this "action" in the "bookmark created" condition otherwise it'll be confusing to people.

packages/trpc/lib/ruleEngine.ts
```diff
@@ -90,6 +90,12 @@ export class RuleEngine {
           !(this.bookmark.link?.url ?? "").includes(condition.str)
         );
       }
+      case "titleContains": {
+        return (this.bookmark.title ?? "").includes(condition.str);
```
MohamedBassem: crawled bookmarks has their links in `this.bookmark.content.title` (if they're links) with no user overrides. So this might need to be something like:

```
this.bookmark.title ?? (this bookmark.type == BookmarkTypes.LINK ? this.bookmark.content.title : "")
```

mokhovyk: thanks, fixed

mokhovyk:
> Hey, thanks for opening the PR! Although the logic itself is correct, `title contains` is tricky though, because the rule engine is evaluated when the bookmark is "created" but it usually gets a title when its "crawled". So this rule will only work for bookmarks created with the API and explicitly setting a title. Or with other triggers like "tagging", or lists, etc. So we might need to hide this "action" in the "bookmark created" condition otherwise it'll be confusing to people.

Thanks for the clarification. I’ve updated the PR to hide this action under the “bookmark created” condition to avoid confusion

MohamedBassem:
Thank you!

PR #2356, opened 2026-01-06
fix: Add JSON instruction to default text tagging prompt before content insertion

ElectricTea:
Added the prompt instruction
```
- You must respond in valid JSON with the key "tags" and the value is list of tags. Don't wrap the response in a markdown code.
```
to the default text tagging instructions before the content insertion. This change significantly improves the success rate of a response containing structured JSON when a prompt is truncated by the LLM due to the prompt exceeding a maximum token limit.

I kept the original JSON instruction at the end of the prompt because it "reminds" the LLM to use JSON structure after the content insertion, and it causes no issues.


ElectricTea:
Woops, I added it to the incorrect function :facepalm: 

Sorry about that. Will close this PR and open another to keep the commit history clean.

PR #2366, opened 2026-01-10
Autofocus on note open

SnowSquire:
qol feature for chrome extension


MohamedBassem:
Hey, thanks for the PR. As someone who doesn't usually attach notes to their bookmarks, I'm not sure yet if that's the behavior we should adopt in the plugin (to steal the focus).

SnowSquire:
To get the pop up to go away, you have to click away (either with mouse or ESC) anyways. So I don't think this adds any actually friction.

MohamedBassem:
Ok, thank you!

PR #2380, opened 2026-01-12
Fix relative links in Docker installation guide

chrismessina:
Relative links are 404s on this page, so I updated them to use `../../`.

### DIFF
--- a/docs/versioned_docs/version-v0.30.0/02-installation/01-docker.md
+++ b/docs/versioned_docs/version-v0.30.0/02-installation/01-docker.md
@@ -44,7 +44,7 @@ Persistent storage and the wiring between the different services is already take
 
 Keep in mind that every time you change the `.env` file, you'll need to re-run `docker compose up`.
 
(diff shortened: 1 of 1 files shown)


MohamedBassem:
I don't think your fixes work either (the bots are right). Seems like some quirk with the relative URLs with docusaurus. And beyond only this page. This fixed it (https://github.com/karakeep-app/karakeep/commit/4858d3faaf9cf91bac93c20c9723d2cc868c99b4). Thanks for the report and attempt though!

chrismessina:
> I don't think your fixes work either (the bots are right). Seems like some quirk with the relative URLs with docusaurus. And beyond only this page. This fixed it ([4858d3f](https://github.com/karakeep-app/karakeep/commit/4858d3faaf9cf91bac93c20c9723d2cc868c99b4)). Thanks for the report and attempt though!

Alright! As long as it's fixed, I care less about how it happened! :)

PR #2390, opened 2026-01-13
Fix memory leak in in-memory rate limiter

Mxrk:
The rate limiter used a probabilistic cleanup (1% chance per request) which allowed the internal Map to grow unbounded over time, causing continuous memory growth until container restart.

- Replace manual Map with lru-cache (as we're also using that in another package) for automatic LRU eviction and bounded size (50k entries)
- Eliminates O(n) cleanup iterations that could block the event loop

### DIFF
--- a/packages/plugins/package.json
+++ b/packages/plugins/package.json
@@ -23,6 +23,7 @@
     "@restatedev/restate-sdk": "^1.9.0",
     "@restatedev/restate-sdk-clients": "^1.9.0",
     "liteque": "^0.7.0",
+    "lru-cache": "^11.2.2",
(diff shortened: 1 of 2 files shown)


MohamedBassem:
Honestly, not convinced that the ratelimited is what's causing this. The cloud is being hit by various different IPs and doesn't have the same problem. Also, not sure if using an LRU cache for ratelimiting as it seems like it's something you can get around by overloading the cache. As for the `O(n)`, this should be rare, and even a million iteration seems unlikely to block the event loop for any meaningful period. So I think let's keep it as is for now.

Issue #2372, opened 2026-01-11
Grid layout has horizontal scroll bars


evfrenkel:
### Describe the Bug

At various window widths, the main grid has a horizontal scroll bar. Not sure if it's intended, if so consider this a feature request. I'm happy to help code a PR if this is recognized.

https://github.com/user-attachments/assets/0f40662b-d919-4cea-81f2-7745294c6dff

### Steps to Reproduce

1. Move the window width until you notice a scroll bar.

### Expected Behaviour

The grid should resize / reflow to fit the width.

### Device Details

Safari and Chrome

### Exact Karakeep Version

0.30

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

MohamedBassem:
This is not intended. Thanks for reporting. Please send a PR if you can!

evan6seven:
Added PR

MohamedBassem:
Sovled by #2400, thank @evan6seven for reporting and fixing it

PR #2400, opened 2026-01-14
#2372 avoid masonry overflow in bookmark grids

evan6seven:
Solves #2372

Simple change. Tested all the masonry column options on my machine.

Swapped the masonry gutter implementation to use column padding instead of flex gaps so the react-masonry-css column widths (100/n) no longer add extra horizontal space, which removes the scrollbar at 4 columns. The overflow came from className="flex gap-4" on Masonry while the library already sets column widths by percentage; the gap adds (cols - 1) * gap to the total width.

(I had to recreate this PR because I changed the source branch)


MohamedBassem:
That's great, thank you!

Issue #2172, opened 2025-11-25
Uncertain origin of alpine chrome image


MemQu:
### Describe the Bug

This was pointed out earlier:

https://github.com/karakeep-app/karakeep/issues/1241#issuecomment-2818253924

We have confirmation from the owner of the repo:

https://github.com/jlandure/alpine-chrome/issues/272#issuecomment-3536582057

https://github.com/jlandure/alpine-chrome/blob/master/README.md

Noticed it when sec-comp broke with the new image.

### Steps to Reproduce

None

### Expected Behaviour

Using new open source alpine chrome image

### Exact Karakeep Version

NA

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

xuatz:
Thank you for sharing the update from the original maintainer.

How about https://hub.docker.com/r/chromedp/headless-shell? @MemQu @MohamedBassem 

I think I'll start by trying it out on my own instance

edit1: mmm i'm having problems getting it to work lol, maybe i will try ghcr.io/browserless/chromium instead

edit2: okay, got ghcr.io/browserless/chromium working

Ra72xx:
The image docker.io/chromedp/headless-shell:latest seems to work without any further parameters.

MohamedBassem:
Thanks for flagging the issue. I know that browserless container works (because it's what people use on unraid) and is geared toward crawling, so it sounds like a good fit. Only problem with it is that last time I tried it, the container was quite big (1GB+), but maybe that's fine?

MohamedBassem:
One problem with browserless is that it has some non conventional license, which adds caveat to people who are using karakeep in commercial environments.

MattTW:
When scanning this image with a security tool like Grype, this container also comes back with lots of critical security vulnerabilities including many being actively leveraged in the wild. I don't know if any of these are applicable vectors for Karakeep or typical homelab installations, but maybe this is fuel to bump on the priority on replacing this.

<img width="1424" height="1013" alt="Image" src="https://github.com/user-attachments/assets/ea79c06c-0f69-4d19-8ea5-5a42886b33c3" />

MohamedBassem:
This seems to be stemming from the fact that it's using a very old chrome installation. I do agree, that it's not a priority to find a replacement. Upgrading it to pri/high and will aim to include it in the next release.

pdc1:
[...]

PR #2403, opened 2026-01-16
Update Chrome image reference in docker-compose

buildplan:
gcr.io images are getting any updates because gcr was deprecated - switched chrome image from gcr to docker hub registry.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -20,7 +20,7 @@ services:
       # If you want to mount a custom directory, change the volume mapping above instead.
       DATA_DIR: /data # DON'T CHANGE THIS
   chrome:
-    image: gcr.io/zenika-hub/alpine-chrome:124
+    image: docker.io/zenika/alpine-chrome:124
     restart: unless-stopped
     command:
       - --no-sandbox


docker/docker-compose.yml
```diff
@@ -20,7 +20,7 @@ services:
       # If you want to mount a custom directory, change the volume mapping above instead.
       DATA_DIR: /data # DON'T CHANGE THIS
   chrome:
-    image: gcr.io/zenika-hub/alpine-chrome:124
+    image: docker.io/zenika/alpine-chrome:124
```
xuatz, line 23: due to the uncertain origin of the recent images of `zenika/alphine-chrome`, we likely will not continue using it as mentioned here https://github.com/karakeep-app/karakeep/issues/2172

however we haven't really decided if we will go with `docker.io/chromedp/headless-shell` or `ghcr.io/browserless/chromium`

buildplan: Thanks I didn't know any of the history. Apologies should have checked. 

I'm using docker hub image which seems to work fine but feel free to close this I can wait for a permanent solution.

MohamedBassem:
Thanks for the PR, but as @xuatz mentioned, we're still not certain about the future of this image. I'm currently considering just pulling chrome inside the worker container or something. So for now, I'll close the PR.

PR #2418, opened 2026-01-20
suggestion: "Karakept!" as an alternative to the phrasing "Hoarded!" in the browser and phone apps.

dalinicus:
Just a suggestion a change to the browser and phone apps to replace the status result of `Hoarded!` to `Karakept!` to better reflect the new name of the app


xuatz:
nice lol

PR #2456, opened 2026-02-05
feat(search): fallback search when Meilisearch unavailable

### LINKED ISSUE #1187: Search fallback

Hi, and thanks for developing Hoarder, it's great.

Could a fallback search method be added, in case where Meilisearch is either not installed or not currently available ? This fallback would only support the most basic functionalities, such as search in bookmarks titles only.

Although the benefits of using external tools for deep, complex searches are impressive, I believe Hoarder would be much more accessible if a lighter version, such as the one presented in the minimal installation of the docs, could work on its own. 

Currently, if Meilisearch is not available, the only option that I'm aware of to browse your bookmarks is to scroll and open lists, using your mouse. If a very simple search functionality was available as a fallback, users on lower grade hardware could still browse their bookmarks. Again, I'm picturing search in bookmarks title only, not in the content.  

No (search does not work at all without Meilisearch, although it is not a hard requirement as Hoarder still loads without).

### Additional context

_No response_

LouisLau-art:
Fixes #1187.

Adds a lightweight fallback for bookmark search when Meilisearch isn’t configured or is temporarily unavailable.

Behavior:
- If a search engine client is available, keep using it.
- If it’s missing (minimal install) or errors at query time, fall back to a title-only LIKE query in the DB.

Notes:
- This is intentionally minimal: title-only search (no full-text / content search) to preserve usability without Meilisearch.
- Keeps existing qualifiers/matchers support since filtering still happens in the router.

Tests:
- `packages/trpc/routers/bookmarks.searchFallback.test.ts`

### DIFF
--- a/packages/trpc/routers/bookmarks.searchFallback.test.ts
+++ b/packages/trpc/routers/bookmarks.searchFallback.test.ts
@@ -49,14 +49,31 @@ describe("Bookmark Search Fallback", () => {
   }) => {
     const api = apiCallers[0]!;
 
-    await createBookmark(api, "First");
-    await createBookmark(api, "Second");
+    await createBookmark(api, "First", "https://first.example");
+    await createBookmark(api, "Second", "https://second.example");
 
     const res = await api.bookmarks.searchBookmarks({
       text: "",
       limit: 10,
     });
 
-    expect(res.bookmarks.length).toBeGreaterThan(0);
+    expect(res.bookmarks.length).toBe(2);
+  });
+
+  test<CustomTestContext>("searchBookmarks with matcher that yields zero IDs returns empty results", async ({
+    apiCallers,
+  }) => {
+    const api = apiCallers[0]!;
+
+    await createBookmark(api, "First", "https://first.example");
+    await createBookmark(api, "Second", "https://second.example");
+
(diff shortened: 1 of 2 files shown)


LouisLau-art:
Follow-up pushed to address review notes:

- Handle empty matcher results (`in` filter with empty `values`) by short-circuiting the fallback search to empty results (avoids invalid `inArray(..., [])`).
- Use Drizzle `count()` instead of raw `sql` for the fallback totalHits query.
- Strengthened/extended vitest coverage (empty query + matcher yielding zero IDs).

Commit: ec19cc0e

LouisLau-art:
Addressed the LIKE pattern concern:

- Escape `%`, `_`, and `\` in the fallback title-only search and use an explicit `LIKE ... ESCAPE \'\\\'` clause.
- Updated the scoring CASE (prefix/contains branches) to use the escaped patterns as well.
- Added a regression test that searches for `%` and ensures it only matches titles containing a literal `%`.

Commit: f5329312

MohamedBassem (changes requested):
Sorry for the delay in reviewing this, but if we're implementing fallback search, then we should be implementing it as a plugin similar to the meilisearch plugin.

PR #2465, opened 2026-02-08
feat: add source filter to query language

### LINKED ISSUE #2298: Filter by source in query language

Hello! A nice follow up to #2036 would be to allow filtering in the smart lists by source (`"api", "web", "cli", "mobile", "singlefile", "rss"`).
It may also make sense to add this condition to the rule engine.

I'm looking to replace my current RSS reader with Karakeep. But I also want to keep using it for bookmark keeping.
One way to keep a segregation between links I manually saved vs rss-provided ones could be through smart lists, having a `source:rss` on one list to use it as my news feed and `-source:rss` on others for bookmark keeping

It could be possible to do the distinction based on the `feed:<name>` query qualifier, but one would have to write all the existing feeds in the filter. And adding a new feed would require updating the query 

### Additional context

I saw the related issues but thought it made sense to create a new one since the suggested solution differs:
 - #1536 : we wouldn't be using tags
 - #1535 : it would offer a possible solution to this issue, by using a `feed:.*` filter. But it's a way bigger scope

MohamedBassem:
Add support for filtering bookmarks by source in the search query language. Users can now use `source:<value>` to filter bookmarks by their source (api, web, cli, mobile, extension, singlefile, rss, import).

This allows users to segregate content, for example using `source:rss` to create a news feed list and `-source:rss` to exclude RSS items from bookmark keeping lists.

- Add source matcher type to search types
- Add source qualifier to query parser lexer
- Implement source case handler in parser
- Add database query logic for source filtering
- Add tests for source filter parsing
- Update documentation with source qualifier

Fixes #2298

### DIFF
--- a/apps/web/components/dashboard/search/useSearchAutocomplete.ts
+++ b/apps/web/components/dashboard/search/useSearchAutocomplete.ts
@@ -4,6 +4,7 @@ import type { LucideIcon } from "lucide-react";
 import { useCallback, useMemo } from "react";
 import { useQuery } from "@tanstack/react-query";
 import {
+  Globe,
   History,
   ListTree,
   RssIcon,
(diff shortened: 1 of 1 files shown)


MohamedBassem:
@greptile review

Issue #2459, opened 2026-02-05
Android: Support app in landscape mode on tablets


mossholderm:
### Describe the feature you'd like

On tablets with keyboards, the app is rotated 90°, making it difficult to use. Allow the app to rotate, making the app more usable.

![Image](https://github.com/user-attachments/assets/67f765b2-d148-40f2-ac8b-1ed8e0c71355)

### Describe the benefits this would bring to existing Karakeep users

This would remove the need to hold the tablet sideways to use the app.

### Can the goal of this request already be achieved via other means?

Holding the tablet sideways.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

PR #2479, opened 2026-02-13
Fix tablet rotation to default

BasavarajBankolli:
### Fix tablet rotation when keyboard is attached

fixes: #2459

The app was locked to portrait mode via `orientation: "portrait"` in `app.config.js`.
On tablets with physical keyboards, this caused the UI to appear rotated by 90°.

This change switches orientation handling to `"default"`, allowing the OS to manage rotation.
- Phones remain unaffected
- Tablets can rotate correctly, especially in keyboard/laptop mode

### DIFF
--- a/apps/mobile/app.config.js
+++ b/apps/mobile/app.config.js
@@ -4,7 +4,7 @@ export default {
     slug: "hoarder",
     scheme: "karakeep",
     version: "1.8.5",
-    orientation: "portrait",
+    orientation: "default",
     icon: {
       light: "./assets/icon.png",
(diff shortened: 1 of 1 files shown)


BasavarajBankolli:
@MohamedBassem 
Hi 👋
I’ve opened a PR that fixes the rotation issue on tablets with physical keyboards.

The root cause was a global portrait lock in app.config.js. The PR switches orientation handling to default, allowing the OS to manage rotation properly on tablets while keeping phone behavior unchanged.

PR #2495, opened 2026-02-18
Add OneTab import with optional tag

apo-mak:
Add support for importing OneTab exports. 

UI: new OneTabImportCard in ImportExport with an optional tag input and FilePickerButton. 
Hook: useBookmarkImport now accepts extraTags and passes them through. 
Importer: importBookmarksFromFile accepts extraTags and merges them into bookmark tags (deduplicated). 
Parser: add parseOneTabFile to parse OneTab lines (handles "URL | Title" or plain URL, skips non-URLs) and register the "onetab" source. 

Also add an i18n key for the OneTab import description.

### DIFF
--- a/apps/web/components/settings/ImportExport.tsx
+++ b/apps/web/components/settings/ImportExport.tsx
@@ -330,9 +330,7 @@ export function ImportExportRow() {
         </ImportCard>
         <ImportCard
           text="OneTab"
-          description={t(
-            "settings.import.import_bookmarks_from_onetab_export",
-          )}
+          description={t("settings.import.import_bookmarks_from_onetab_export")}
         >
           <FilePickerButton
             size={"sm"}


MohamedBassem:
Hey, do you mind if we drop the optional tag to keep it consistent with the rest of the importers? Users can go back and add the tag to the imported bookmarks manually if needed.

apo-mak:
> Hey, do you mind if we drop the optional tag to keep it consistent with the rest of the importers? Users can go back and add the tag to the imported bookmarks manually if needed.

No problem, I can easily drop it.
The idea was that because OneTab is a “quick save of open tabs,” it is usually unorganized and messy. Therefore, I thought it would be better to give users the ability to add a tag in case they need to go through the links later.

MohamedBassem:
@claude fix the formatting errors

MohamedBassem:
Thank you!

Issue #2369, opened 2026-01-10
[bug] Chromium containers stuck with high idle CPU after large imports


Emstar:
First, thanks for sharing this project and especially thanks for making the setup instructions complete. Too many times have little details like the .env file been left out project documentation! I was easily able to spin up Karakeep 0.30.0 on Ubuntu and feed it my list of HTML bookmarks from Firefox. It took about 9 hours but it chewed through about 2900 links, though I didn't set up AI stuff yet. Cool. 

However, while Karakeep tells me it is done, and Background Jobs shows nothing is queued up, I have about a dozen chromium processes active, consuming a _lot_ of CPU. These processes have been running for _hours_. I assume this is not intended behavior. 

I restarted the container, and a handful of chromium processes returned, but they were all idle. That seemed pretty reasonable. 

If I get a stuck chromium process again... What should I do, besides restarting to clean it up? Is there anything I should look at to find info for a formal bug report?

MohamedBassem:
@Emstar this is a behavior I've recently started observing as well, I think it's a bug somewhere. Will make it as such so that we can try to debug it.

MohamedBassem:
I've sent https://github.com/karakeep-app/karakeep/pull/2503 as a way to mitigate this issue, and will be included in the next release happening today.

PR #2503, opened 2026-02-22
fix(workers): mitigate leaking browser contexts and setup and auto reaper

MohamedBassem:
#2369


MohamedBassem:
@greptile review

PR #2514, opened 2026-02-24
Revise Gemini API documentation for billing and models

faximan:
As I just went through this I thought this might be helpful.

* If you have no billing set up, you get 429s (https://github.com/karakeep-app/karakeep/discussions/2347).
* If you're trying to use 2.0-flash, you get 404s (The gemini api returns an error message "not available for new users" but this is not forwarded to the karakeep logs)

### DIFF
--- a/docs/docs/03-configuration/02-different-ai-providers.md
+++ b/docs/docs/03-configuration/02-different-ai-providers.md
@@ -59,7 +59,7 @@ If you experience issues with certain models (especially OpenAI's gpt-oss models
 
 ## Gemini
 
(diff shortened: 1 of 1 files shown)


benjaminfayaz:
I have just set up karakeep with `gemini-2.5-flash-lite` as I can confirm it doesn't work with `gemini-2.0-flash`. Instead of the described 404 Errors, mine were 429 with gemini 2.0.
However I didn't need to setup a billing account and it seems to work fine. I guess the 429 error only applies to `gemini-2.5-flash-lite` when doing more than the free 10 requests/minute.
image

MohamedBassem:
Thank you!

PR #2524, opened 2026-02-26
fix(mobile): disable KeyboardProvider on Android to fix header inset

esimkowitz:
## Summary

Skip `KeyboardProvider` from `react-native-keyboard-controller` on Android entirely via a `RootKeyboardProvider` wrapper that returns bare children on Android

## Problem

On Android, `KeyboardProvider`'s `EdgeToEdgeReactViewGroup` installs a window insets listener at the root view level and zeros out the status bar top inset when `isStatusBarTranslucent=true`. Since `react-native-is-edge-to-edge` (used in Expo SDK 54+) unconditionally returns `isEdgeToEdge() => true`, this flag is always forced on.

This prevents `react-native-screens`' `CustomToolbar` from receiving the correct `systemBarInsets.top` value, causing all native stack headers to appear cramped against the status bar with no padding.

## Known upstream issues

- https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1013
- https://github.com/kirillzyusko/react-native-keyboard-controller/issues/1292

## Trade-offs

Disabling `KeyboardProvider` on Android means `KeyboardAwareScrollView` and `KeyboardGestureArea` from this library won't function on Android. These are used in 3 screens (server-address, custom headers modal, bookmark info edit). React Native's built-in keyboard avoidance will need to handle those cases instead.

## Test plan

- [ ] Verify all tab screens (Home, Lists, Tags, Highlights, Settings) show headers with proper spacing below the status bar on Android
- [ ] Verify pushed screens (bookmark detail, favourites, archive, search, settings sub-pages) have correct header positioning on Android
- [ ] Verify modal screens (new bookmark, manage tags, manage lists, edit bookmark, new list, edit list) are not affected
- [ ] Verify keyboard behavior on affected screens (server-address, bookmark info edit) is still usable
- [ ] Verify iOS is not affected by these changes

### DIFF
--- a/apps/mobile/app/dashboard/(tabs)/(highlights)/_layout.tsx
+++ b/apps/mobile/app/dashboard/(tabs)/(highlights)/_layout.tsx
@@ -14,6 +14,9 @@ export default function Layout() {
             headerLargeStyle: { backgroundColor: "transparent" },
           },
           android: {
+            headerStyle: {
+              backgroundColor: "transparent",
+            },
             contentStyle: {
               // Manual padding to avoid the native tabbar until expo fixes this in sdk 55.
               paddingBottom: 100,
--- a/apps/mobile/app/dashboard/(tabs)/(home)/_layout.tsx
+++ b/apps/mobile/app/dashboard/(tabs)/(home)/_layout.tsx
@@ -14,6 +14,9 @@ export default function Layout() {
             headerLargeStyle: { backgroundColor: "transparent" },
           },
           android: {
+            headerStyle: {
+              backgroundColor: "transparent",
+            },
             contentStyle: {
               // Manual padding to avoid the native tabbar until expo fixes this in sdk 55.
               paddingBottom: 100,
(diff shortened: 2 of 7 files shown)


apps/mobile/app/_layout.tsx
```diff
@@ -1,6 +1,7 @@
 import "@/globals.css";
 import "expo-dev-client";
 
+import type React from "react";
```
xuatz, line 4:
```suggestion
import type { ReactNode } from "react";
```
nitpick ^^

xuatz:
@esimkowitz I'm sorry but do you mind sharing a before and after screenshot to show what the fix is? It might be because somehow it doesn't affect my device (samsung fold 7) or I'm looking at the wrong place, but comparing this branch's build with the latest published android version, i can't see any differences.

MohamedBassem:
Yeah, I'm also a bit confused about what this solves. And the tradeoff it mentions are also problematic no?

esimkowitz:
This was the result of a bug introduced in `react-native-screens` 4.19 (https://github.com/software-mansion/react-native-screens/issues/3499), which is the new version installed in my other PR #2523. I didn't notice this originally because I forgot to rerun `pnpm i` after switching branches.

PR #2537, opened 2026-03-01
fix(mobile): status bar text invisible in dark mode

esimkowitz:
## Summary
- Use `StatusBar style="auto"` instead of manually computing the style from nativewind's color scheme
- Fixes the status bar text being black (invisible) in dark mode due to a race condition where nativewind's color scheme observable hasn't resolved yet on initial render

## Test plan
- [x] Open the iOS Simulator in Dark appearance and verify status bar text is white
- [x] Toggle to Light appearance and verify status bar text is black
- [x] Test in-app theme setting (Light / Dark / System) still works correctly

### DIFF
--- a/apps/mobile/app/_layout.tsx
+++ b/apps/mobile/app/_layout.tsx
@@ -22,7 +22,7 @@ export default function RootLayout() {
   useInitialAndroidBarSync();
   const router = useRouter();
   const { hasShareIntent } = useShareIntent();
-  const { colorScheme, isDarkColorScheme } = useColorScheme();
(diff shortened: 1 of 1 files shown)


MohamedBassem:
I'm seeing a white text in ios light mode with and without the fix for some reason. But I'm ok with merging this.

Screenshot 2026-03-01 at 11  45 13@2x

MohamedBassem:
Closing the app seems to have fixed it 🤷‍♂️

esimkowitz:
Yeah it gets kind of messed up with hot reloading

PR #2538, opened 2026-03-01
feat(mobile): add systemMaterial blur and improve tab icon readability on iOS <26

esimkowitz:
## Summary
- Add `systemMaterial` blur effect to the tab bar on iOS < 26 for a translucent frosted-glass appearance
- Update unselected icon and text colors for the iOS < 26 tab bar for light/dark mode
- On iOS 26+, the blur effect is left to the system's fluid tab bar design

## Test plan
- [x] Build on iOS 18.x simulator — tab bar should show translucent systemMaterial blur and updated light and dark mode unselected icon colors
- [x] Build on iOS 26 simulator — tab bar should use native fluid appearance
- [x] Build on Android — tab bar should remain unchanged with solid background

### DIFF
--- a/apps/mobile/app/dashboard/(tabs)/_layout.tsx
+++ b/apps/mobile/app/dashboard/(tabs)/_layout.tsx
@@ -1,4 +1,5 @@
 import React from "react";
+import { Platform } from "react-native";
 import {
   Icon,
   Label,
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Any you intentionally applying it only to the root layout and not the different per screen layouts?

esimkowitz:
> Any you intentionally applying it only to the root layout and not the different per screen layouts?

It didn't seem like it was necessary to add them to the per-screen layouts, the tab bar seemed to take on the systemMaterial for all the tabs if I applied it at the root. Are you noticing that there's layouts where the systemMaterial isn't being applied in the tab bar?

PR #2543, opened 2026-03-01
fix: cleanup orphaned assets if crawler crashes before database transaction commits

### LINKED ISSUE #2519: If the parser OOMs, orphaned assets / invalid data can be left in the database
### Describe the Bug

I tried adding an asset through SinglePage (an article on X.com). It didn't work because the parser was OOM. I deleted the failed bookmark and increased CRAWLER_PARSER_MEM_LIMIT_MB to 768. That fixed the issue. 

However, it left an orphaned banner image for the deleted asset in the database. I've attached screenshots. I think the image was deleted from storage, but this entry was left in the database with a null bookmark ID. I was able to fix the issue by manually deleting that entry from the database.

Is there, or should there be, a process to remove invalid data from the database, or maybe there needs to be some error handling on the crawl process so it fails gracefully if OOM?

### Steps to Reproduce

Try to add an asset that causes parser to be OOM
Delete the failed asset

### Expected Behaviour

Failed crawling doesn't leave invalid data in the db

Image

Image

### Device Details

_No response_

### Exact Karakeep Version

v0.31.0

### Environment Details

_No response_

### Debug Logs

_No response_

VedantMadane:
Fixes #2519.

### Problem
When the background crawler saves assets (screenshots, PDFs, images) to storage (S3 or local disk), there is a window of time before the database transaction commits the ssetId to the ssets table. If the worker crashes or OOMs during this window, the assets remain in storage with no database record pointing to them, effectively becoming orphaned.

### Solution
- Added 	ry...catch blocks around the core asset processing logic in crawlAndParseUrl and handleAsAssetBookmark.
- Introduced a tracking array 
ewAssetIds to collect all successfully stored assets during a single crawl attempt.
- If an error occurs before the database transaction completes, the catch block iterates through 
ewAssetIds and calls silentDeleteAsset to remove them from storage.
- This ensures data integrity between the storage layer and the database even during unexpected crashes.

### DIFF
--- a/apps/workers/workers/crawlerWorker.ts
+++ b/apps/workers/workers/crawlerWorker.ts
@@ -92,6 +92,7 @@ import {
 
 const tracer = getTracer("@karakeep/workers");
 
+/** Returns a promise that rejects when the given AbortSignal fires. */
 function abortPromise(signal: AbortSignal): Promise<never> {
   if (signal.aborted) {
     const p = Promise.reject(signal.reason ?? new Error("AbortError"));
@@ -146,6 +147,7 @@ function normalizeContentType(header: string | null): string | null {
   return header.split(";", 1)[0]!.trim().toLowerCase();
 }
 
+/** Determines if a crawl should be retried based on the HTTP status code. */
 function shouldRetryCrawlStatusCode(statusCode: number | null): boolean {
   if (statusCode === null) {
     return false;
@@ -181,6 +183,7 @@ interface CrawlerRunResult {
   status: "completed";
 }
 
+/** Builds the Playwright proxy configuration from server config and environment. */
 function getPlaywrightProxyConfig(): BrowserContextOptions["proxy"] {
   const { proxy } = serverConfig;
 
@@ -281,6 +284,7 @@ function startContextReaper() {
   );
 }
 
+/** Starts a Playwright browser instance, reusing the global one if available. */
 async function startBrowserInstance() {
   if (serverConfig.crawler.browserWebSocketUrl) {
     logger.info(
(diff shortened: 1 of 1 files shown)


VedantMadane:
Addressing the CodeRabbit/Greptile review comments:

**Re: asset cleanup after DB commit (comments 1, 3, 4):**
The \ssetPersisted\ flag on line 1620 is set only after the DB transaction completes. The outer catch (line 1638) checks \!assetPersisted\ before deleting. If the enqueue fails post-commit, it's caught in the inner try/catch (lines 1621-1636) and logged without cleanup since the asset is already persisted. This is intentional -- the asset exists in the DB, so deleting the storage object would create an orphan DB record.

**Re: deletions started before transaction commit (comment 2):**
The \ssetDeletionTasks\ array stores lambda functions (\() => silentDeleteAsset(...)\), not promises. They're only invoked on line 2011 *after* the transaction commits (line 2005) and \	ransactionCommitted\ is set (line 2007). No storage deletion runs inside the transaction.

**Re: archive asset rollback (comment 7):**
The archive step (lines 2013+) runs after the main transaction commits. If archive storage or DB linkage fails, the main bookmark data is already committed and consistent. The archive asset becomes orphaned in storage, which is acceptable since it's a supplementary asset.

**Re: retrying after post-commit enqueue failure (comment 8):**
The inner try/catch around \AssetPreprocessingQueue.enqueue\ (line 1631) catches enqueue failures separately and returns gracefully without rethrowing. The crawl is not retried.

Happy to adjust any of this if a maintainer sees issues with the approach.

VedantMadane:
Closing this -- the reviewers correctly identified a fundamental race condition: the catch-block cleanup deletes assets even after the database transaction has already committed, which creates orphaned DB references. A proper fix needs a different approach (e.g., deferred cleanup that checks the assetPersisted flag before any deletion, or a two-phase approach where deletion promises are collected but only awaited after verifying the transaction result). Thanks to the reviewers for the thorough analysis.

PR #2545, opened 2026-03-01
feat(api): adds video asset handler to API

erik-nilcoast:


MohamedBassem (changes requested):
Hey, thanks for the PR! I'm ok with adding video support but we'll need to make sure that its experience is reasonable. E.g. we need to think about what image banner do we show in the cards, etc. Also, I think the AI reviews uncovered a bunch of places that needs to handled as well.

erik-nilcoast:
@MohamedBassem makes total sense, but as that's a bigger issue. I would want to plan that out first. I can jump in the discord to talk through ideas?

PR #2546, opened 2026-03-02
fix: disable password change when password auth is disabled

### LINKED ISSUE #784: Disable password change if password login is disabled

When using an OAuth provider and disabling the password based login, I would expect that the change password form in the settings page is also disabled.

Consistent behavior with oauth provider.

An alternative would be a separate environment variable for the like for signup or password-based login. Nevertheless, I would not think that it would be that great.

- [X] I have searched for existing issues and none cover my fundamental request

### Additional context

First time user and therefore there is a possibility that I did not find the correct information. In this case, excuse my foolish request.

xingzihai:
## Summary
Fixes #784 by making password-change behavior consistent with auth config when password login is disabled.

## Root Cause
`users.changePassword` could still be called even when `auth.disablePasswordAuth=true`, and the settings UI still rendered the Change Password section.

## Changes
- Web: hide `ChangePassword` settings section when `disablePasswordAuth` is enabled.
- tRPC: add guard in `users.changePassword` to reject with `FORBIDDEN` when password auth is disabled.
- Tests: add regression test ensuring `changePassword` throws when password auth is disabled.

## Verification
- Passed pre-commit preflight (`typecheck`, `lint`, `format`).
- OpenAPI check passed.
- Added/updated unit tests in `packages/trpc/routers/users.test.ts`.

## Risk
Low. Small, issue-scoped change with explicit backend guard + test coverage.

## Rollback
Revert commit `ca378d5e15e5658d449595fcef0c8db19d892c49`.

### DIFF
--- a/apps/web/components/settings/ChangePassword.tsx
+++ b/apps/web/components/settings/ChangePassword.tsx
@@ -13,6 +13,7 @@ import {
 } from "@/components/ui/form";
 import { Input } from "@/components/ui/input";
 import { toast } from "@/components/ui/sonner";
+import { useClientConfig } from "@/lib/clientConfig";
 import { useTranslation } from "@/lib/i18n/client";
 import { zodResolver } from "@hookform/resolvers/zod";
 import { useMutation } from "@tanstack/react-query";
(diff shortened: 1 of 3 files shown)


MohamedBassem:
Thanks for the PR! I think it's fine the keep the password form when password auth is disabled. Maybe instead, we can just hide it if the user is `!localUser` as it means that they just authed with oauth and no password set for their account.

PR #2554, opened 2026-03-03
fix(mobile): fix layout issues when Android bold text accessibility option is enabled

### LINKED ISSUE #1905: Android bold font issues
### Describe the Bug

On Android 16 with bold font enabled when you go to the lists the list names do not render correctly. 

### Steps to Reproduce

1. Enable bold font. 
2. Open karakeep. 
3. Go to lists tab. 

### Expected Behaviour

Text should render correctly.

Image

### Device Details

Android app

### Exact Karakeep Version

V0.26.0

Go2Engle:
Fixes #1905

## Problem

Android's "Bold text" accessibility option causes the system to render all text at a heavier weight, which makes characters approximately 20–30% wider than normal. The app's layouts did not account for this, resulting in several visual issues:

- **Lists screen**: List item names (e.g. "Favourites", "Archive") would wrap onto multiple lines or get clipped because the label `Text` used `shrink` instead of `flex-1`, giving it no defined space to occupy
- **Settings screen navigation rows**: "Theme", "Default Bookmark View", and "Reader Text Settings" row labels and their right-side value/chevron would overlap or hard-clip (e.g. "System" → "Syste", "Reader" → "Reade") because the value text was inside a nested `View` with no flex constraints, causing React Native to measure it at full content width before applying parent bounds
- **Settings About section**: The "Server", "App Version", and "Server Version" label/value pairs would collide because neither side had proper flex sizing to prevent overlap
- **Empty states** (Favourites, Highlights, Tags): The `Animated.View` container from Reanimated did not reliably receive `width: 100%` via NativeWind, so the title text ("No Bookmarks", "No Highlights") would overflow and hard-clip, displaying only "No"
- **Theme and Default Bookmark View selection screens**: Option labels ("Light Mode", "System", "Reader", etc.) had no `numberOfLines` or flex constraint, causing wrapping or clipping in bold mode

## Fix

- **Lists screen**: Changed label `Text` from `shrink` to `flex-1 mr-2` with `numberOfLines={1}` so it takes available space and truncates cleanly
- **Settings navigation rows**: Flattened the label/value/chevron into direct children of `Pressable` (removing the intermediate `View` wrapper for the right side), gave labels `flex-1 mr-2` and `numberOfLines={1}` and value texts `numberOfLines={1}` so React Native can measure all siblings together and properly constrain each
- **Settings About section**: Added `shrink-0` to labels and `flex-1 text-right numberOfLines={1}` to values so long server URLs truncate instead of colliding
- **`EmptyState` component**: Replaced NativeWind `w-full` class with an inline `style={{ width: "100%" }}` on the `Animated.View` (Reanimated does not reliably process NativeWind classes), and added explicit `width: "100%"` and `textAlign: "center"` inline styles to the title and subtitle `Text` elements
- **Theme / Default Bookmark View selection screens**: Added `flex-1 mr-2 numberOfLines={1}` and `items-center` to option rows

### DIFF
--- a/apps/mobile/components/ui/GroupedList.tsx
+++ b/apps/mobile/components/ui/GroupedList.tsx
@@ -59,7 +59,9 @@ function NavigationRow({
       onPress={onPress}
       className="flex-row items-center justify-between px-4 py-3 active:opacity-70"
     >
-      <Text className="flex-1">{label}</Text>
+      <Text className="flex-1" numberOfLines={1}>
+        {label}
+      </Text>
       <ChevronRight size={16} />
     </Pressable>
   );


MohamedBassem:
Thank you! I've seen people report this problem multiple time and never knew it was caused by the bold text. Thanks a lot, this is amazing!

MohamedBassem:
@greptile review

PR #2556, opened 2026-03-05
fix(crawler): normalize lazy-load img attributes to src before Readability

### LINKED ISSUE #2493: Images fail to load in articles using non-standard lazy-loading tags (e.g., WeChat data-src)
### Describe the Bug

When capturing articles from certain platforms (specifically WeChat Official Accounts), images fail to display in the KaraKeep reader/preview view.
Upon inspecting the DOM, I found that these images use non-standard attributes like data-src or data-srv for lazy loading, instead of the standard src attribute. KaraKeep's parser saves the original tag but doesn't seem to map these attributes to src, causing the browser to ignore the image source.

### Steps to Reproduce

1. Capture a WeChat article (e.g., https://mp.weixin.qq.com/s/kvhc3xjpBl8Uk4-Yv0-CAw).
2. Open the bookmark in KaraKeep.
3. Observe that images are broken or replaced by placeholders.
4. Inspect the image element (which can't be seen without inspecting) to see  without a valid src attribute.

### Expected Behaviour

The parser should identify common lazy-loading attributes (like data-src, data-srv, etc.) and rewrite them to the standard src attribute during the article cleaning process.

or make it able to be seen in KaraKeep.

Image

Image

### Device Details

_No response_

### Exact Karakeep Version

v0.30.0

### Environment Details

_No response_

### Debug Logs

_No response_

zerone0x:
## Problem

Many sites — notably WeChat official accounts — defer image loading by placing the real URL in a custom data-* attribute instead of the standard `src` attribute. Common examples:

| Attribute | Used by |
|-----------|---------|
| `data-src` | Most lazy-load libs |
| `data-actualsrc` | WeChat |
| `data-srv` | WeChat |
| `data-original` | jQuery lazy plugins |
| `data-lazy` / `data-lazyload` | Various plugins |

When `` has no `src`, Mozilla Readability ignores the element entirely; DOMPurify then strips the data-* attribute. The resulting article has broken image placeholders even though the URL was available in the original HTML.

Reported in #2493.

## Fix

Add `normalizeLazyLoadImages()`, called once on the JSDOM document **before** it is passed to Readability. The function iterates over every `` element and promotes the first recognised lazy-loading attribute to `src` when `src` is absent or contains a known low-resolution placeholder (1 × 1 GIF/PNG data URI or a bare `#`).

The attribute priority list (in order):
```
data-src, data-actualsrc, data-srv, data-original,
data-lazy, data-lazyload, data-img-src, data-url
```

Images that already have a valid `src` are left untouched.

## Testing

1. Save a WeChat article (e.g. `https://mp.weixin.qq.com/s/kvhc3xjpBl8Uk4-Yv0-CAw`).
2. Open the bookmark in the reader view.
3. Images that previously showed as broken placeholders should now render correctly.

Fixes #2493

### DIFF
--- a/apps/workers/scripts/parseHtmlSubprocess.ts
+++ b/apps/workers/scripts/parseHtmlSubprocess.ts
@@ -74,6 +74,52 @@ const metascraperParser = metascraper([
   metascraperUrl(),
 ]);
 
+/**
+ * Many sites use custom data-* attributes for lazy loading images instead of the
+ * standard `src` attribute (e.g. WeChat's data-src, data-actualsrc, data-srv, or
+ * the common data-original / data-lazy patterns used by jQuery plugins).
+ *
+ * Readability and DOMPurify both operate on the DOM, so images with no `src` are
+ * either ignored or their placeholders are stripped.  We normalise these attributes
+ * to `src` *before* passing the document to Readability so that the extracted
+ * article retains the actual image URLs.
+ */
+const LAZY_SRC_ATTRS = [
+  "data-src",
+  "data-actualsrc",
+  "data-srv",
+  "data-original",
+  "data-lazy",
+  "data-lazyload",
+  "data-img-src",
+  "data-url",
+];
+
+function normalizeLazyLoadImages(document: Document): void {
+  const images = document.querySelectorAll("img");
+  for (const img of images) {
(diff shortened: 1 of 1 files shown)


MohamedBassem:
thank you!

PR #2557, opened 2026-03-05
fix(web): expand $tags/$aiTags/$userTags in AI prompt preview

### LINKED ISSUE #2511: Tag variable mentioned in the docs "$tags" does not work. "$Tags" does work (maybe)
### Describe the Bug

The [documentation](https://docs.karakeep.app/configuration/environment-variables/#inference-configs-for-automatic-tagging) says:
> You can use the placeholders $tags, $aiTags, $userTags in the prompt.

When I use "$tags" in a custom tagging rule, AI tagging does not work. I tried "$Tags" and tagging does work. This behavior changed in some way in 0.31. 

While troubleshooting this issue I noticed that in the Prompt Preview section of AI settings it says:
> Preview the actual prompts sent to AI based on your settings

Is what is shown in the prompt preview _actually_ what is sent at the prompt? The tag variables are not expanded in the preview, so either the tag variables aren't really working, or what is shown in the preview isn't what is actually sent. 

Are the API requests logged anywhere for debugging? I was wondering if the "$tags" variable possibly wasn't being expanded properly in the API request, but I don't see any way to view the actual prompt being sent.  

### Steps to Reproduce

Create custom tagging rules using "$tags". Example:

1. Check if any of the previously used tags are relevant for this content. Reuse them when appropriate to maintain consistency.
2. Here are the previously used tags that you should prioritize when appropriate: $tags

### Expected Behaviour

Tags should be created

_No response_

### Device Details

_No response_

### Exact Karakeep Version

v0.31.0

### Environment Details

Docker on Unraid

### Debug Logs

_No response_

zerone0x:
## Problem

The **Prompt Preview** section in AI Settings renders custom prompt texts verbatim, so tag-list placeholders (`$tags`, `$aiTags`, `$userTags`) appear unexpanded even though the inference worker substitutes them at runtime.

This mismatch misleads users into thinking the placeholders are not working at all — the preview shows the literal string `$tags` rather than the resolved tag list.

Reported in #2511.

## Fix

1. **`hasTagPlaceholder(texts)`** – detects whether any custom prompt uses a tag placeholder (same check as `containsTagsPlaceholder` in the worker).
2. **Conditional tag fetch** – when a placeholder is detected, all user tags are fetched via `api.tags.list`.  The query is disabled when no placeholder is present, so users who don't use this feature pay no extra network cost.
3. **`expandTagPlaceholders(texts, tags)`** – mirrors the `replaceTagsPlaceholders` logic from the inference worker so the preview stays in sync with runtime behaviour.
4. The text, image, and summarization prompt previews all run through `withTagsExpanded()` before being passed to the builder functions.

## Before / After

**Before:** prompt preview shows
```
- 
```

**After:** prompt preview shows
```
- [tag-a, tag-b, tag-c, ...]
```

Fixes #2511

### DIFF
--- a/apps/web/components/settings/AISettings.tsx
+++ b/apps/web/components/settings/AISettings.tsx
@@ -723,6 +723,42 @@ export function TaggingRules() {
   );
 }
 
+/**
+ * Expand $tags / $aiTags / $userTags placeholders in custom prompt texts so
+ * that the preview panel shows exactly what will be sent to the AI.
+ *
+ * The logic mirrors apps/workers/workers/inference/tagging.ts
+ * `replaceTagsPlaceholders` to keep the preview accurate.
+ */
+function expandTagPlaceholders(
+  texts: string[],
+  tags: { name: string; numBookmarksByAttachedType: Record<string, number> }[],
+): string[] {
+  const tagsStr = `[${tags.map((t) => t.name).join(", ")}]`;
+  const aiTagsStr = `[${tags
+    .filter((t) => t.numBookmarksByAttachedType["human"] ?? true)
+    .map((t) => t.name)
+    .join(", ")}]`;
+  const userTagsStr = `[${tags
+    .filter((t) => t.numBookmarksByAttachedType["human"] ?? false)
+    .map((t) => t.name)
+    .join(", ")}]`;
+
+  return texts.map((text) =>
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thank you!

PR #2558, opened 2026-03-05
fix(web): show "+" in tag count badges when more pages exist

### LINKED ISSUE #2395: Tag count is incorrect while "Load More" button exists
### Describe the Bug

When you have so many tags that the tags list (e.g., "Your Tags") gets paginated, the count in this list doesn't show the actual correct count of tags available. 

### Steps to Reproduce

1. Have more than 50 tags (self-written or made by AI).
2. Open the Tags-page.
3. See the count of the tags be 50.
4. Click "Load More".
5. See the count of tags increase.

### Expected Behaviour

I'd expect to see the full amount/count of tags regardless if I see them all or not.

I'm talking about this count:
Image

### Device Details

Firefox 147.0 (64-Bit) on Windows 11

### Exact Karakeep Version

v0.30.0

zerone0x:
## Problem

The Tags page shows a count badge next to each category header (*Your Tags*, *AI Tags*, *Unused Tags*).  The badge only reflects the number of tags loaded into the current page, so with more than 50 tags the badge shows a stale partial count (e.g. **50**) and jumps unexpectedly each time "Load More" is clicked.

Reported in #2395 (`pri/low`, `status/approved`).

## Root Cause

`AllTagsView` derives the count from `allHumanTags.length` / `allAiTags.length` / `allEmptyTags.length`, which are the accumulated in-memory arrays from `usePaginatedSearchTags`.  The paginated API response does not include a grand total, so computing the true count without fetching all pages would require an extra network round-trip.

## Fix

Append **"+"** to each count badge when `useInfiniteQuery` reports `hasNextPage === true`.

* `50+` → there are more tags not yet loaded
* `50` → all pages are loaded; the badge shows the exact count as before

This is already a well-established UX convention (GitHub issue counts, package registries, etc.) and avoids any extra API calls.

Once all pages are loaded the "+" disappears automatically and the badge shows the precise total, keeping the existing behaviour unchanged for users who load everything.

Fixes #2395

### DIFF
--- a/apps/web/components/dashboard/tags/AllTagsView.tsx
+++ b/apps/web/components/dashboard/tags/AllTagsView.tsx
@@ -332,7 +332,10 @@ export default function AllTagsView() {
         <CardHeader>
           <CardTitle className="flex items-center gap-2">
             <span>{t("tags.your_tags")}</span>
-            <Badge variant="secondary">{allHumanTags.length}</Badge>
+            <Badge variant="secondary">
+              {allHumanTags.length}
+              {hasNextPageHumanTags ? "+" : ""}
+            </Badge>
           </CardTitle>
           <CardDescription>{t("tags.your_tags_info")}</CardDescription>
         </CardHeader>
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thank you!

PR #2574, opened 2026-03-08
fix(api): ensure sourceUrl is saved when creating asset bookmarks

hoilc:
Previously, when creating a new bookmark of type `asset` from API `/v1/bookmarks`, the `sourceUrl` was hardcoded to `null` during database insertion, ignoring the `sourceUrl` passed in the request body. This fix ensures that the provided `sourceUrl` is properly saved to the `bookmarkAssets` table and correctly returned in the API response.


MohamedBassem:
thank you!

PR #2587, opened 2026-03-15
fix(db): use millisecond precision for bookmark timestamps

### LINKED ISSUE #2583: New bookmarks not appearing first when sorted by newest
### Describe the Bug

When I save a new bookmark (from the browser extension or the web UI) and sort by "Newest first", the new bookmark does not appear at the top.

Refreshing the page does not fix it.

After debugging my self-hosted instance I found that some bookmarks are stored with `createdAt` in Unix seconds instead of milliseconds.

Example value stored:
1773238339

Correct millisecond timestamp example:
1773238379976

Because of this, bookmarks created with seconds timestamps sort far below the normal millisecond timestamps.

Database query used to confirm:

select id,title,createdAt
from bookmarks
order by createdAt desc;

I suspect the issue may be related to timestamp handling in:

packages/db/schema.ts  
packages/shared/types/bookmarks.ts

Changing timestamp mode from `timestamp` to `timestamp_ms` appears to fix it.

Environment:
Self-hosted Karakeep (Docker)

### Steps to Reproduce

1. Save a new bookmark from the browser extension or web UI.
2. Open the bookmarks page.
3. Sort bookmarks by **Newest first**.
4. The newly created bookmark does not appear at the top.
5. Refreshing the page does not change the order. 

### Expected Behaviour

When a new bookmark is saved, it should appear at the top when sorting by **Newest first**.

_No response_

### Device Details

_No response_

### Exact Karakeep Version

Latest Docker image (ghcr.io/karakeep-app/karakeep:latest)

### Environment Details

Docker

### Debug Logs

_No response_

zerone0x:
## Summary
Fixes bookmark `createdAt` being stored in Unix seconds instead of milliseconds, which causes incorrect sort order when sorting by "Newest first".

The `createdAtField()` and `modifiedAtField()` helper functions in the Drizzle schema used `mode: "timestamp"` (seconds precision). This meant new bookmarks got timestamps like `1773238339` instead of `1773238379976`, causing them to sort far below bookmarks with millisecond timestamps.

Changed both helpers to use `mode: "timestamp_ms"` for consistent millisecond-precision storage, matching other timestamp fields in the schema (e.g. `emailVerified`, `expires`).

Fixes #2583

### DIFF
--- a/packages/db/schema.ts
+++ b/packages/db/schema.ts
@@ -16,13 +16,13 @@ import {
 import { BookmarkTypes } from "@karakeep/shared/types/bookmarks";
 
 function createdAtField() {
-  return integer("createdAt", { mode: "timestamp" })
+  return integer("createdAt", { mode: "timestamp_ms" })
     .notNull()
     .$defaultFn(() => new Date());
 }
 
 function modifiedAtField() {
-  return integer("modifiedAt", { mode: "timestamp" })
+  return integer("modifiedAt", { mode: "timestamp_ms" })
     .$defaultFn(() => new Date())
     .$onUpdate(() => new Date());
 }


MohamedBassem:
This is backward incompatible and I don't see how it can fix the issue mentioned in the original issue.

PR #2589, opened 2026-03-16
fix(workers): handle TWZ RSS category objects

### LINKED ISSUE #2569: TWZ RSS feed fetch succeeds but imports 0 entries in v0.31.0
### Describe the Bug

  Karakeep fetches `https://www.twz.com/feed` successfully, but processes it as `0 entries`, so nothing imports.

  This is not a network failure:
  - the feed returns `200 OK`
  - KaraKeep marks the fetch as `success`
  - the worker log shows the feed was fetched successfully

  But after fetching, KaraKeep logs:
  - `Found 0 entries in feed "The War Zone"`
  - `No entries found.`

  As a result:
  - no new bookmarks are created
  - no new `rssFeedImports` rows are added
  - the feed appears healthy in the UI even though nothing is imported

### Steps to Reproduce

  1. Add the RSS feed `https://www.twz.com/feed`
  2. Open `Settings > Feeds`
  3. Trigger `Fetch now` for the feed
  4. Check the feed status and logs
  5. Observe that the fetch succeeds, but KaraKeep reports `0 entries` and imports nothing

### Expected Behaviour

  Karakeep should detect the RSS `<item>` entries from `https://www.twz.com/feed` and import new bookmarks normally.

  Additional context:

  - Exact feed URL: `https://www.twz.com/feed`
  - The feed XML is valid and contains normal RSS `<item>` elements with fields like:
    - `title`
    - `link`
    - `guid`
    - `pubDate`
    - `category`

  I also tested:
  - disabling category/tag import for the feed
  - re-running `Fetch now`

  That did not resolve the issue. The worker still reported `Found 0 entries`.

  Possible parser clue:

  While debugging, I noticed that TWZ feed items include category data that may not match a simple `string[]` assumption. If KaraKeep is validating parsed items too strictly, that may explain why the feed is fetched
  successfully but all items are discarded and the worker reports `Found 0 entries`.

  I do not know if this is the exact root cause in upstream code, but it seems like the failure is happening after fetch and before import, during RSS item parsing/normalization.

### Device Details

Truenas 25.10

### Exact Karakeep Version

 v0.31.0

### Environment Details

  Docker Compose stack managed through Portainer on TrueNAS SCALE, behind reverse proxy/public hostname. Also tested directly via LAN URL.

### Debug Logs

  Relevant worker log output:

  ```text
  [feed][56372] Starting fetching feed "The War Zone" (y1cwipc28507bzkvn6myrlmk) ...
  [feed][56372] Successfully fetched feed "The War Zone" (y1cwipc28507bzkvn6myrlmk) ...
  [feed][56372] Found 0 entries in feed "The War Zone" (y1cwipc28507bzkvn6myrlmk) ...
  [feed][56372] No entries found.
  [feed][56372] Completed successfully
```

  Database state after rerun:

  Feed: The War Zone
  Enabled: true
  Import tags: false
  Last fetched status: success
  Last fetched at: 2026-03-06 18:01:44 local

  rssFeedImports count for active TWZ feed: 361
  Last import time: 2026-01-14 19:00:05 local

  This indicates the feed fetch is succeeding, but KaraKeep is not recognizing any entries as importable.

venku122:
## Summary
- add a fixture-backed regression for the TWZ feed shape that started returning zero imported entries
- normalize `rss-parser` category objects into plain tag strings and fall back to `guid` when feeds do not provide an item `id`
- route worker feed parsing through the shared helper and keep the verification deterministic and local

## Verification
- `pnpm --filter @karakeep/workers test -- workers/feedParser.test.ts`
- `pnpm --filter @karakeep/workers typecheck`
- `pnpm --filter @karakeep/workers lint`
- manual live parse of `https://www.twz.com/feed` now returns 30 items with string category names

Closes #2569.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,4 @@ auth_failures.log
 .aicontext/
 .codex
 .claude/worktrees
+jean.json
--- a/apps/landing/index.html
+++ b/apps/landing/index.html
@@ -1,39 +1,110 @@
 <!doctype html>
 <html lang="en">
   <head>
-    <title>Karakeep</title>
+    <meta charset="UTF-8" />
+    <title>
+      Karakeep - The Bookmark Everything App | Save, Organize & Tag with AI
+    </title>
     <meta
       name="description"
-      content="The Bookmark Everything app. Hoard links, notes, and images and they will get automatically tagged AI."
+      content="Karakeep is the open-source bookmark manager for links, notes, and images. Automatically organize and tag your bookmarks with AI. Self-hostable, with apps for iOS, Android, Chrome, and Firefox."
     />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <meta property="og:title" content="Karakeep" />
+    <link rel="canonical" href="https://karakeep.app/" />
+    <meta name="robots" content="index, follow" />
+    <meta name="theme-color" content="#7c3aed" />
+
+    <!-- Open Graph -->
+    <meta property="og:type" content="website" />
+    <meta property="og:site_name" content="Karakeep" />
+    <meta property="og:url" content="https://karakeep.app/" />
+    <meta
+      property="og:title"
+      content="Karakeep - The Bookmark Everything App"
+    />
     <meta
       property="og:description"
-      content="The Bookmark Everything app. Hoard links, notes, and images and they will get automatically tagged AI."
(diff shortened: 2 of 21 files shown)


venku122:
This PR was made by Codex to fix the issue in https://github.com/karakeep-app/karakeep/issues/2569 

I had it make a red-green regression test to figure out the missing guid issue. Happy to remove the testing dependencies if it add overhead to the project.

MohamedBassem:
Thanks for the fix!

PR #2609, opened 2026-03-20
fix(import-export): preserve list hierarchy on karakeep round-trip

mgrusso:
When importing a karakeep JSON export, top-level lists (those with no parent in the source file) were placed under the import-session root list instead of being created at the top level of the user's list space. This meant that a round-trip export → import produced a different hierarchy than the original, making it impossible to restore state on another instance.

Add a `restoreTopLevelLists` option to `importBookmarksFromFile` that, when enabled, creates exported top-level lists without a parent so the full original hierarchy is faithfully reproduced.  The import-session root list is still created and serves as the fallback container for any bookmarks that carry no list assignment.

Enable the option automatically for the "karakeep" source in the web hook so that every karakeep-format import is a true state restoration.

### DIFF
--- a/apps/web/lib/hooks/useBookmarkImport.ts
+++ b/apps/web/lib/hooks/useBookmarkImport.ts
@@ -93,6 +93,9 @@ export function useBookmarkImport() {
           onProgress: (done, total) => setImportProgress({ done, total }),
         },
         {
+          // Karakeep imports restore the original list hierarchy at the top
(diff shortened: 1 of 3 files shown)


MohamedBassem:
Thanks for the PR. And while I get what you're suggesting, I think for consistency, we should keep the root "imported bookmarks" list. Users can easily remove that root list and all of its children will get promoted to root lists automatically.

PR #2615, opened 2026-03-22
docs: add karaclean to community projects

lmgarret:


MohamedBassem:
Thanks for the PR @lmgarret. Can you move the docs to `docs/docs` instead of `versioned_docs`? Otherwise it'll get removed in the next versioned docs release.

lmgarret:
> Thanks for the PR @lmgarret. Can you move the docs to `docs/docs` instead of `versioned_docs`? Otherwise it'll get removed in the next versioned docs release.

🤦will do, thanks!

EDIT: updated the right file

PR #2620, opened 2026-03-24
feat: Allow multiple lists for Rule Engine #2490

### LINKED ISSUE #2490: Allow multiple lists for Rule Engine

Select multiple lists in rule engine

I like to create rules that automatically archive saves when added to a list.  I have to do this for every list individually.  It would be nice to have one archive when added to list rule that I can just select the lists I want it to apply to.

Yes, but requires many rules for the same task.

### Additional context

_No response_

Ahmed-Abdel-karim:
## Summary

Previously, rule engine events (`addedToList` / `removedFromList`) only 
supported a single list per rule, requiring users to create duplicate 
rules for each list. This PR allows multiple lists to be selected for 
a single rule. Closes #2490

### Schema
- Remove `listId` from `ruleEngineRulesTable` and rely solely on `listIds` stored within the `event` payload
- Remove the composite foreign key `(userId, listId) → bookmarkLists` that was enforcing single list ownership
- Introduce `cleanupRulesAfterListDeletion` to Remove deleted list IDs from `event.listIds` and Delete rules when no list IDs  remain (i.e. the rule becomes invalid)
- Execute `cleanupRulesAfterListDeletion` as part of the list deletion flow

### Migration
- Data migration that preserves existing rules by updates the embedded `event` JSON to replace `listId` with `listIds` array

### Types
- Split `zRuleEngineEventSchema` into two separate schemas:
  - `zRuleEngineEventSchema` — dispatched events, retains single `listId` (what actually happened)
  - `zRuleEngineRuleEventSchema` — rule filter definition, uses `listIds` array (what the rule matches against)

### Rule Engine
- Replace `deepEqual` comparison for list events with a dedicated `doesEventMatchRule` function
- The new function checks if the dispatched `listId` is included in the rule's `listIds` array

### UI
- Update `BookmarkListSelector` to support both single and multi-select modes
- Extract `ListNameFromPath` as a shared utility replacing inline path formatting that was duplicated across multiple components

## Screenshots
image
multi-select

## Testing
- Existing single-list rules are automatically migrated and continue working
- New rules can target multiple lists in a single rule
- Rule engine correctly triggers when a bookmark is added to any of the selected lists


packages/trpc/models/lists.ts
```diff
@@ -462,19 +475,111 @@ export abstract class List {
...
+        // Log and skip corrupted rule, continue with others
+        console.error(`Failed to parse event JSON for rule ${rule.id}`);
+        continue;
+      }
+
+      const ruleEvent = zRuleEngineRuleEventSchema.safeParse(parsedEvent);
+      if (!ruleEvent.success) {
+        // Log and skip invalid rule, continue with others
+        console.error(`Failed to validate event schema for rule ${rule.id}`);
+        continue;
```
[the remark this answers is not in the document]
Ahmed-Abdel-karim: @coderabbitai
Does this fix the issue (adding  sql`json_valid(${ruleEngineRulesTable.event})`), if not suggest a fix for this issue
.where(
        and(
          eq(ruleEngineRulesTable.userId, this.ctx.user.id),
          sql`json_valid(${ruleEngineRulesTable.event})`,
          sql`json_extract(${ruleEngineRulesTable.event}, '$.type') IN ('addedToList', 'removedFromList')`,
          sql`EXISTS (
            SELECT 1
            FROM json_each(json_extract(${ruleEngineRulesTable.event}, '$.listIds'))
            WHERE value = ${this.list.id}
          )`,
        ),
      );

packages/db/drizzle/0083_rule_engine_multi_list_support.sql
```diff
@@ -0,0 +1,45 @@
...
+	FOREIGN KEY (`userId`,`tagId`) REFERENCES `bookmarkTags`(`userId`,`id`) ON UPDATE no action ON DELETE cascade
+);
+--> statement-breakpoint
+DELETE FROM ruleEngineRules 
+WHERE json_extract("event", '$.type') 
+IN ('addedToList', 'removedFromList') 
+AND (
+    json_extract("event", '$.listId') IS NULL
+    OR json_extract("event", '$.listId') = ''
+);
```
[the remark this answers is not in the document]
Ahmed-Abdel-karim: @coderabbitai Does moving this block of code for deleting the malformed data at the top of the migration file before the 'PRAGMA foreign_keys=OFF;' line fix the issue like this fix the issue, if not provide the recommended solution

`DELETE FROM ruleEngineRules 
WHERE json_extract("event", '$.type') 
IN ('addedToList', 'removedFromList') 
AND (
    json_extract("event", '$.listId') IS NULL
    OR json_extract("event", '$.listId') = ''
);
--> statement-breakpoint
PRAGMA foreign_keys=OFF;
--> statement-breakpoint
CREATE TEMPORARY TABLE `_backup_ruleEngineActions` AS 
SELECT * FROM `ruleEngineActions`;
--> statement-breakpoint
CREATE TABLE `__new_ruleEngineRules` (
	`id` text PRIMARY KEY NOT NULL,
	`enabled` integer DEFAULT true NOT NULL,
	`name` text NOT NULL,
	`description` text,
	`event` text NOT NULL,
	`condition` text NOT NULL,
	`userId` text NOT NULL,
	`tagId` text,
	FOREIGN KEY (`userId`) REFERENCES `user`(`id`) ON UPDATE no action ON DELETE cascade,
	FOREIGN KEY (`userId`,`tagId`) REFERENCES `bookmarkTags`(`userId`,`id`) ON UPDATE no action ON DELETE cascade
);
--> statement-breakpoint
.....`

packages/shared/utils/listUtils.ts
```diff
@@ -58,3 +58,6 @@ export function listsToTree(lists: ZBookmarkList[]) {
       allPaths.find((path) => path[path.length - 1].id === id),
   };
 }
+
+export const ListNameFromPath = (path: ZBookmarkList[]) =>
```
MohamedBassem, line 62: nit: should this be camel case similar to other utils in this file?

packages/db/drizzle/0083_rule_engine_multi_list_support.sql
```diff
@@ -0,0 +1,53 @@
...
```
MohamedBassem, line 33: is this still needed now that you've done the deletion in the beginning?

packages/db/drizzle/0083_rule_engine_multi_list_support.sql
```diff
@@ -0,0 +1,53 @@
...
+--> statement-breakpoint
+PRAGMA foreign_keys=OFF;
+--> statement-breakpoint
+CREATE TEMPORARY TABLE `_backup_ruleEngineActions` AS 
+SELECT * FROM `ruleEngineActions`;
```
MohamedBassem, line 12: Today I learned about `CREATE TEMPORARY TABLE` :)

packages/shared/types/rules.ts
```diff
@@ -35,6 +35,16 @@ const zArchivedEvent = z.object({
...
 
+const zAddedToListRuleEvent = z.object({
+  type: z.literal("addedToList"),
+  listIds: z.array(z.string()),
+});
+
+const zRemovedFromListRuleEvent = z.object({
+  type: z.literal("removedFromList"),
+  listIds: z.array(z.string()),
+});
```
MohamedBassem, line 46: mind expanding on why we need to create new rule event instead of modifying the existing one? With the database migration it should be safe to just modify the old event? I feel like copying `zRuleEngineEventSchema` to only include the new events sounds a bit error prone.

packages/trpc/routers/rules.test.ts
```diff
@@ -347,14 +347,21 @@ describe("Rules Routes", () => {
...
       };
 
       await expect(() => api.create(invalidRuleInput)).rejects.toThrow(
         /List not found/,
       );
+
+      await expect(() =>
+        api.create({
+          ...invalidRuleInput,
+          event: { type: "addedToList", listIds: [listId, otherUserListId] },
```
MohamedBassem, line 362: Ah, that's a good test. Thanks for adding this!

packages/trpc/models/lists.ts
```diff
@@ -462,19 +464,105 @@ export abstract class List {
     }
   }
 
-  async delete() {
-    this.ensureCanManage();
-    const res = await this.ctx.db
-      .delete(bookmarkLists)
+  protected async cleanupRulesAfterListDeletion(tx: KarakeepDBTransaction) {
```
MohamedBassem, line 467: This seems like it needs to be covered with some exhaustive tests for all the different cases here. Rule deletions, rule updates, other rules not impacted, etc.

packages/trpc/models/lists.ts
```diff
@@ -462,19 +464,105 @@ export abstract class List {
...
+        .join(" ");
+      await tx.run(sql`
+          UPDATE ${ruleEngineRulesTable}
+          SET event = CASE ${sql.raw(caseStatements)} END
+          WHERE id IN (${sql.join(
+            rulesToUpdate.map((r) => r.id),
+            sql`, `,
+          )})
+      `);
+    }
```
MohamedBassem, line 547: Not a big fan of the raw SQL here. Is it possible to construct single rule engine updates and await them with `Promise.all`?

MohamedBassem (changes requested):
This was a pretty big PR and you seemed to covered all the angles for it perfectly. Thanks a lot. Overall, looks good to me, but left some minor comments here and there before we can merge it.

packages/db/drizzle/0083_rule_engine_multi_list_support.sql
Ahmed-Abdel-karim: Good catch — the deletion at the beginning of the script is actually replacing this one.
Looks like this part stayed here due to a merge issue 😅 I’ll remove it.

packages/shared/types/rules.ts
Ahmed-Abdel-karim: Yeah, that makes sense. I think they are two separate concepts that both have the same name (event):

`RuleEngineEvent` is the concrete event produced by the system, e.g. one bookmark was added to one list, so it carries listId.
`RuleEngineRuleEvent` is the rule `matcher/filter`, so e.g. match if the event list is any of these configured lists, so it carries listIds.

So I didn’t want to change all places that produce this event to listIds (expecting listIds and producing listIds) just because there is a change in the rule `matcher/filter`.

But of course, I understand that this can be error-prone, and I can change/refactor it if you think that’s better.

MohamedBassem: Oh, I see what you mean now. This makes a lot of sense actually!

MohamedBassem:
This looks good to me! Thanks a lot @Ahmed-Abdel-karim.

@greptile review
@codex review

PR #2629, opened 2026-03-27
feat(web): auto-add bookmark to list on selection

### LINKED ISSUE #2604: Delete the 'Add' Button from Manage Lists

After manually going through a whole bunch of bookmarks that I wanted to delete an existing list and add a new one, I found using the Add button is almost like forcing double entry. In my opinion, if I select a list from the dropdown box, just add it automatically. I don't need to double verify by pressing the add button. If the list is added in error, then you can just hit the 'x' to delete it. Subtle feature request but I think it will make a big difference!

Streamlines list editing.

no

### Additional context

_No response_

xingzihai:
Fixes #2604

## Summary
Automatically add bookmark to the selected list when user clicks on a list item in the ManageListsModal.

## Changes
- BookmarkListSelector now triggers addToList when a list is selected
- Improved UX: users don't need to click "Add" button separately
- Added loading state and prevented duplicate submission

## Testing
- Tested locally with list selection
- Verified bookmark is added correctly


MohamedBassem:
@greptile review

MohamedBassem:
Thank you!

PR #2635, opened 2026-03-29
feat(api): expose feed subscriptions via REST API

larsborn:
## Summary

Adds REST endpoints for managing RSS feed subscriptions under `/api/v1/feeds`, matching the existing tRPC functionality.

| Method | Endpoint | Description |
|--------|----------|-------------|
| `GET` | `/feeds` | List all feeds |
| `POST` | `/feeds` | Create a feed |
| `GET` | `/feeds/{feedId}` | Get a single feed |
| `PATCH` | `/feeds/{feedId}` | Update a feed |
| `DELETE` | `/feeds/{feedId}` | Delete a feed |
| `PUT` | `/feeds/{feedId}/fetch` | Trigger an immediate fetch |

## Motivation

I'm building an automation that scrapes various sources on the web and cross-checks whether I already have a matching feed subscription in Karakeep. The REST API is the natural integration point for this, but feeds were the one resource not yet exposed through it.

## Changes

- **`packages/api/routes/feeds.ts`** — new Hono route handlers (follows the tags/lists pattern exactly)
- **`packages/open-api/lib/feeds.ts`** — OpenAPI registry for the 6 endpoints
- **`packages/api/index.ts`** — register `/feeds` route on the v1 router
- **`packages/open-api/index.ts`** — register feeds registry + tag description
- **`packages/open-api/karakeep-openapi-spec.json`** — regenerated

No new dependencies. No changes to existing behavior.

## Test plan

- [x] `pnpm run typecheck` — all 22 packages pass
- [x] `pnpm run check` in `packages/open-api` — spec matches source

### DIFF
--- a/packages/api/index.ts
+++ b/packages/api/index.ts
@@ -14,6 +14,7 @@ import admin from "./routes/admin";
 import assets from "./routes/assets";
 import backups from "./routes/backups";
 import bookmarks from "./routes/bookmarks";
+import feeds from "./routes/feeds";
 import health from "./routes/health";
 import highlights from "./routes/highlights";
 import lists from "./routes/lists";
@@ -41,7 +42,8 @@ const v1 = new Hono<{
   .route("/assets", assets)
   .route("/admin", admin)
   .route("/rss", rss)
-  .route("/backups", backups);
+  .route("/backups", backups)
+  .route("/feeds", feeds);
 
 const app = new Hono<{
   Variables: {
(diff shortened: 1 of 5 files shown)


larsborn:
`updateTag` (the analogous `PATCH` endpoint in `tags.ts`) does not document a 400. The `zValidator` 400 behavior is implicit and undocumented across all similar `PATCH` endpoints. Adding it only to `updateFeed` would be inconsistent with the rest of the spec but happy to do so if you want me to!

MohamedBassem:
Thank you! I think I'd prefer the `/fetch` endpoint to be a `POST` instead of `PUT` to be consistent with some of the other actions that we have (e.g. `/bookmarks/:id/summarize`). I'll do that in a followup commit

PR #2657, opened 2026-04-07
fix: add data: to style-src and font-src CSP directives in serveAsset

howwohmm:
**Problem**
The `serveAsset` handler in `packages/api/utils/assets.ts` was missing `data:` in the `style-src` CSP directive and had no `font-src` directive at all. This caused browsers to block inline data URIs used by fonts and styles when assets are served through the API.

**Fix**
Added `data:` to `style-src` and inserted a new `font-src https: data:` directive between `style-src` and `connect-src` in the Content-Security-Policy header built inside `serveAsset`.

**What this doesn't change**
CSP directives for other handlers (e.g. the main app routes) are unaffected.


MohamedBassem:
Thank you.

PR #2659, opened 2026-04-08
feat(workers): make monolith timeout and arguments configurable

### LINKED ISSUE #1437: Increase monolith timeout?

Currently, the monolith timeout is hardcoded to 5sec (FTR: monolith's default is 120sec). I feel this leads to many unneccessary timeouts when adding bookmarks from slow servers (some home servers, sometimes archive.org).

It would be nice if the timeout could be increased and/or made configurable (like all the crawler timeouts).

Make it easier to bookmark pages from slow servers.

Nope, since the timeout is hardcoded.

### Additional context

_No response_

### LINKED ISSUE #1686: Make Monolith arguments configurable

Please add a configuration parameter/environment variable for Monolith arguments.  Ideally it will fully replace the arguments Karakeep gives by default and include documentation on the default args (excluding anything required like `-` to output to stdout).

This will allow users to leave out (or include) any content they'd like from archived webpages, for example webfonts, audio, (inline) images.  This should help mitigate/fix archived pages becoming very large: https://github.com/karakeep-app/karakeep/issues/770#issuecomment-3015919241

Not currently.

### Additional context

See https://github.com/karakeep-app/karakeep/issues/770#issuecomment-3015919241

NikhilChowdhury27:
Closes #1437
Closes #1686

## Summary

Right now the monolith timeout is hardcoded to 5 seconds — which is pretty
aggressive considering monolith's own default is 120s. This causes archival
failures on slower servers (home servers, archive.org, etc). There's also no
way to customize the monolith arguments, so users can't control things like
whether web fonts or audio get included in archives.

This adds two new env vars following the same pattern as `CRAWLER_YTDLP_ARGS`:

- `CRAWLER_MONOLITH_TIMEOUT_SEC` (default: 5) — makes the timeout configurable
- `CRAWLER_MONOLITH_ARGS` (default: []) — pass extra args to monolith, `%%` separated

Default behavior is unchanged — existing users won't notice anything different.

## Example

```env
CRAWLER_MONOLITH_TIMEOUT_SEC=30
CRAWLER_MONOLITH_ARGS=--no-fonts%%--no-audio
```

## What changed

- `packages/shared/config.ts` — zod schema + config mapping
- `apps/workers/workers/crawlerWorker.ts` — use config instead of hardcoded values
- `docs/.../01-environment-variables.md` — documented both vars

## Testing

- Typecheck, lint, format, tests all pass
- Config loads correctly with defaults (backward compatible) and custom values
- Tested monolith binary directly with different timeout and arg combinations


MohamedBassem:
Thank you!

PR #2674, opened 2026-04-08
fix(crawler): skip getContentType network call when precrawled archive is provided

### LINKED ISSUE #2579: [Bug] SingleFile uploads incorrectly trigger HTTP GET on original URL, causing missing titles and crawling failures
### Describe the Bug

When uploading a packaged HTML file via the /api/v1/bookmarks/singlefile endpoint (e.g., using the SingleFile extension), the backend still attempts to make an HTTP GET request to the original URL to determine its content type.

If the original website blocks crawlers, requires authentication, or is otherwise inaccessible, this pre-check fails. As a result, the crawler's parsing logic breaks, causing the bookmark to lose its title and other metadata, even though a perfectly valid HTML file was successfully uploaded.

### Steps to Reproduce

Configure the SingleFile extension to upload to the /api/v1/bookmarks/singlefile endpoint as per the official documentation.
Visit a webpage that actively blocks crawlers or requires being logged in.
Use the SingleFile extension to package and upload the page to Karakeep.
Check the newly created bookmark in Karakeep.

### Expected Behaviour

The system should recognize that a precrawledArchiveAssetId (the uploaded HTML file) was provided and entirely skip making any network requests to the original URL. The uploaded HTML file should be parsed directly to extract the correct title, description, and content.

I looked into the source code and found the root cause. This issue was introduced in commit `2743d9e` (#1163), which added support for saving direct image/PDF bookmarks.

In `apps/workers/workers/crawlerWorker.ts`, inside the `runCrawler` function, the crawler unconditionally executes `getContentType(url)`:

```typescript
// crawlerWorker.ts - runCrawler function
const contentType = await getContentType(url, jobId, job.abortSignal);
```

This HTTP `GET` request fires *before* the system checks for `precrawledArchiveAssetId` inside `crawlAndParseUrl` .

When `getContentType` hits an anti-bot restriction, it affects the crawling pipeline, even though `crawlAndParseUrl`itself correctly tries to read the `precrawledArchiveAssetId`.

### Device Details

_No response_

### Exact Karakeep Version

v0.31.0

### Environment Details

Docker

### Debug Logs

_No response_

howwohmm:
## Problem

When uploading a SingleFile HTML archive, the crawler still made an outbound HTTP GET to the original URL to determine content type — even though `precrawledArchiveAssetId` was already set, meaning the archive was already available locally.

This caused metadata loss for auth-gated or crawler-blocking sites: the pre-check HTTP request would fail, causing the bookmark to lose its title and other metadata, even though the uploaded archive was perfectly valid.

Root cause identified by reporter in #2579: `getContentType(url, ...)` is called unconditionally at line 2093 of `crawlerWorker.ts`, before checking whether `precrawledArchiveAssetId` is set.

## Fix

Guard the `getContentType` call so it is skipped when `precrawledArchiveAssetId` is set:

```typescript
// Before
const contentType = await getContentType(url, jobId, job.abortSignal);

// After
const contentType = precrawledArchiveAssetId
  ? null
  : await getContentType(url, jobId, job.abortSignal);
```

`precrawledArchiveAssetId` is already in scope (destructured at line 2084). A `null` content type safely falls through to the `else` branch that calls `crawlAndParseUrl`, which already has first-class handling for precrawled archives — it reads the uploaded HTML directly and skips all network requests.

## Scope

1 file, 3-line diff. No existing test infrastructure for `crawlerWorker.ts`.

Fixes #2579

### DIFF
--- a/apps/workers/workers/crawlerWorker.ts
+++ b/apps/workers/workers/crawlerWorker.ts
@@ -2090,7 +2090,9 @@ async function runCrawler(
     `[Crawler][${jobId}] Will crawl "${url}" for link with id "${bookmarkId}"`,
   );
 
-  const contentType = await getContentType(url, jobId, job.abortSignal);
+  const contentType = precrawledArchiveAssetId
+    ? null
+    : await getContentType(url, jobId, job.abortSignal);
   job.abortSignal.throwIfAborted();
 
   // Link bookmarks get transformed into asset bookmarks if they point to a supported asset instead of a webpage


MohamedBassem:
This has already been fixed in another PR. Thanks.

PR #2675, opened 2026-04-08
fix(crawler): strip Twitter/X reply articles before Readability extraction

### LINKED ISSUE #2651: Twitter/X: Crawler extracts unrelated reply content into reader view
### Describe the Bug

When bookmarking a single tweet from X (Twitter) with cookies configured, the crawler's content extraction picks up reply/comment content from below the original tweet, mixing unrelated text into the reader view and the indexed content.
For example, bookmarking a tweet by @rusino777 saying something about chord progression, the reader view includes replies from another user posting a meme image and tons of emoji images.
These are other users' replies, not the original tweet content. This pollutes:
Reader view — shows irrelevant text
Full-text search index — replies become searchable as if they were the bookmarked content
AI tagging / summarization — tags and summaries are generated from reply content rather than the original tweet

### Steps to Reproduce

1. Configure browser cookies for x.com / twitter.com in Karakeep settings
2. Bookmark any popular tweet URL (one with visible replies), e.g. https://x.com/rusino777/status/2015243512728707399
3. Wait for the crawling job to complete
4. Open the bookmark and switch to Reader View
5. Observe that reply content from other users appears below the original tweet text

### Expected Behaviour

The extracted content (reader view / indexed text) should contain only the original tweet's text and media, not replies from other users (Except user specify it explicitly, like #1468). 

Image
A Tweet by ルシノ, talking about chord progression

Image

Image
Reader view, showing the original tweet text followed by unrelated reply content and emoji image

### Device Details

_No response_

### Exact Karakeep Version

v0.31.0

### Environment Details

Docker on Mac Mini 2, 26.0.1

### Debug Logs

_No response_

howwohmm:
## Problem

When bookmarking a tweet with browser cookies configured, Mozilla Readability treats reply threads as part of the article body. This pollutes reader view, the full-text search index, and AI tagging with content from other users' replies — not just the original tweet.

## Root Cause

Twitter/X renders replies as additional `article[data-testid="tweet"]` elements in the DOM after the main tweet. Readability has no way to distinguish the original tweet from replies, so it extracts everything.

## Fix

Add a `stripTwitterReplies` preprocessing step in `extractReadableContent` that:
1. Detects Twitter/X URLs by hostname
2. Queries all `article[data-testid="tweet"]` elements in the JSDOM document
3. Removes all but the first (the original tweet) before Readability runs

```typescript
function stripTwitterReplies(document: Document, url: string): void {
  const hostname = new URL(url).hostname;
  if (!hostname.includes("twitter.com") && !hostname.includes("x.com")) return;
  const articles = document.querySelectorAll('article[data-testid="tweet"]');
  for (let i = 1; i < articles.length; i++) articles[i].remove();
}
```

This runs before `new Readability(dom.window.document).parse()`, ensuring only the original tweet's text and media is extracted.

## Scope

1 file, 17-line addition. No behavior change for non-Twitter URLs.

Fixes #2651

### DIFF
--- a/apps/workers/scripts/parseHtmlSubprocess.ts
+++ b/apps/workers/scripts/parseHtmlSubprocess.ts
@@ -120,6 +120,22 @@ function normalizeLazyLoadImages(document: Document): void {
   }
 }
 
+function stripTwitterReplies(document: Document, url: string): void {
+  try {
+    const hostname = new URL(url).hostname;
+    if (!hostname.includes("twitter.com") && !hostname.includes("x.com")) {
+      return;
+    }
+  } catch {
+    return;
+  }
+  // Keep only the first tweet article element; remove all replies that follow
+  const articles = document.querySelectorAll('article[data-testid="tweet"]');
+  for (let i = 1; i < articles.length; i++) {
+    articles[i].remove();
+  }
+}
+
 function extractReadableContent(
   htmlContent: string,
   url: string,
@@ -128,6 +144,7 @@ function extractReadableContent(
   const dom = new JSDOM(htmlContent, { url, virtualConsole });
   try {
     normalizeLazyLoadImages(dom.window.document);
+    stripTwitterReplies(dom.window.document, url);
     const readableContent = new Readability(dom.window.document).parse();
     if (!readableContent || typeof readableContent.content !== "string") {
       return null;


NachtgeistW:
Yes, as the bots said, it will catch hostname like "xbox.com""lax.com" unexpectedly. I'd like to suggest to use a function `IsTwitterHost` to catch Twitter hostname exactly.

PR #2685, opened 2026-04-10
feat(mobile): revamp search and navigation chrome

esimkowitz:
## Summary

Revamps mobile search with platform-native treatments and cleans up surrounding navigation chrome (FAB, settings, tab layout).

### Search

- **iOS**: adds a dedicated Search tab using Expo Router's `headerSearchBarOptions`, which gives us the native `UISearchController` large-title search bar (and the Liquid Glass search tab role on iOS 26).

simulator_screenshot_8ED489FD-588D-4D68-AB3B-AEF71E83FD3D

- **Android**: the Search tab is hidden. Home, Lists, Highlights, and Tags now show an `AndroidSearchBar` pill at the top that expands into a full-screen `InlineSearch` overlay. Tags reuses the same pattern to toggle its existing tag filter into focused mode.
image

- Extracts `BookmarkSearchResults` from `app/dashboard/search.tsx` so the tRPC infinite query, debounced input, and recent-searches history are shared across the iOS tab, the Android overlay, and the legacy search route.
- The legacy `dashboard/search.tsx` is slimmed down and gains an iOS `formSheet` presentation with a custom header search input.

### FAB

- New `components/ui/FAB/` with platform splits. iOS uses `expo-glass-effect`'s `GlassView` on iOS 26 and falls back to `expo-blur` `BlurView` on earlier versions; Android uses a Material 3 primary-colored elevated button.
- The Home tab's "New bookmark / Photo library" menu and the Lists tab's "New list" button move out of the header and into the FAB.
- Adds `expo-blur` and `expo-glass-effect` to `apps/mobile/package.json`.

### Settings

- Demoted from a top-level tab to a modal: `formSheet` on iOS, `modal` on Android, with a sheet grabber.
- `(tabs)/(settings)/` is removed; the screen lives at `dashboard/settings/index.tsx` and is registered in the dashboard stack.
- New `ProfileAvatarButton` (user avatar fetched via `users.whoami`) opens the settings modal. Placed in the iOS header right slot and as the `rightElement` of the Android search bar.

### Navigation cleanup

- New `lib/tabScreenOptions.ts` extracts the ~20-line `screenOptions` block (iOS large transparent header, Android `paddingBottom: 100` workaround for SDK 54 tab-bar overlap) that was duplicated across five tab layouts.
- Android tab bar now uses `labelVisibilityMode="labeled"` so inactive tabs keep their labels.

### Drive-by fix

The Home photo-library upload flow now shows a loading toast as soon as the picker opens, dismisses it on cancel, and rewrites it to "Uploading..." once an asset is picked. The error path no longer references a null toast ID.

## Test plan

- [ ] iOS: Search tab shows the native large-title search bar; typing debounces; recent searches render when focused with empty query; results render when focus is lost and query is non-empty.
- [ ] iOS 26: Search tab renders as the Liquid Glass search role; FAB uses `GlassView`.
- [ ] iOS < 26: FAB falls back to `BlurView`.
- [ ] Android: Home, Lists, Highlights, Tags all show the `AndroidSearchBar` pill; tapping opens `InlineSearch` full-screen; closing returns to the tab with scroll position preserved.
- [ ] Android: Tags search pill toggles the tag filter into focused mode and back.
- [ ] Both platforms: FAB opens the "New bookmark / Photo library" menu on Home and creates a new list on Lists.
- [ ] Both platforms: tapping the profile avatar opens the settings modal; dismissing returns to the previous tab.
- [ ] Both platforms: photo library upload shows "Opening photo library..." toast, rewrites to "Uploading..." on pick, dismisses on cancel.
- [ ] `pnpm typecheck` clean.
- [ ] `pnpm lint` clean.


apps/mobile/package.json
```diff
@@ -39,12 +39,14 @@
     "clsx": "^2.1.0",
     "date-fns": "^3.6.0",
     "expo": "~54.0.31",
+    "expo-blur": "~55.0.7",
```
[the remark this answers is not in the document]
esimkowitz: this is fine

apps/mobile/components/search/BookmarkSearchResults.tsx
```diff
@@ -0,0 +1,138 @@
...
+  }, [rawQuery, history]);
+
+  if (error) {
+    return <FullPageError error={error.message} onRetry={() => refetch()} />;
+  }
+
+  const handleSelectHistory = (term: string) => {
+    addTerm(term);
+    onSelectHistory(term);
+  };
```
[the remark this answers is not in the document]
esimkowitz: Fixed — added a `useEffect` in `BookmarkSearchResults` that calls `addTerm(rawQuery.trim())` when `isInputFocused` transitions from true→false with non-empty input. The iOS search tab (`index.ios.tsx`) already had this in its `onBlur` handler.

apps/mobile/components/search/BookmarkSearchResults.tsx
```diff
@@ -0,0 +1,138 @@
...
+            No matching searches.
+          </Text>
+        }
+        keyboardShouldPersistTaps="handled"
+      />
+    );
+  }
+
+  if (isFetching && query.length > 0) {
+    return <FullPageSpinner />;
```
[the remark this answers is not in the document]
esimkowitz: Fixed — changed `isFetching` to `isPending` so `keepPreviousData` works as intended. Applied in both `BookmarkSearchResults.tsx` and `index.ios.tsx`.

apps/mobile/app/dashboard/search.tsx
```diff
@@ -1,157 +1,67 @@
...
-          onRefresh={onRefresh}
-          isRefreshing={isPending}
+      {Platform.OS === "ios" ? (
+        <Stack.Screen
+          options={{
+            headerTitle: () => searchInput,
+            headerRight: () => (
+              <Pressable onPress={() => router.back()}>
+                <X size={22} color={colors.foreground} />
+              </Pressable>
```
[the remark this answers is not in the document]
esimkowitz: Fixed — added `accessibilityLabel="Close"` and `accessibilityRole="button"` to the close Pressable in both `search.tsx` and `InlineSearch.tsx`.

apps/mobile/components/search/BookmarkSearchResults.tsx
```diff
@@ -0,0 +1,138 @@
...
+    api.bookmarks.searchBookmarks.infiniteQueryOptions(
+      { text: query },
+      {
+        placeholderData: keepPreviousData,
+        gcTime: 0,
+        initialCursor: null,
+        getNextPageParam: (lastPage) => lastPage.nextCursor,
+      },
+    ),
+  );
```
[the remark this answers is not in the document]
esimkowitz: Fixed — added `enabled: query.trim().length > 0` to both `BookmarkSearchResults.tsx` and `index.ios.tsx` so the search query only fires when there's real input.

apps/mobile/components/search/BookmarkSearchResults.tsx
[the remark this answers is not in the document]
esimkowitz: Fixed — added a `useEffect` that detects the `isInputFocused` true→false transition and calls `addTerm(rawQuery.trim())` for non-empty queries. This keeps history logic co-located with `useSearchHistory` in the component.

apps/mobile/components/settings/ProfileAvatarButton.tsx
```diff
@@ -0,0 +1,37 @@
+import { Pressable, View } from "react-native";
+import { useRouter } from "expo-router";
+import { Avatar } from "@/components/ui/Avatar";
+import { useQuery } from "@tanstack/react-query";
+
+import { useTRPC } from "@karakeep/shared-react/trpc";
+
+const AVATAR_SIZE = 28;
+const MARGIN = 2;
+const CONTAINER_SIZE = AVATAR_SIZE + MARGIN * 2;
```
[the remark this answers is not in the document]
esimkowitz: Fixed — increased `CONTAINER_SIZE` from 32px to 44px (minimum recommended touch target), added `accessibilityLabel="Open profile settings"` and `accessibilityRole="button"` to the Pressable, and converted static centering/rounding styles to Tailwind classes.

apps/mobile/components/ui/AndroidSearchBar.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+export default function AndroidSearchBar({
+  label,
+  onPress,
+  rightElement,
+}: {
+  label: string;
+  onPress: () => void;
+  rightElement?: React.ReactNode;
+}) {
```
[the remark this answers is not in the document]
esimkowitz: Dismissing — `React.ReactNode` without an explicit React import is the consistent convention across this codebase (used in 5 files: `TailwindResolver.tsx`, `InlineSearch.tsx`, `providers.tsx`, `GroupedList.tsx`, and here). The Expo TS config uses the `react-jsx` transform which makes React types globally available. Changing just this file would be inconsistent.

apps/mobile/app/dashboard/(tabs)/(search)/index.ios.tsx
```diff
@@ -0,0 +1,169 @@
...
+          headerSearchBarOptions: {
+            ref: searchBarRef,
+            placeholder: "Search bookmarks...",
+            onChangeText: (event) => setSearch(event.nativeEvent.text),
+            onFocus: () => setIsSearchFocused(true),
+            onBlur: () => {
+              setIsSearchFocused(false);
+              if (search.trim().length > 0) {
+                addTerm(search);
+              }
```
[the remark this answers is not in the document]
esimkowitz: Good catch — fixed in 3fb1acb. The `onBlur` handler now computes `const normalized = search.trim()` and passes that to `addTerm`, mirroring `handleSearchSubmit`.

apps/mobile/app/dashboard/(tabs)/(highlights)/index.tsx
```diff
@@ -39,12 +48,20 @@ export default function Highlights() {
...
-      isFetchingNextPage={isFetchingNextPage}
-      isRefreshing={isPending || isPlaceholderData}
-    />
+    <>
+      {Platform.OS === "android" && (
+        <AndroidSearchBar
+          label="Search bookmarks..."
+          onPress={() => setSearchActive(true)}
+        />
+      )}
```
MohamedBassem, line 57: Doesn't feel like we should be showing the search bar in the highlights page and the lists page. Doing it in home is good enough and not very different than what we have today?

apps/mobile/app/dashboard/(tabs)/(lists)/index.tsx
```diff
@@ -186,25 +185,23 @@ export default function Lists() {
...
-            />
-          ),
-        }}
-      />
+      {Platform.OS === "android" && (
+        <AndroidSearchBar
+          label="Search bookmarks..."
+          onPress={() => setSearchActive(true)}
+        />
+      )}
```
MohamedBassem, line 193: same comment with regards to inline search here

apps/mobile/app/dashboard/(tabs)/(lists)/index.tsx
```diff
@@ -186,25 +185,23 @@ export default function Lists() {
...
       <FlatList
         className="h-full"
         contentInsetAdjustmentBehavior="automatic"
         contentContainerStyle={{
-          gap: 6,
-          paddingBottom: 20,
+          gap: 15,
```
MohamedBassem, line 198: Don't you think the tighter gaps look better?

apps/mobile/app/dashboard/(tabs)/(search)/index.ios.tsx
```diff
@@ -0,0 +1,170 @@
...
+    if (term.length > 0) {
+      addTerm(term);
+      setSearch(term);
+    }
+    Keyboard.dismiss();
+  };
+
+  const renderHistoryItem = ({ item }: { item: string }) => (
+    <Pressable
+      onPress={() => handleSearchSubmit(item)}
```
MohamedBassem, line 88: This seems to be broken. Clicking on a search history item doesn't seem to fill the search bar with that term, and causing some weird animation where it disappears from the recent searches before the result shows up (maybe missing some spinner). Also it's unclickable when you have only a single item.

apps/mobile/app/dashboard/(tabs)/(search)/index.ios.tsx
```diff
@@ -0,0 +1,170 @@
...
+  return (
+    <>
+      <Stack.Screen
+        options={{
+          headerSearchBarOptions: {
+            ref: searchBarRef,
+            placeholder: "Search bookmarks...",
+            onChangeText: (event) => setSearch(event.nativeEvent.text),
+            onFocus: () => setIsSearchFocused(true),
+            onBlur: () => {
```
MohamedBassem, line 153: I'm finding the behavior of the dismiss button a bit unintuitive and doesn't align with the rest of ios. When you type something in the search bar there are two "X"s that appear. One that clears the search bar and takes us back to the recent searches, and another big one that in the rest of iOS takes us "out" of search completely (in this PR it also kinda clears the search bar, which is a bit weird). What do you think?

apps/mobile/app/dashboard/(tabs)/(tags)/index.tsx
```diff
@@ -75,68 +77,98 @@ export default function Tags() {
...
+            <SearchInput
+              containerClassName="mb-2"
+              placeholder="Search tags..."
+              value={searchQuery}
+              onChangeText={setSearchQuery}
+            />
+          ) : undefined
+        }
+        contentContainerStyle={{
+          gap: 15,
```
MohamedBassem, line 118: same comments about the gaps as with the lists

apps/mobile/app/dashboard/(tabs)/(tags)/index.tsx
```diff
@@ -75,68 +77,98 @@ export default function Tags() {
...
+            <SearchInput
+              placeholder="Search tags..."
+              value={searchQuery}
+              onChangeText={setSearchQuery}
+              autoFocus
+              onCancel={() => {
+                setSearchActive(false);
+                setSearchQuery("");
+              }}
+            />
```
MohamedBassem, line 96: For the tags search in particular, do you think the divergence from the ios approach is worth it? I feel like it looked pretty reasonable already on both ios and android.

apps/mobile/app/dashboard/search.tsx
```diff
@@ -1,157 +1,71 @@
-import { useMemo, useRef, useState } from "react";
-import { FlatList, Keyboard, Pressable, TextInput, View } from "react-native";
-import { router } from "expo-router";
-import BookmarkList from "@/components/bookmarks/BookmarkList";
-import FullPageError from "@/components/FullPageError";
-import FullPageSpinner from "@/components/ui/FullPageSpinner";
+import { useRef, useState } from "react";
```
MohamedBassem, line 1: Is this page still used anywhere?

MohamedBassem (changes requested):
As usual, I love what you've done here. Overall, looks great. I left a bunch of comments. The main ones being:
1. Not a big fan of the "search bookmark" tab showing on android in other tabs beyond the home page. Once we do this, two other followups:
  a. we should restore the header title for the pages after we remove the search from them.
  b. For the tags page, I feel like how it looked before the change was already reasonable on both ios and android.
2. The search behavior on iOS feels a bit unintuitive, left the comments on the code itself.

apps/mobile/app/dashboard/(tabs)/(highlights)/index.tsx
esimkowitz: I was following the examples I was seeing in some other Android apps, where they had a persistent search bar at the top of the app in all pages. I agree it's not my favorite either, but I was trying to find something similar to the always-available search functionality we'll now have on iOS

apps/mobile/app/dashboard/(tabs)/(search)/index.ios.tsx
esimkowitz: Screenshot 2026-04-23 at 8 39 19 PM
This is the way that i've seen it handled in other apps, namely Apple News and 1Password. Do you have some examples of apps that handle the new search tab differently?

apps/mobile/app/dashboard/(tabs)/(lists)/index.tsx
esimkowitz: I think you're right, I don't really remember why I changed this, will revert.

apps/mobile/app/dashboard/search.tsx
esimkowitz: no, will remove

apps/mobile/app/dashboard/(tabs)/(search)/index.ios.tsx
esimkowitz: ah, the view was just never switching. should be fixed now

apps/mobile/app/dashboard/(tabs)/(tags)/index.tsx
esimkowitz: you mean the styling of the "search tags..." search bar on Android? I think this is in line with the comment about adding the search bar to the top of the app on all tabs on Android, this was a behavior I'd observed in some of the native apps like the phone and contacts apps, but I can definitely revert this and keep the search bar for the tags tab more similar on each platform.

apps/mobile/app/dashboard/(tabs)/(highlights)/index.tsx
MohamedBassem: Yeah, I see what you mean in the phone and contacts apps. Seems like it's also there in google home, and google drive. At the same time, not all apps seem to be adhering to this. For example, play store doesn't do it (and has a dedicated search in the tab bar). Youtube maintains a constant header but search is collapsed by default. Github doesn't have a persistent header across tabs. Same with bitwarden. 

Checked material 3 guidelines (here: https://m3.material.io/components/search/guidelines#941d0bf7-2337-4cc7-8a5a-ac3f63c1de62). Seems like there are three modes. In-view search, global search, and when search is a secondary action. I feel like given that both of us are finding it weird, we should treat the bookmark search as an in-view search rather than a global one. This is also why the tags search feels a bit weird in the current layout, it's placed as a global search rather than an in-page search.

Let me take a stab at re-organizing this.

MohamedBassem:
Did some of the changes. In android, reverted the inline search to only be in the home page. The rest of the pages look as they used to look before. And I added a search-on-type which I think improves the UX quite a bit in the new mode. Thanks a lot @esimkowitz for driving this, the mobile app is looking more and more native thanks to your work!

PR #2693, opened 2026-04-11
feat: add manual save toggle for browser extension

### LINKED ISSUE #2692: feat: Add option to disable auto saving of bookmarks in browser extensions

I would like an option to disable the auto-save bookmark on click workflow in the existing browser app, instead requiring a manual confirmation/save click to actually save the bookmark. 

For users coming from Raindrop this would provide a more familiar workflow than the current auto-save functionality. 

Would also help people like me who have poor mouse targeting from adding sites they don't mean to.

Not any way I can think of

### Additional context

_No response_

keslerm:
I tried to keep this intentionally light only including the options that can be saved on the initial create bookmark call (no tags/lists) but could add those as well.

Fixes https://github.com/karakeep-app/karakeep/issues/2692


MohamedBassem:
Thank you!

Issue #2612, opened 2026-03-22
Feature Request: Allow configuring reasoning behavior (e.g. disable reasoning) for LLM calls


dennisvanderpool:
### Describe the feature you'd like

Karakeep currently does not provide a way to control reasoning behavior (e.g. enabling/disabling chain-of-thought / reasoning tokens) when making LLM calls.

This causes issues when using reasoning-capable models (e.g. Qwen3.5 via vLLM), especially in combination with structured output (JSON schema).

Problem

When using Karakeep with a reasoning-capable model:

The model generates reasoning tokens first
Then generates the final answer

However, Karakeep also:

Enforces a max_tokens limit (e.g. 2048)
Uses response_format: json_schema with strict: true

This leads to the following failure mode:

The model consumes all tokens on reasoning
No tokens remain for the final structured output
The response returns:
content: null
finish_reason: "length"
Example (real trace)
prompt_tokens: ~2244
completion_tokens: 2048
content: null
reasoning_content: present

Result: No usable output despite successful call

Root Cause

Karakeep does not expose or allow passing reasoning-related parameters such as:

"reasoning": { "effort": "none" }

Without this, reasoning-enabled models default to generating chain-of-thought, which:

wastes token budget
breaks structured output
increases latency
Proposed Solution

Allow users to configure reasoning behavior.

### Describe the benefits this would bring to existing Karakeep users

This feature would significantly improve reliability, performance, and compatibility of Karakeep’s LLM integrations, especially as newer reasoning-capable models become more common.

### Can the goal of this request already be achieved via other means?

I use LiteLLM in between, I might control reasoning behaviour there for the karakeep virtual key.

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

Laephis:
Apologies if this is a stupid question, but does this issue explain why attempting to use qwen3.5:9b as a model does not seem to work at all for tagging?

dennisvanderpool:
@Laephis afaik yes. Try qwen3 VL?  I am using that one, works fine.

cinconueves:
I'm working on this.

Fmstrat:
@cinconueves Thank you! Any ETA on when we think this will work? Karakeep is the only app that can't use Qwen3.5 in my system setup.

EDIT: I see the pull was merged, but I'm still having issues in the latest release, so I'm assuming it hasn't made it's way through yet.

dchakro:
@Fmstrat 

[...]

PR #2718, opened 2026-04-17
Add support for configuring OpenAI reasoning effort

cinconueves:
- Addresses #2612 
- Adds a reasoning effort config var support, passes that along to the OpenAI call.
- Add edge case handling for reasoning models that might return the response in the reasoning field.


cinconueves:
@MohamedBassem q: on your changes to the branch, having reasoning passed in always, even if undefined, makes it possible to pass an undefined value field to the request, doesn't it? Which might lead to errors, depending on how strict the end APIs are. This for Open router/non OpenAI yet OpenAI compatible endpoints. 

Similarly, still, some reasoning models _might_ return their  response in the reasoning field. I think it's a fair edge case. 

Is within the real of possibility making the LLM integration more flexible in terms of provider support? i.e openRouter has a sdk which might be more flexible than the Open AI one. Maybe adding a conditional branch for Open router requests to better support these edge cases.

MohamedBassem:
@cinconueves typically, passing undefined to such APIs are equivalent to not passing the value at all. So I think semantically, it'll end up being the same.

As for the model reasoning field, I'd consider this to be an edge case, that I'm not comfortable supporting generically, specially with the typescript error silencing.

As for the LLM integration, I'm trying to keep that slim as much as possible given that our needs are minimal. I already regret having ollama as a separate SDK versus using the openai compatible API that they have. Each new branch adds maintenance burden for me. That's why I'd rather only have the openai compatible API. I might in the future consider using something like the AI SDK from vercel to abstract that up, but for now, I'd like to keep it minimal.

cinconueves:
> @cinconueves typically, passing undefined to such APIs are equivalent to not passing the value at all. So I think semantically, it'll end up being the same.
> 
> As for the model reasoning field, I'd consider this to be an edge case, that I'm not comfortable supporting generically, specially with the typescript error silencing.
> 
> As for the LLM integration, I'm trying to keep that slim as much as possible given that our needs are minimal. I already regret having ollama as a separate SDK versus using the openai compatible API that they have. Each new branch adds maintenance burden for me. That's why I'd rather only have the openai compatible API. I might in the future consider using something like the AI SDK from vercel to abstract that up, but for now, I'd like to keep it minimal.

Thanks, that makes sense.

PR #2749, opened 2026-04-30
feat: redesign list and tag page headers

MohamedBassem:
Adds more info about the list/tag (such as item count). For lists, it also shows whether it's private/shared/public, and if it's a smart list, the matchers are shown there as well.


MohamedBassem:
@codex review
@greptile review

PR #2829, opened 2026-05-20
chore: Add seed snapshot generator

MohamedBassem:
This script generates some seed data and commits it to the repo with the ability to re-apply it again if needed. Those snapshots are useful for worktrees and sandboxes to quickly seed the system with useable data. In the future, they can also be used for backward compatibility tests. 

The script works by building the app's docker image, and talking to the APIs


MohamedBassem:
@greptile review
@codex review

PR #2833, opened 2026-05-20
chore: upgrade to pnpm 11

MohamedBassem:


dotlambda:
Can we have a release with this change as pnpm 9 has reached EOL? See https://endoflife.date/pnpm.

Issue #2593, opened 2026-03-17
Feature Request: Add Flaresolver support


T-REX-XP:
### Describe the feature you'd like

Add the ability to connect the FlareSolver proxy to solve the Cloudflare capcha.
- https://github.com/FlareSolverr/FlareSolverr

### Describe the benefits this would bring to existing Karakeep users

Karakeep will be able to scrape the web pages that are protected with Cloudflare

### Can the goal of this request already be achieved via other means?

no

### Have you searched for an existing open/closed issue?

- [x] I have searched for existing issues and none cover my fundamental request

NotaInutilis:
Yes! That would be great! Byparr is another tool that could be used for that goal https://github.com/ThePhaseless/Byparr

Issue #2423, opened 2026-01-23
Cloudflare blocking karakeep downloads


grimwiz:
### Describe the Bug

Some of my RSS feeds are showing errors in the captured page view and summary. I suspect the browser header has somehow got blacklisted. The text is:

Cloudflare has blocked your access to the site because a security rule was triggered. The page explains that this can happen when you submit certain words, phrases, SQL commands, or malformed data. It advises you to contact the site owner with details of your activity and the Cloudflare Ray ID displayed at the bottom. No additional numbered items are present in the title.

I'm attaching what the pages look like in the karakeep web gui. I can still read the title but cached content, summary and tags are missing. I don't see any mention of a Ray ID in the karakeep bookmark information.

<img width="888" height="394" alt="Image" src="https://github.com/user-attachments/assets/e1d7a735-9160-4e36-b9e9-5f60607c4fd1" />

### Steps to Reproduce

Set up an RSS feed, for example : http://www.theregister.co.uk/security/headlines.atom
wait for some pages to appear.
I think they appear as broken links, error code 403.

### Expected Behaviour

Up until a few days ago everything was working fine, with tags, summaries and pictures of the webpage.

### Screenshots or Additional Context

<img width="1279" height="62" alt="Image" src="https://github.com/user-attachments/assets/54ce9a7c-1922-4b10-a9c4-a555316351a8" />
I'm able to click through and see the article without issue, so I know that my IP address has not been blocked and that the site is functioning.

### Device Details

N/A

### Exact Karakeep Version

v0.30.0

### Environment Details

Karakeep is running in a LXC container on Proxmox, Installation was via Proxmox VE Helper-Scripts.

### Debug Logs

Jan 23 22:26:21 karakeep pnpm[31678]: 2026-01-23T22:26:21.529Z info: <-- GET /api/trpc/users.settings,users.whoami,bookmarks.getBrokenLinks?batch=1&input=%7B%220%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%2C%221%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%2C%222%22%3A%7B%22json%22%3Anull%2C%22meta%22%3A%7B%22values%22%3A%5B%22undefined%22%5D%7D%7D%7D
[...]

Issue #2381, opened 2026-01-12
"Enable JavaScript and cookies to continue"


obarlev:
I’m getting this message when trying to crawl a website. I checked the Karakeep configuration section and couldn’t find any setting that seems to control it: 
https://docs.karakeep.app/configuration/environment-variables/

BTW, I've successfully crawled many pages from that website using prior Karakeep versions. I'm not sure if something changed in Karakeep or on the website itself.

Do you have any advice?

<img width="755" height="462" alt="Image" src="https://github.com/user-attachments/assets/5af64a5e-658c-4853-a482-523f30e5548b" />

<img width="781" height="388" alt="Image" src="https://github.com/user-attachments/assets/b52c8562-8f7f-4258-a671-b587aadd193e" />

Laimonas-Mituzas:
Cloudflare?

MrKernelPanic-dev:
Hello,  
I encountered the same issue.

In my case, it was caused by the crawler failing to resolve the browser container hostname, as shown by the error in the logs:

```
web-1          | 2026-01-22T13:04:50.238Z error: [Crawler] Failed to connect to the browser instance, will retry in 5 secs: Error: getaddrinfo ENOTFOUND chrome
web-1          |     at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:internal/dns/promises:88:17)
```

The root cause was a misconfiguration in my `docker-compose.yml`: 
the `web` service could not reach the `chrome` service (DNS/hostname resolution issue). After fixing the Docker networking/service configuration, the crawler worked correctly.

Other potential causes are also discussed in this related issue: https://github.com/karakeep-app/karakeep/issues/674

Issue #2073, opened 2025-10-26
Page not hoarded


anupamck:
### Describe the Bug

When I try to hoard the following page: https://www.jamesshore.com/v2/blog/2025/the-accountability-problem, it doesn't get hoarded. 

### Steps to Reproduce

1. Hoard this page https://www.jamesshore.com/v2/blog/2025/the-accountability-problem

### Expected Behaviour

When I preview the hoarded page on my Karakeep server, I should be able to read a preview of it. However, said preview is absent. 

### Screenshots or Additional Context

Here's the preview on my self-hosted server. 

<img width="1727" height="836" alt="Image" src="https://github.com/user-attachments/assets/92cf2539-7071-4830-9960-6546c24793d2" />

Here are the logs on my server (they appear to be fine):
``` 
web-1  | 2025-10-26T18:33:11.586Z info: [Crawler][19076] Navigating to "https://www.jamesshore.com/v2/blog/2025/the-accountability-problem"
web-1  | 2025-10-26T18:33:12.331Z info: [Crawler][19076] Successfully navigated to "https://www.jamesshore.com/v2/blog/2025/the-accountability-problem". Waiting for the page to load ...
web-1  | 2025-10-26T18:33:13.345Z info: [Crawler][19076] Finished waiting for the page to load.
web-1  | 2025-10-26T18:33:13.363Z info: [Crawler][19076] Successfully fetched the page content.
web-1  | 2025-10-26T18:33:13.581Z info: [Crawler][19076] Finished capturing page content and a screenshot. FullPageScreenshot: false
web-1  | 2025-10-26T18:33:13.599Z info: [Crawler][19076] Will attempt to extract metadata from page ...
web-1  | 2025-10-26T18:33:13.659Z info: [Crawler][19076] Will attempt to extract readable content ...
web-1  | 2025-10-26T18:33:13.956Z info: [Crawler][19076] Done extracting readable content.
web-1  | 2025-10-26T18:33:14.031Z info: [Crawler][19076] Stored the screenshot as assetId: a072ff23-51f0-4f5b-ba79-81f128927325
``` 

### Exact Karakeep Version

0.27.1

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

cloudchristoph:
Interesting.

Tried to hoard this page on my system.
Error log shows a timeout.

I even tried another blog post from the page and the homepage itself.
[...]

PR #2849, opened 2026-05-27
docs(proposal): Firefox-based stealth backend for crawler worker

feder-cr:
opening as draft to check interest before building the env-resolver branch.

would an optional Firefox-based stealth backend for `crawlerWorker.ts` be in scope, parallel to the current `playwright-extra` + `StealthPlugin()` chromium path? selected via env, no change to defaults.

motivation: the stealth plugin patches a fixed set of JS-visible signals by injecting overrides into every context. Cloudflare and current anti-bot WAFs increasingly flag the override surface itself, which is why this path gets blocked on the sites tracked in #2381, #2423, #2073 and is the underlying reason behind the #2593 Flaresolver request. a Firefox build with fingerprint patches at the C++ source code level avoids the JS-shim detection surface entirely.

the backend would wrap feder-cr/invisible_playwright (on PyPI as invisible-playwright), which drives a patched Firefox binary built from feder-cr/firefox_antidetect_patch and published there as GitHub Release assets (MPL-2.0, same license as Firefox upstream). for karakeep's Node worker only the binary plus a small prefs map is needed: `firefox.launch({ executablePath, firefoxUserPrefs })`. a Node helper package is on the roadmap if there's interest. issues against the backend route to feder-cr/invisible_playwright.

this PR only adds a proposal doc in docs/docs/08-development/ so the idea has somewhere concrete to land. tracking discussion: #2848

if not in scope i'll close without noise.

### DIFF
--- a/docs/docs/08-development/99-invisible-firefox-proposal.md
+++ b/docs/docs/08-development/99-invisible-firefox-proposal.md
@@ -1,0 +1,29 @@
+# Firefox-based stealth backend (proposal)
+
+> Status: Draft proposal
+> Created: 2026-05-27
+> Tracking discussion: TBD
+
+## Goal
+
+Optional Firefox-based stealth backend for the crawler worker, parallel to the current `chromium` plus `puppeteer-extra-plugin-stealth` setup. Selected via env, no change to defaults.
+
+## Motivation
+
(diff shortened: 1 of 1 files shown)


feder-cr:
@MohamedBassem would a Firefox-stealth backend for the crawler worker be useful for pages that block archiving? proposal doc only, happy to close otherwise.

feder-cr:
@MohamedBassem checking back on this proposal, still opt-in and no change to the default crawler. Happy to close if it's not useful for the project.

feder-cr:
updated the description, invisible_firefox doesn't exist anymore (it was retired). the mechanism doesn't actually depend on it though, it was just the wrong name for where the binary lives: the patched Firefox has always shipped as GitHub Release assets on firefox_antidetect_patch, and invisible_playwright is what drives it. swapped the reference, nothing else about the proposal changes.

MohamedBassem:
Thanks for the PR, but no plans now to switch the browser.

PR #2895, opened 2026-06-17
feat(web): play videos inline on bookmark cards

Santofer:
## What

Bookmark cards can now **play video inline** instead of only linking out:

- **Self-hosted videos** archived by yt-dlp (`videoAssetId`) play via a native HTML5 `<video>` player. The `/api/assets` route already supports range requests, so seeking works.
- **YouTube / Vimeo** links show a poster + play button and load the privacy-friendly embed (`youtube-nocookie` / `player.vimeo`) only on click — so no iframe is mounted per card up front.

The player fills the card's image slot, so it works across all layouts (grid / masonry / list / compact).

## Why

Today, videos only play in the bookmark preview. Since karakeep already downloads videos (yt-dlp) and serves them with range support, surfacing a player directly on the card is a natural, low-cost improvement for anyone who saves a lot of video.

## Notes

- **No schema or backend changes** — `videoAssetId` is already exposed on the link bookmark content, and `/api/assets` already streams with range support.
- Embeds are loaded lazily (only on click) and use the privacy-friendly player domains.
- `lint`, `format` and `typecheck` pass for `@karakeep/web`.

## Screenshots

_Self-hosted `<video>` player card + YouTube poster/play card — added below._

### DIFF
--- a/apps/web/components/dashboard/bookmarks/LinkCard.tsx
+++ b/apps/web/components/dashboard/bookmarks/LinkCard.tsx
@@ -88,27 +88,33 @@ function VideoEmbed({
   let inner: React.ReactNode = null;
   if (videoAssetId) {
     inner = (
+      // eslint-disable-next-line jsx-a11y/media-has-caption -- captions not (yet) available
       <video
         controls
         preload="metadata"
         playsInline
         poster={posterUrl ?? undefined}
         className="absolute inset-0 h-full w-full object-contain"
-        src={`/api/assets/${videoAssetId}`}
       >
-        <track kind="captions" />
+        <source src={`/api/assets/${videoAssetId}`} />
       </video>
     );
   } else if (youTubeId || vimeoId) {
     const embedSrc = youTubeId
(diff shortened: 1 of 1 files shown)


Santofer:
Thanks for the review! Addressed all three points in c6c96c16:

- **iframe sandbox** — added `sandbox="allow-scripts allow-same-origin allow-presentation allow-popups allow-popups-to-escape-sandbox"` to the YouTube/Vimeo embed, so it can't navigate the top frame while autoplay/fullscreen/popouts keep working.
- **Inert `<track>`** — replaced it with the same `{/* eslint-disable jsx-a11y/media-has-caption */}` + `<source>` pattern already used in `LinkContentSection.tsx`, so there's no runtime console warning.
- **Vimeo thumbnail** — the pre-play poster now falls back to the bookmark's crawled image when the provider has no thumbnail (Vimeo, or YouTube without one), instead of a black frame.

`lint`, `format` and `typecheck` pass for `@karakeep/web`.

PR #2899, opened 2026-06-23
feat(search): support list:<id> matcher for stable Smart List references (#2879)

### LINKED ISSUE #2879: Search by list id (list:<id>)

The [Search Query Language](https://docs.karakeep.app/using-karakeep/search-query-language/) offers `list:<name>` for searching, but relying only on list names isn’t ideal since they can change at any time, breaking multiple Smart Lists. 

Provide a `list:<id>` search filter so users can target lists by their immutable identifier instead of by list name. Currently Karakeep supports `list:<name>` in the Search Query Language, but relying on names breaks Smart Lists when list names change; adding `list:<id>` would let searches and Smart Lists remain stable even if a list is renamed.

- Stability for Smart Lists: Smart Lists that use `list:<id>` would not break when a list is renamed, reducing maintenance and user frustration. 
- Measurable reduction in support/issue volume: Fewer broken Smart Lists and fewer user reports about “missing” results after renames.
- Safer automation and integrations: External scripts, widgets, or home‑screen features that rely on list queries will be resilient to name changes, improving reliability for power users and third‑party tooling. 

**No**. You have to maintain a manual mapping from name→id in user scripts or updating Smart Lists after a rename, but these are error‑prone and require manual upkeep. A native `list:<id>` filter is the robust solution.

### Additional context

Example query: `list:kidcwffcnie6rky83n9xof8x` (use the list’s internal id).

studiozeroseven:
## Problem

The search query language only resolves `list:<name>`. Any rename of the underlying list silently breaks every Smart List that referenced it (and every external script / widget / home-screen shortcut that hard-coded the name). That's a real foot-gun for power users with automation layered on top of Karakeep.

## Fix

When the argument to `list:` matches Karakeep's list-id shape (cuid2 — 24 lowercase alphanumeric chars), the parser now emits a new `listId` matcher instead of a `listName` matcher. The id is immutable, so Smart Lists and external automation keep working through renames.

- `packages/shared/searchQueryParser.ts` — detect the id shape and switch matchers.
- `packages/shared/types/search.ts` — new `zListIdMatcher` added to the union.
- `packages/trpc/lib/search.ts` — new `case "listId":` in `getIds`. Looks the list up scoped to the current user (so other users' ids cannot leak), fetches the list's bookmark ids, and applies inverse semantics consistently with `listName`. Unknown ids match nothing — same null-safety pattern the rest of `getIds` uses.
- `apps/web/.../QueryExplainerTooltip.tsx` — new `case "listId":` so the exhaustive `never` check still passes.
- `docs/.../search-query-language.md` — documented `list:<id>` alongside `list:<name>`.

## How to test

1. Create a list, note its id from the URL or via `select id from bookmark_lists where name = '...'` (your data, your call).
2. Search `list:<id>` — the list's bookmarks appear, even after you rename the list.
3. Search `-list:<id>` — everything outside the list appears.
4. Search `list:<unknown-id>` — empty result set, no error.
5. Run the new tests:
   ```
   pnpm --filter @karakeep/shared test
   pnpm --filter @karakeep/trpc exec vitest run lib/__tests__/search.test.ts
   ```

## Tests

- 7 new assertions in the parser tests (id-shape routing, inverse, `!` prefix, uppercase rejection, short-id fallback to name).
- 5 new tests in `trpc/lib/__tests__/search.test.ts` (happy path, inverse, unknown id, missing inverse, cross-user id isolation).
- Without the source change: 1/7 parser assertions and 5/5 trpc tests fail.
- With it: full suite — `shared` 93/93, `trpc` 412/412. `pnpm typecheck` and `pnpm lint` clean across the workspace.

Fixes #2879.

---

Made for studiozeroseven (Ian Schwartz) — theapk llc OSS contribution pipeline.


studiozeroseven:
Closing this PR — not because the fix is wrong, but because I'm pausing my OSS contributions to reassess workflow.

After a maintainer on another project flagged my recent PRs as "vibe coded" and closed them, I realized the shape of my PRs (short fix + many tests, conventional-commits title with issue ref, polished PR body, signed-off-by footer) reads as AI-generated even when the work was done by hand. That look damages the maintainer's time and the project's signal-to-noise, and I'd rather pause than keep shipping that shape.

When I come back to contributions on this repo, the plan is:
- Draft PRs first per this repo's contributing guide (please point me at it if I missed it).
- Lead the PR body with what I personally reproduced (input, output, log) instead of polished Problem/Fix/How/Trade-off prose.
- Trim the test suite to the actual regression case — reuse existing tests where they cover adjacent paths.
- Drop the Signed-off-by footer unless the project requires DCO.

If you already started reviewing this PR and want me to leave it open, let me know in the next few days and I'll reopen. Otherwise this stays closed.

Thanks for the maintainer time you've already spent on this.

Issue #2789, opened 2026-05-12
BUG: INFERENCE_OUTPUT_SCHEMA=json or structured (only one can be chosen). Only one of AI tagging and AI summarization will take effect.


mapleshadow:
### Describe the Bug

The base model is deepseek-v4-flash.  

When INFERENCE_OUTPUT_SCHEMA=json, AI tagging works normally, but AI summarization fails with the following error:  

2026-05-12T07:12:24.186Z info: <-- POST /api/trpc/bookmarks.summarizeBookmark?batch=1
2026-05-12T07:12:25.120Z error: tRPC failed on bookmarks.summarizeBookmark: 400 Prompt must contain the word 'json' in some form to use 'response_format' of type 'json_object'.
2026-05-12T07:12:25.120Z error: Error: 400 Prompt must contain the word 'json' in some form to use 'response_format' of type 'json_object'.
    at i.generate (/app/apps/web/.next/server/chunks/9556.js:9:27518)
    at si.makeStatusError (/app/apps/web/.next/server/chunks/9556.js:16:53421)
    at si.makeRequest (/app/apps/web/.next/server/chunks/9556.js:16:57167)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async F.inferFromText (/app/apps/web/.next/server/chunks/3014.js:1:14991)
    at async /app/apps/web/.next/server/chunks/3014.js:12:510
    at async m.middlewares (/app/apps/web/.next/server/chunks/9556.js:16:98401)
    at async b (/app/apps/web/.next/server/chunks/9556.js:19:68)
    at async b (/app/apps/web/.next/server/chunks/9556.js:19:68)
    at async /app/apps/web/.next/server/chunks/3014.js:25:44829
2026-05-12T07:12:25.122Z info: --> POST /api/trpc/bookmarks.summarizeBookmark?batch=1 500 936ms

When INFERENCE_OUTPUT_SCHEMA=structured, AI summarization works normally, but AI tagging fails with the following error:  

2026-05-12T07:27:01.868Z info: [inference][49040] Starting an inference job for bookmark with id "h1r8vwedsm3qjn4honmh4o1n"
2026-05-12T07:27:02.015Z error: [inference][49040] inference job failed: Error: 400 This response_format type is unavailable now
Error: 400 This response_format type is unavailable now
    at APIError.generate (file:///app/apps/workers/dist/index.js:1675:30)
[...]

PR #2900, opened 2026-06-23
fix(inference): fallback to plain output when call has no schema (#2789)

studiozeroseven:
## Problem

`INFERENCE_OUTPUT_SCHEMA` is a global setting, but the inference calls have different needs:

- **Tagging** always passes a Zod schema and needs a structured response.
- **Summarization** passes `schema: null` (summaries are free-form text) and needs no `response_format`.

Forcing one global schema across both makes at least one of them fail on most models:

| Setting | Tagging | Summarization |
|---|---|---|
| `json` | ✅ works | ❌ 400 "Prompt must contain the word 'json' in some form to use 'response_format' of type 'json_object'" |
| `structured` | ❌ 400 "This response_format type is unavailable now" on models that don't support zodResponseFormat (e.g. deepseek-v4-flash) | ✅ works |
| `plain` | ❌ tagging needs a schema or it gets garbage back | ✅ works |

So no value of the global setting makes both work — the only way to keep both is to make the schema selection per-call, gated on whether the caller actually provided a Zod schema.

## Fix

Introduce `resolveOutputSchema(globalMode, callSchema)` in `packages/shared/inference.ts`:

- If `callSchema === null` → force `plain` (no `response_format`).
- Otherwise → honour the global mode (`structured` → `zodResponseFormat`, `json` → `json_object`, `plain` → undefined).

Applied to:
- `OpenAIInferenceClient.inferFromText`
- `OpenAIInferenceClient.inferFromImage`
- `OllamaInferenceClient.runModel` (mirrors the same logic; `structured` → Zod 4's `z.toJSONSchema`, `json` → "json", `plain` → undefined)

## Tests

8 new vitest cases in `packages/shared/inference.test.ts`. 2 of the 8 (the bug-fix cases for `json`+null and `structured`+null on the text path) **fail on the pre-fix code**, confirming they cover the real regression. Full `packages/shared`, `packages/shared-server`, `packages/trpc`, `packages/db`, `packages/api` suites pass (518/518).

## How to verify

```bash
# Pre-fix repro: with deepseek-v4-flash and INFERENCE_OUTPUT_SCHEMA=json, summarization fails
# Post-fix: both tagging and summarization work
```

### DIFF
--- a/packages/shared/inference.test.ts
+++ b/packages/shared/inference.test.ts
@@ -1,0 +1,138 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { z } from "zod";
+
+import { OpenAIInferenceClient } from "./inference";
+
+const tagSchema = z.object({ tags: z.array(z.string()) });
+
+interface CapturedCall {
+  body: Record<string, unknown>;
+}
+
+const captured: CapturedCall[] = [];
+
+vi.mock("openai", () => {
+  const OpenAIMock = vi.fn().mockImplementation(function (this: unknown) {
+    return {
+      chat: {
+        completions: {
+          create: vi.fn(async (body: Record<string, unknown>) => {
+            captured.push({ body });
+            return {
+              choices: [{ message: { content: "{}" } }],
+              usage: { total_tokens: 1 },
+            };
+          }),
+        },
+      },
+    };
+  });
+  return {
+    default: OpenAIMock,
+  };
+});
+
+vi.mock("openai/helpers/zod", () => ({
+  zodResponseFormat: (schema: unknown, name: string) => ({
+    __zodResponseFormat: true,
+    schema,
+    name,
+    // Mimic the wire shape OpenAI expects, so JSON-stringify round-trips.
+    type: "json_schema",
+    json_schema: { name, schema },
+  }),
+}));
+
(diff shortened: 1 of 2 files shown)


studiozeroseven:
block-scope test comment

studiozeroseven:
Closing this PR — not because the fix is wrong, but because I'm pausing my OSS contributions to reassess workflow.

After a maintainer on another project flagged my recent PRs as "vibe coded" and closed them, I realized the shape of my PRs (short fix + many tests, conventional-commits title with issue ref, polished PR body, signed-off-by footer) reads as AI-generated even when the work was done by hand. That look damages the maintainer's time and the project's signal-to-noise, and I'd rather pause than keep shipping that shape.

When I come back to contributions on this repo, the plan is:
- Draft PRs first per this repo's contributing guide (please point me at it if I missed it).
- Lead the PR body with what I personally reproduced (input, output, log) instead of polished Problem/Fix/How/Trade-off prose.
- Trim the test suite to the actual regression case — reuse existing tests where they cover adjacent paths.
- Drop the Signed-off-by footer unless the project requires DCO.

If you already started reviewing this PR and want me to leave it open, let me know in the next few days and I'll reopen. Otherwise this stays closed.

Thanks for the maintainer time you've already spent on this.

Issue #2914, opened 2026-06-30
[MCP] get-bookmark-content crashes at runtime on link bookmarks with null htmlContent


joestump:
### Describe the bug

`apps/mcp/src/bookmarks.ts:250` calls `turndownService.turndown(htmlContent)` directly, but the SDK schema declares `htmlContent?: string | null`. When a link bookmark's content hasn't been crawled yet (or crawl failed), `htmlContent` is `null`, and `turndown()` throws at runtime (`Cannot read properties of null (reading 'match')` from turndown's input check).

`tsc` doesn't catch this on `main` today because the resolved type narrows in a way that hides it, but the runtime behavior is reproducible.

### To reproduce

1. Create a brand-new link bookmark.
2. Before the crawler runs (or if the crawl fails), call the MCP `get-bookmark-content` tool with that bookmark's id.
3. Tool errors with a generic exception instead of a meaningful MCP error.

### Expected

Empty markdown content (or a meaningful "content not crawled" message), not a thrown exception. Other code paths in the same file already use `?? ""` for nullable fields.

### Fix

One-line null-coerce: `turndownService.turndown(htmlContent ?? "")`. PR coming.

MohamedBassem:
Thanks for fixing it!

PR #2915, opened 2026-06-30
fix(mcp): coerce null htmlContent before turndown in get-bookmark-content

joestump:
## Description

`apps/mcp/src/bookmarks.ts:250` passes `htmlContent` directly to `turndownService.turndown()` without null-checking. The SDK schema declares `htmlContent?: string | null`, so on link bookmarks where the crawl hasn't run (or failed), `turndown()` throws at runtime. One-line coerce to `""` matches the `?? ""` convention used everywhere else in the same file.

Refs #2914

## How Has This Been Tested?

- [x] Local `pnpm --filter @karakeep/mcp typecheck` passes
- [x] Manual reproduction: with the fix, `get-bookmark-content` on an uncrawled link bookmark returns an empty markdown body instead of crashing
- [x] Build still succeeds
- [x] **Regression test landed in the stacked PR #2917** — the vitest harness it needs already exists there (`apps/mcp/src/bookmarks.test.ts`). The regression test asserts that null `htmlContent` is coerced to `""` before being passed to `turndownService.turndown()`, covers a string-htmlContent happy path, and confirms text bookmarks bypass turndown entirely. If you'd prefer the test land here too, happy to bootstrap vitest in this branch as a follow-up commit — flagged separately because it would import the same lockfile cascade we disclose in #2917.

## Checklist:

- [x] I have made corresponding changes to the documentation if applicable (N/A)
- [x] I have confirmed that any new dependencies are strictly necessary. (none)
- [x] I have written tests for new code (if applicable) — regression test in stacked PR #2917, see above

Claude (Anthropic, via Claude Code) authored the fix. I (Joe Stump) reviewed it on my fork before this submission. Closed PRs on my fork that landed this and related work: https://github.com/joestump/karakeep/pulls?q=is%3Apr+is%3Aclosed

### DIFF
--- a/apps/mcp/src/bookmarks.ts
+++ b/apps/mcp/src/bookmarks.ts
@@ -247,7 +247,7 @@ mcpServer.tool(
     let content;
     if (res.data.content.type === "link") {
       const htmlContent = res.data.content.htmlContent;
-      content = turndownService.turndown(htmlContent);
+      content = turndownService.turndown(htmlContent ?? "");
     } else if (res.data.content.type === "text") {
       content = res.data.content.text;
     } else if (res.data.content.type === "asset") {


joestump:
Note: regression test for this fix now lives in the stacked PR #2917 (`apps/mcp/src/bookmarks.test.ts`), since it needs the vitest harness that PR introduces. Body updated to reflect this. Happy to also bootstrap vitest in this branch if you'd prefer the test colocated — would import the same lockfile cascade we disclose in #2917.

MohamedBassem:
Thank you!

PR #2934, opened 2026-07-05
Resolve known link-shortener URLs (search.app / share.google) to their destination

### LINKED ISSUE #2235: resolve https://search.app/ links to the actual URL

In its infinite Wisdom, Google doesn't use the actual domain but a sort of Short Link Version via https://search.app/. I would like a feature that resolves the actual URL and also stores that in the Bookmark Entry if not actually replacing the shortlink itself.

Karakeep does show & archive the correct Websites content but the Bookmark Entry still only contains the Shortlink. For the Moment that works sure but we all know Googles History about supporting their Features. But even besides that I do believe it would be benefitial to have the actual url archived along the Entry

No

the link gets resolved correctly so what I'm requesting is just a way to tell the App to store what it resolved the Link to somewhere in the Entry.

### Additional context

_No response_

bcamarneiro:
## Description

When a bookmark is saved as a Google short/share link (`https://search.app/...` or `https://share.google/...`), the crawler already follows the redirect to fetch and archive the real page — but the stored bookmark URL keeps the opaque short link. This means the entry never reflects the actual destination, and if the shortener ever stops resolving, the real URL is lost.

This PR overwrites the stored URL with the URL the crawler actually resolved to, but **only** when:

- the original host is a **known** link-shortener (`search.app`, `share.google`), and
- the crawler ended up at a **different** URL, and
- that destination is a **safe** `http(s)` URL.

Scoping it to a small allowlist (rather than following every redirect) means ordinary redirects — `http`→`https`, tracking-param strips, trailing-slash canonicalization — never silently rewrite the user's URL. The safe-scheme guard (reusing the existing `isAllowedBookmarkUrl`) ensures a shortener can never rewrite a bookmark to a `javascript:`/`data:`/`file:` URL.

The decision logic is extracted into a pure, unit-tested helper `resolveShortenedBookmarkUrl(originalUrl, crawledUrl)` in `packages/shared/utils/url.ts`. New shorteners can be added by extending the `KNOWN_LINK_SHORTENERS` list.

Resolution is applied consistently across all three outcomes of a crawl:

- **HTML pages** — folded into the existing `bookmarkLinks` update (**no new query, no DB migration**), using the browser's final URL (robust to client-side redirects).
- **Direct PDF/image links** — the content-type probe now returns the redirect's final URL, so the asset-bookmark path stores the resolved destination as `sourceUrl` instead of the short link.
- **Downstream jobs** — `crawlAndParseUrl` returns the effective (post-resolution) URL so the video-download queue receives the real destination, not the opaque short link.

Fixes #2235

## How Has This Been Tested?

- [x] Unit tests for `isKnownLinkShortener` and `resolveShortenedBookmarkUrl` in `packages/shared/utils/url.test.ts`, covering: known-shortener → resolved URL returned; non-shortener → unchanged; no-op redirect → unchanged; dangerous-scheme destination rejected; lookalike hosts (`search.app.evil.com`, `notsearch.app`, `www.search.app`) rejected. (18/18 passing.)
- [x] e2e tests in `packages/e2e_tests` — the nginx fixture container is aliased as `search.app` (allow-listed as a known shortener) with `/shortlink` and `/shortlink-image` redirects, asserting the stored `url` (HTML path) and `sourceUrl` (asset path) are the resolved destinations, not the short link. Run locally against the full container stack: 7/7 crawler tests passing (`http://search.app/shortlink` → stored `.../hello.html`; `http://search.app/shortlink-image` → `sourceUrl .../image.png`).
- [x] `pnpm typecheck`, `pnpm lint`, and `pnpm format` pass across the monorepo.

## Checklist:

- [ ] I have made corresponding changes to the documentation if applicable
- [x] I have confirmed that any new dependencies are strictly necessary. (No new dependencies.)

This PR was developed with the help of an AI coding assistant, following a test-driven approach (tests written and watched to fail before the implementation). All code was reviewed, run, and verified by me before submission.

### DIFF
--- a/apps/workers/workers/crawler/crawlAndParse.ts
+++ b/apps/workers/workers/crawler/crawlAndParse.ts
@@ -34,6 +34,7 @@ import {
 import serverConfig from "@karakeep/shared/config";
 import logger from "@karakeep/shared/logger";
 import { BookmarkTypes } from "@karakeep/shared/types/bookmarks";
+import { resolveShortenedBookmarkUrl } from "@karakeep/shared/utils/url";
 
 import type { ParseSubprocessOutput } from "../utils/parseHtmlSubprocessIpc";
 import {
@@ -50,7 +51,11 @@ import {
 } from "./assetStorage";
 import { crawlPage } from "./crawlPage";
 import { runParseSubprocess } from "./parseSubprocess";
-import { redactUrlCredentials, shouldRetryCrawlStatusCode } from "./utils";
+import {
+  redactUrlCredentials,
+  shouldRetryCrawlStatusCode,
+  truncateUrl,
+} from "./utils";
 
 const tracer = getTracer("@karakeep/workers");
 
@@ -170,7 +175,7 @@ export interface CrawlAndParseUrlArgs {
  */
 export async function crawlAndParseUrl(
   args: CrawlAndParseUrlArgs,
-): Promise<() => Promise<void>> {
+): Promise<{ runArchival: () => Promise<void>; effectiveUrl: string }> {
   const {
     url,
     userId,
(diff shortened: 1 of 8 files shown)


bcamarneiro:
Good catch on the stale local `url` in `runCrawler` — I've addressed it in b6d08fa.

One clarification on the stated impact, though: the runtime effect on video downloads is nil today. `videoWorker` doesn't read `url` from the queue payload — it re-fetches it from the DB via `getBookmarkDetails(bookmarkId)` (`videoWorker.ts:108-112`), and by the time the job is enqueued the resolved URL has already been written to `bookmarkLinks.url` (the Phase-1 `await db.update(...)` completes before `crawlAndParseUrl` returns). It then resolves redirects itself via `resolveValidatedRedirectUrl`. So the payload `url` was effectively dead data.

That said, passing a stale value is a smell worth removing, so `crawlAndParseUrl` now returns the effective (post-resolution) URL and the video enqueue uses it — the payload is accurate and won't mislead if something starts trusting it later.

pambroziak-rtbh:
@MohamedBassem any chance to get this merged? This feature is a game changer for using Karakeep with  the Google Discovery feed

bcamarneiro:
Rebased onto current `main` — the branch is conflict-free again. Conflicts were mechanical: an import in `crawlAndParse.ts` and the `CRAWLER_ALLOWED_INTERNAL_HOSTNAMES` line in the e2e `docker-compose.yml`, which now carries both `search.app` and the upstream `CRAWLER_STORE_PDF` addition. No behaviour change from the previous review round.

@MohamedBassem this is ready whenever you have a moment.

PR #2935, opened 2026-07-06
Extract Instagram post/reel text into searchable content (opt-in)

bcamarneiro:
## Description

When a user bookmarks an Instagram URL (post / reel / tv), the normal browser crawl hits Instagram's login wall and returns an empty shell, so `bookmarkLinks.htmlContent` ends up empty. The caption and any spoken content are never stored — so the bookmark is **not searchable, not summarizable, and disappears if the post is later deleted or made private**. The existing `InstagramRenderer` only shows the official embed iframe, which is display-only.

This PR adds an optional crawler path that extracts the **text** of an Instagram bookmark (caption + auto-generated transcript for reels) into `htmlContent`, so the existing search / summarization / tagging / embedding pipeline picks it up. It reuses infrastructure already in the repo — **yt-dlp** (already wired in `videoWorker`) for fetching, and the existing `CRAWLER_YTDLP_ARGS` for cookies — so there are **no new dependencies and no DB migration**.

**Off by default.** The feature is gated behind a new `CRAWLER_INSTAGRAM_ENABLED` flag (default `false`), because it needs cookies to be useful and must not change behavior for users who don't configure it.

### How it works

In `runCrawler`, before the normal content-type probe, when `CRAWLER_INSTAGRAM_ENABLED` is on **and** the URL is an Instagram post/reel/tv link, the crawl is routed to a dedicated handler that:

1. Runs `yt-dlp --skip-download --write-info-json --write-auto-subs` (plus the configured `CRAWLER_YTDLP_ARGS` and the run proxy) into a temp dir.
2. Parses the caption/author/date from the `.info.json` and the transcript from the `.vtt` (timestamps stripped, consecutive duplicate cues collapsed).
3. Composes a small HTML document (all user text HTML-escaped) and writes `htmlContent` / `title` / `description` / `author` to `bookmarkLinks` — the same fields the normal crawl writes.
4. Enqueues the same downstream jobs (summarize / tag / embed / search reindex) via the existing `enqueuePostCrawlJobs`.

The whole Instagram concern lives in one new module, `apps/workers/workers/crawler/instagram.ts`, mirroring the structure of the existing `handleAsAssetBookmark`.

### Security

yt-dlp is invoked via `execa` (no shell). The bookmark URL is passed **after a `--` end-of-options separator** and non-`http(s)` URLs are rejected before yt-dlp runs, so a crafted URL can't smuggle yt-dlp flags. The `isInstagramUrl` gate additionally restricts the host to `instagram.com`. SSRF exposure is the same yt-dlp-fetches-a-URL property the existing `videoWorker` already has.

### Notes / limitations

- On extraction failure (rate limit, expired/absent cookies, private post) the handler logs a warning and leaves the embed untouched — it does **not** hard-fail the bookmark, and does not fall through to the browser crawl (which would only hit the login wall).
- Reel **media download** still requires the separate `CRAWLER_VIDEO_DOWNLOAD` flag; this PR only adds text extraction.
- Static image posts (text baked into the image) are out of scope for now — no OCR. Stories are out of scope (they expire / require following).

## How Has This Been Tested?

- [x] **Unit tests** (`apps/workers/workers/crawler/instagram.test.ts`, 8/8 passing) covering the pure logic: `isInstagramUrl` (accepts `/p/`, `/reel/`, `/reels/`, `/tv/`; rejects profile URLs, non-Instagram hosts, and lookalike hosts); VTT parsing (timestamps stripped, duplicate cues collapsed); HTML composition (caption-only, caption+transcript, HTML-escaping of user content); and dump-directory parsing from a temp fixture.
- [x] `pnpm --filter @karakeep/workers typecheck` and `pnpm --filter @karakeep/shared typecheck` pass; the module follows existing crawler patterns.
- [ ] **Live end-to-end fetch against Instagram is not yet verified.** It requires a logged-in account's cookies (via `CRAWLER_YTDLP_ARGS`) and cannot run in CI. The yt-dlp subprocess boundary is therefore covered only by the unit-tested parsing layer, not by a real fetch. I'd appreciate guidance on whether a maintainer can validate this against a real account, or whether a recorded fixture would be preferred.

## Checklist:

- [x] I have made corresponding changes to the documentation (documented `CRAWLER_INSTAGRAM_ENABLED` in the configuration docs).
- [x] I have confirmed that any new dependencies are strictly necessary. (No new dependencies — reuses the existing yt-dlp integration.)
- [x] I have written tests for new code (the pure/parsing logic; the live yt-dlp fetch is not unit-testable — see above).

This PR was developed with the help of an AI coding assistant, following a test-driven approach (tests written and watched to fail before the implementation). All code was reviewed, run, and verified by me before submission. The live Instagram fetch has not been manually verified yet, as noted above.


bcamarneiro:
Rebased onto current `main` (the conflict was only the env-var docs table) and addressed all four review comments in 5b78035:

**P1 — VTT header metadata leaking into the transcript.** Confirmed. `parseVtt` now tracks whether it is inside a cue rather than enumerating lines to skip: a blank line closes the current block, a `-->` line opens a cue, and only the lines after that timestamp are treated as payload. The header block (`Kind:`, `Language:`, `X-TIMESTAMP-MAP=...`) never follows a timestamp, so it is dropped without special-casing each key.

**P2 — non-numeric cue identifiers.** Fixed by the same change: an identifier sits before the timestamp line, so it falls outside a cue regardless of whether it is numeric. `NOTE`/`STYLE`/`REGION` blocks are dropped for the same reason.

**P1 — title/description nullified on transcript-only reels.** Confirmed. `title`/`description` are now only included in the `.set()` object when there is a value, falling back to the transcript when the caption is empty. `author` had the same problem (it is `null` when yt-dlp reports neither `uploader` nor `channel`), so it is guarded the same way.

**P2 — downstream jobs enqueued after a failed extraction.** `handleInstagramBookmark` now returns whether anything was written, and `runCrawler` only calls `enqueuePostCrawlJobs` when it was.

Added two `parseVtt` regression tests (header metadata, non-numeric identifiers plus a `NOTE` block). `pnpm preflight` and the workers test suite pass.

bcamarneiro:
Rebased onto current `main` and pushed three more commits. Two of them change what this PR does, so a summary:

**The transcript half of this PR never worked.** While debugging my own instance I found that Instagram serves no subtitle track to yt-dlp — `--list-subs` reports `has no subtitles` and the `info.json` carries `subtitles: null` / `automatic_captions: null` for every post type. So `--write-auto-subs --sub-langs 'en.*'` never produced a `.vtt`, and `transcript` was empty for every reel (36 reels in my database, zero transcripts). Sorry for not catching that before opening the PR.

**What changed**

1. **Posts are now read from Instagram's page, not from yt-dlp** (`instagramPage.ts`). Instagram embeds the post as JSON in the page it serves to browser-like requests (the `Sec-Fetch-*` / `sec-ch-ua` headers are what flip it) — no cookies or session for a public post. The JSON has everything the yt-dlp metadata pass produced (caption, author, date) plus what it never could: every item of a carousel, typed image/video, with a direct CDN URL and Instagram's alt text. yt-dlp is kept as a fallback for non-public posts, and the cookie requirement in the docs is now described as what it is — the fallback's requirement, not the feature's.
2. **Carousel images become searchable text.** Each image yields its alt text for free (Instagram's `May be an image of text that says '…'`), and with `CRAWLER_INSTAGRAM_DESCRIBE_IMAGES=true` it is also run through the existing LLM OCR prompt (`buildOCRPrompt`). Capped per post by `CRAWLER_INSTAGRAM_MAX_IMAGES`.
3. **Videos are actually transcribed** (`CRAWLER_INSTAGRAM_TRANSCRIBE=true`): fetch the video, ffmpeg takes the audio, and a new `InferenceClient.transcribeAudio` sends it to `INFERENCE_AUDIO_MODEL` (default `whisper-1`). The Ollama client returns `null` there, so transcription degrades to "skipped" rather than failing the crawl.
4. yt-dlp is handed a **private copy of the cookie jar** on every run. It rewrites the jar it is given on exit, and Instagram's answer to a burst of requests can be a `Set-Cookie` that drops the session — after which the mounted jar was logged out for good (that is how I lost mine).

Everything new is opt-in and off by default, since transcription and OCR are paid inference calls. The env-var table and the setup guide are updated accordingly.

Happy to split this into smaller PRs if that is easier to review — e.g. the page-based extraction (which removes the cookie dependency) first, and transcription / OCR as follow-ups.

bcamarneiro:
Related: #2984 asks for gallery-dl to handle "Instagram multi-image posts". This PR now reads those posts (every carousel slide, typed image/video, with Instagram's alt text and optional OCR) straight from the page, without a new downloader — but only as text. It does not store the images as assets, which is the other half of that issue.

bcamarneiro:
Deployed this branch to production (homelab karakeep instance) and validated against real Instagram posts. Seven fixes plus one follow-up fix found by that deploy:

- **Video-only streams silently lost the transcript.** Instagram serves some reels with the audio on a separate DASH track that ffmpeg's direct pass can't see. Reel `DaTZsvDKgOd` and carousel `DbLXBy4jp1u` both had video-only streams and lost 13 transcripts between them with no error surfaced anywhere. Fixed by retrying via `yt-dlp`'s `bestaudio` selector, which reaches the separate track anonymously, when the direct pass transcribes nothing.
- **Transient Instagram failures were stored as success.** A page fetch that failed with a retryable error (rate limit, transient 5xx, etc.) still ended the job as `crawlStatus: success` with nothing extracted — indistinguishable from "there was nothing to extract." Now retried with backoff, and if every attempt fails, the job is correctly marked `crawlStatus: failure`.
- **The 60s crawl timeout dropped work already done.** A caption/transcript/images that had already been extracted and were sitting in memory got thrown away if the job hit the timeout before the DB write. `CRAWLER_JOB_TIMEOUT_SEC` is now respected and partial content already extracted is stored instead of discarded; on abort, the audio fallback is skipped but alt text for unprocessed images is kept.
- **No visibility into what actually got extracted.** Added a `[ig] path=<page|yt-dlp> images=<got>/<expected> videos=<got>/<expected> status=<ok|partial|failed>` log line per crawl, plus an HTML comment marker (`<!-- karakeep-ig ... -->`) at the end of stored content, so extraction quality is auditable after the fact instead of only inferred from the text.
- **New opt-in env vars**, both off/default unless set: `CRAWLER_INSTAGRAM_OCR_DETAIL` (image OCR detail level for carousels) and `CRAWLER_INSTAGRAM_HEADERS_JSON` (override the Instagram page request headers). Nothing here changes default behavior for anyone not setting them.

**Regression found by the deploy e2e, then fixed:** the streamed-video-download change (buffering to disk instead of memory, to avoid OOM on large videos) called `Readable.fromWeb(response.body)`, but `fetchWithProxy` is backed by node-fetch v3, whose `Response.body` is already a Node `Readable`, not a Web `ReadableStream` — so every direct video transcription threw and silently fell through to the (weaker) yt-dlp fallback. A first deploy attempt caught this via the same production e2e used above and was rolled back before publishing. Fixed by duck-typing the body on `.pipe` and only calling `Readable.fromWeb` for genuine Web streams; a new unit test streams a real node-fetch-shaped body to disk and asserts ffmpeg is invoked and a transcript comes back — this test fails against the buggy commit and passes with the fix.

All changes are opt-in / backward compatible: existing behavior for accounts not setting the new env vars is unchanged, and the retry/partial-content/marker logic only adds visibility and resilience around the existing extraction path.

bcamarneiro:
A whole-branch review of this PR turned up four more issues, now fixed and re-reviewed in three commits:

- **Stats counted padded alt text as extracted.** `images.got` was incrementing for every image that got *any* string back, including images where OCR was skipped and only the (possibly empty) alt text placeholder was kept. `stats.got` now only counts an image as extracted when it was actually processed as configured (OCR ran, or alt text was genuinely present) — so `[ig] status=` and the htmlContent marker stop over-reporting success.
- **The video byte cap wasn't enforced mid-stream.** The size check ran against the `Content-Length` header and again after the full file hit disk, but a video whose declared length lied (or was absent) could still stream an unbounded amount to disk before the post-write check caught it. A byte-counting `Transform` now enforces the cap as bytes arrive, so a large stream is aborted mid-flight instead of after it's already been written.
- **The yt-dlp audio fallback fired on any shortfall, not just no-audio-stream posts.** Previously any failure to produce a transcript — network hiccup, ffmpeg error, whatever — triggered the yt-dlp retry, which is the correct behavior only for the "video track has no audio, it's in a separate DASH stream" case. The fallback is now scoped to that specific signature so it doesn't mask other failure modes as "needs yt-dlp."
- **`CRAWLER_INSTAGRAM_HEADERS_JSON` could echo its own value on a validation error.** A malformed value used to end up in the thrown error message, which risked leaking whatever was in that env var (cookies, auth headers) into logs. Validation now reports a bad value via `ctx.addIssue` without ever echoing the value itself.

## Known behaviour

A few things are working as designed, not bugs, worth calling out for anyone reading crawl logs on this feature:

- Retries of transient Instagram failures are paced by the job queue (roughly 1s apart, 5 attempts). That's enough to give visibility — a persistently-failing post correctly ends up `crawlStatus=failure` instead of silently `success` with nothing extracted — but it is not recovery. Recovering from a transient outage is the operator's periodic recrawl, not the retry loop itself.
- Anonymous `yt-dlp bestaudio` may legitimately find no audio for a given reel. When that happens the post is stored with an honest `status=partial` rather than pretending the video was transcribed.
- Non-public (private/deleted/login-walled) posts complete the crawl with no new content extracted; this is logged at `warn` with yt-dlp's stderr attached, so the reason is visible without failing the job.
- `[ig] path=ytdlp status=ok` marks that the degraded metadata-only fallback path was used, not that the extraction is healthy — `path=page` is the primary path and the one to expect for a working post.

PR #2944, opened 2026-07-11
fix(crawler): anonymous Reddit fallback via JS-challenge solve (#2885)

### LINKED ISSUE #2885: [Crawler] Reddit crawling is now getting blocked
As part of a recent reddit crackdown on crawlers ([reference](https://www.reddit.com/r/modnews/comments/1tq9vxo/protecting_communities_from_scrapers_and_platform/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button)), the endpoint that we were using for getting post metadata is now completely blocked. All reddit crawlings will show a banner like this:

Image

We'll need to find a way to either restore the functionality or a mechanism for users to provide some auth tokens so that the requests to reddit are authenticated.

beaglemoo:
Fixes #2885.

Reddit now blocks unauthenticated `.json` metadata requests, so Reddit bookmarks fail to crawl. #2887 adds an OAuth client-credentials path, but Reddit closed self-serve API app creation under its Responsible Builder Policy, so many self-hosters can't obtain credentials.

This keeps the OAuth path (used when `REDDIT_CLIENT_ID`/`REDDIT_CLIENT_SECRET` are set) and adds an anonymous fallback: with a browser User-Agent, Reddit serves a lightweight JS challenge (`solution = seed` repeated twice) instead of a hard block. The metascraper warms up on the HTML post URL, solves the challenge, submits the hidden form to earn clearance cookies, and reuses them for the `.json` request. Cookies are threaded manually since `fetchWithProxy` (node-fetch) has no cookie jar. If the challenge format changes, it falls back to the existing DOM scraping.

Verified end-to-end on a self-hosted instance: Reddit bookmarks crawl with title, author, subreddit, preview image, and content, with no credentials configured.

Note: this deliberately solves Reddit's anti-bot challenge, flagging that explicitly in case it raises ToS concerns for the project.

### DIFF
--- a/apps/workers/metascraper-plugins/metascraper-reddit.ts
+++ b/apps/workers/metascraper-plugins/metascraper-reddit.ts
@@ -248,11 +248,13 @@ const solveRedditChallenge = async (
     token: token[1],
     jsc_orig_r: "",
   });
-  const submitUrl =
-    "https://www.reddit.com" + action[1] + "?" + params.toString();
+  // action may be a relative path, absolute URL, or protocol-relative URL;
+  // resolve against reddit.com so we never build a malformed submit URL.
+  const submitUrl = new URL(action[1], "https://www.reddit.com");
+  submitUrl.search = params.toString();
   try {
     await browserGet(
-      submitUrl,
+      submitUrl.toString(),
       "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
       jar,
     );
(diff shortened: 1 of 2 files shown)


beaglemoo:
Thanks for the review. Addressed 2 of the 3 in the latest commit:

- **Absolute/protocol-relative form action** — fixed with `new URL(action, "https://www.reddit.com")`.
- **Rejected OAuth token skips fallback** — on a 401/403 from the OAuth path, the cached token is now cleared and the code falls through to the anonymous challenge path.

On **redirect cookies dropped**: this is a real limitation but I've left it as-is deliberately. Capturing per-hop `Set-Cookie` would mean reimplementing `fetchWithProxy`'s redirect loop (which also does per-hop SSRF revalidation and proxy-agent selection), and I'd rather not fork that logic here. In practice Reddit serves the challenge page and the challenge-submit response as terminal `200`s and sets the clearance cookies there; the only redirect observed on the warmup is same-site trailing-slash canonicalization, which sets no auth cookie. If the challenge submit ever fails, the plugin falls back to the existing DOM scraping, so the failure mode is graceful. Happy to revisit if a maintainer would prefer a redirect-aware cookie jar.

beaglemoo:
Good catch on the OAuth fallback - fixed in the latest commit. The OAuth path now falls back to the anonymous challenge on any non-OK response (including 429 rate limits and 5xx), not just 401/403. The cached token is only dropped on 401/403, since a rate limit or transient server error doesn't mean the token itself is invalid.

ncareau:
Hi @beaglemoo, looking forward to getting this merged. In the meantime, do you have a docker registry with this version on the web ? Would love to try it.

beaglemoo:
Hey @ncareau — sure thing. I pushed an unofficial build of this PR to GHCR so you can try it before it merges:

```
ghcr.io/beaglemoo/karakeep:reddit-pr2944
```

It's the standard all-in-one image, just built from this PR branch. Swap your existing karakeep `aio` image tag for that one in your compose and keep all your env/volumes the same.

A couple of caveats:
- **Unofficial, not for production** — it just tracks this PR and I won't be keeping it updated. Once this merges, switch back to an official release.
- **amd64 only** for now. If you're on arm64 (Pi, etc.) let me know and I'll build that too.

Note that AI tagging still needs your usual inference config; the fix itself just restores Reddit title/body/comments. Let me know how it goes!

PR #2947, opened 2026-07-12
feat: add AI agent backend

MohamedBassem:


packages/trpc/routers/chat/tools.ts
```diff
@@ -0,0 +1,1202 @@
...
+            Type.String({
+              description:
+                "The smart-list query. It must use search qualifiers and no unqualified full-text terms.",
+            }),
+          ),
+          public: Type.Optional(
+            Type.Boolean({
+              description: "Whether the list is publicly shared.",
+            }),
+          ),
```
rivassec, line 749: The `public` param is the one I would take out of the agent's reach. The agent ingests untrusted page content (`get-bookmark-content`, crawled titles/notes), so a saved page that says "make my lists public" can get the model to call `update-list {public: true}` and turn a private list into a public share URL. The other fields are low-stakes; this one is data exfiltration, gated only by prompt wording.

Removing it here stops the agent changing list visibility (users still can in the UI). You would also drop "or public state" from the tool description just above. The `delete-*` / `merge-tags` tools are a separate case that probably wants an explicit client-side confirmation rather than removal.

```suggestion
```

PR #2948, opened 2026-07-13
fix: respect PORT in Docker healthchecks

### LINKED ISSUE #2940: docker-compose healthcheck fails when PORT environment variable has been customized
### Describe the Bug

Using the provided docker-compose, healthcheck for web service constantly reports "unhealthy" for me, but Karakeep is working fine. The healthcheck defined in Dockerfile is looking at "http://127.0.0.1:3000/api/health" - i.e., hardcoded to the default listening port.

My install runs on port 80, defined through the PORT variable in the .env file. Assuming this is a supported configuration, healthcheck should account for it.

The documentation does warn: "DON'T CHANGE THIS IF YOU'RE USING DOCKER, instead changed the docker bound external port." But I'm running karakeep on a custom macvlan network where port mapping doesn't apply.

Healthcheck works correctly (reports "healthy") when I override in docker-compose with one minor edit:
```
services:
  web:
    healthcheck:
      test: wget --no-verbose --tries=1 --spider http://127.0.0.1:${PORT}/api/health || exit 1
      interval: 30s
      timeout: 10s
      retries: 3
      start_period: 5s

```

### Steps to Reproduce

See above

### Expected Behaviour

Healthchecks should report "healthy" when containers are running properly.

_No response_

### Device Details

_No response_

### Exact Karakeep Version

v0.32.0

### Environment Details

Unraid

### Debug Logs

_No response_

zerone0x:
## Summary
Fixes #2940

Update the Docker image healthchecks to use the configured `PORT` environment variable, while retaining `3000` as the default when `PORT` is unset.

## Changes
- Use `${PORT:-3000}` in the AIO image healthcheck URL
- Use `${PORT:-3000}` in the legacy web image healthcheck URL

## Test Plan
- Verified both Dockerfile `HEALTHCHECK` lines now reference `http://127.0.0.1:${PORT:-3000}/api/health`
- Smoke-tested shell expansion: unset `PORT` resolves to `:3000`; `PORT=80` resolves to `:80`
- Docker image build was not run locally because Docker/Podman is unavailable in this cron environment

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -178,7 +178,7 @@ RUN touch /etc/s6-overlay/s6-rc.d/user/contents.d/init-db-migration \
     /etc/s6-overlay/s6-rc.d/user/contents.d/svc-web \
     /etc/s6-overlay/s6-rc.d/user/contents.d/svc-workers
 
-HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 CMD wget --no-verbose --tries=1 --spider http://127.0.0.1:3000/api/health || exit 1
+HEALTHCHECK --interval=30s --timeout=10s --start-period=5s --retries=3 CMD wget --no-verbose --tries=1 --spider "http://127.0.0.1:${PORT:-3000}/api/health" || exit 1
 
 ################# The web container ##############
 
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thank you!

PR #2966, opened 2026-07-28
fix(web): default the preview pane to the Video section when a video asset exists

yazzbert:
When `CRAWLER_VIDEO_DOWNLOAD` is enabled and a link bookmark has a downloaded video (`videoAssetId` set), the preview pane still defaults to the cached reader view — which for JS-heavy sources (Pinterest, Vimeo, Behance) is often
signed-out boilerplate. The downloaded video sits behind the section dropdown where it is easy to miss.

This changes the default section order to:

1. custom content renderer (YouTube/TikTok embeds — unchanged, still first)
2. **video, when the bookmark has a `videoAssetId`** (new)
3. cached reader view (unchanged fallback)

No behavior change for bookmarks without a video asset.


MohamedBassem:
Thanks for the PR, but I'm not sure that this is the most intuitive approach specially when ytp-dl sometimes downloads irrelevant videos.

PR #2988, opened 2026-08-06
docs: add a Railway installation page

hmseeb:
Adds `Installation → Railway`, a sibling to the existing PikaPods page.

Same audience as PikaPods: people who want Karakeep running without administering a server.
One click brings up Karakeep, Meilisearch and headless Chrome wired together, generates
`NEXTAUTH_SECRET` and the Meilisearch master key, and serves the app over HTTPS on an assigned
domain. The page follows the PikaPods structure — requirements, numbered steps, optional
settings — and points at the same `DISABLE_SIGNUPS` and `OPENAI_API_KEY` knobs the rest of the
docs use.

Being upfront about incentives, since it matters here: unlike PikaPods there is **no revenue
share** on this, and the page is flagged as community-maintained rather than an official build.
The info banner at the top links straight to Karakeep Cloud for anyone who'd rather you handled
hosting, and I've deliberately left the README alone so Cloud stays the only hosted option
surfaced there. If you'd rather this page didn't exist at all given Cloud, say so and I'll
close it — no hard feelings.

### How I verified

Deployed the template twice from scratch today. All three services reach a healthy running
state on first deploy with no retries and no manual steps, and the app serves over HTTPS on
the assigned domain.

Docs only, no source touched.

### DIFF
--- a/docs/docs/02-installation/11-railway.md
+++ b/docs/docs/02-installation/11-railway.md
@@ -20,6 +20,10 @@ This provisions three services wired together: Karakeep itself, Meilisearch for 
 search, and a headless Chrome instance for crawling. `NEXTAUTH_SECRET` and the Meilisearch
 master key are generated for you, and `NEXTAUTH_URL` is set to the assigned domain.
 
+A persistent volume is attached to the Karakeep service at `/data`, and a second one to
+Meilisearch at `/meili_data`. See [Data persistence](#data-persistence) below for why this
+matters.
+
 ### 2. Wait for the first deploy
 
 All three services start in a couple of minutes. Karakeep is then served over HTTPS on an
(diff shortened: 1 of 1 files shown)


hmseeb:
Good catch, and the substance is right — fixed in 8124bbc6.

To be clear about what was and wasn't broken, since the two have different consequences:

**The docs were wrong.** I described `/data` as optional storage for crawled assets. Your own
[configuration docs](https://docs.karakeep.app/configuration) say plainly that `DATA_DIR` is
"where the db lives", so that line was both inaccurate and, as flagged, dangerous advice.

**The template was already correct.** It attaches a persistent volume to the Karakeep service
at `/data` and another to Meilisearch at `/meili_data`. So nobody who used the button was ever
at risk — I just documented it as an optional afterthought instead of stating it was already
handled.

What changed:

- Step 1 now says up front that both volumes are attached.
- New **Data persistence** section explaining that `/data` holds the SQLite database as well as
  assets, with a warning against changing `DATA_DIR` or detaching the volume.
- Added a note that Railway doesn't back volumes up automatically, so anyone who cares about
  their library should set up scheduled backups and keep a copy off Railway. That felt like the
  more useful safety point once the volume itself stopped being the risk.
- Removed the misleading bullet from **Optional**, which now only covers `OPENAI_API_KEY`.

MohamedBassem:
The more options that we give to the users the better, I'm ok with having a railway template for karakeep in general (and I don't mind it competing with karakeep cloud or pikapods), but given that this is community maintained, I'd rather be the one maintaining it instead to keep tighter control on it. Thanks for the PR, though!

Issue #72, opened 2024-04-02
Why chrome error?


jiangaq:
When  doing `docker compose up -d`,it returns: 

```
chrome Error                                                                                                                                         
Error response from daemon: Head "https://gcr.io/v2/zenika-hub/alpine-chrome/manifests/123":
```

MohamedBassem:
Oh, that's a bug. Can you in the docker compose file replace `gcr.io/zenika-hub/alpine-chrome:123` with `zenika-hub/alpine-chrome:123`. Let me know if this fixes it. I'll send a quick fix now.

MohamedBassem:
Actually, the GCR link should work. But you can try the suggestion that I mentioned if it didn't.

jiangaq:
> Oh, that's a bug. Can you in the docker compose file replace `gcr.io/zenika-hub/alpine-chrome:123` with `zenika/alpine-chrome:123`. Let me know if this fixes it. I'll send a quick fix now.

it returns:
`Error response from daemon: pull access denied for zenika-hub/alpine-chrome, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

jiangaq:
It works now.

By the way, is there possible to use proxy openai base url instead?

jiangaq:
I found some pages "Failed to fetch link content ..." Why it got error? Are they waiting to fetch in background?

MohamedBassem:
Checking the logs of the `workers` container should tell you why they failed. You can retry them by clicking the `refresh` button in the bookmark option menu.

MohamedBassem:
Please reopen if you still need help :)

PR #3008, opened 2026-08-11
Fix broken gcr.io chrome image in docker/docker-compose.yml

robklingler:
## Problem

`docker/docker-compose.yml` still references `gcr.io/zenika-hub/alpine-chrome:124` for the `chrome` service, which currently fails to pull:

```
Error response from daemon: Head "https://gcr.io/v2/zenika-hub/alpine-chrome/manifests/124":
denied: This API method requires billing to be enabled. Please enable billing on project #...
```

This is the same failure mode reported in #72 back in 2024, where @MohamedBassem confirmed switching away from the `gcr.io/zenika-hub/` mirror as the fix.

## What this PR does

`docker/docker-compose.yml` is the *only* deployment file left in the repo still pointing at the old `gcr.io/zenika-hub/alpine-chrome` image. Every other compose/deployment file already uses the first-party image:

- `docker/docker-compose.dev.yml`
- `docker/docker-compose.build.yml`
- `kubernetes/chrome-deployment.yaml`
- `packages/benchmarks/docker-compose.yml`
- `tools/seed-snapshot/docker-compose.yml`
- `start-dev.sh`

...all use `ghcr.io/karakeep-app/karakeep-chrome:release`. This PR just brings the main production compose template in line with the rest of the repo, so users following the primary install docs don't hit the GCR billing wall.

## Testing

Applied this exact change to my own self-hosted instance and confirmed `docker compose up -d` pulls and starts `chrome` cleanly with no image errors.

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -22,10 +22,12 @@ services:
   chrome:
     image: ghcr.io/karakeep-app/karakeep-chrome:release
     restart: unless-stopped
-    init: true
     command:
+      - --no-sandbox
       - --disable-gpu
       - --disable-dev-shm-usage
+      - --remote-debugging-address=0.0.0.0
+      - --remote-debugging-port=9222
       - --hide-scrollbars
       - --disable-blink-features=AutomationControlled
       - --window-size=1440,900


robklingler:
Closing this — looks like #3005 already fixed this exact issue (migrated `docker/docker-compose.yml` to `ghcr.io/karakeep-app/karakeep-chrome:release` with the correct `init: true` + flags) while this PR was in flight. My branch was cut before that landed, and the diff here would actually regress `main` back to a stale flag set rather than add anything. Thanks for the quick fix!

PR #3010, opened 2026-08-11
fix(mobile): server save button clipped by keyboard

### LINKED ISSUE #2999: [iOS] Save button on server setup screen (address + custom headers) is hidden behind the form and snaps back when scrolled into view
### Describe the Bug

On the server setup screen (server address and custom headers in one form), the Save button is positioned behind/below the visible form area and is not reachable in the default layout. Swiping up briefly reveals the Save button, but it immediately snaps back to its hidden position, making it untappable. This happens regardless of whether a custom header is added — occurs with just the server address field filled in as well.

### Steps to Reproduce

1.	Open Karakeep iOS app, start server setup / add server
	2.	On the combined server address + custom headers screen, the Save button is not visible/reachable
	3.	Swipe up on the screen
	4.	Save button appears briefly, then snaps back to a hidden position
	5.	Reproducible with or without adding a custom header — the header content is not the trigger

### Expected Behaviour

The Save button should always be visible and reachable on this screen (e.g. via a fixed/sticky position or a scrollable form that properly accounts for the button’s space), regardless of form content.

https://github.com/user-attachments/assets/3bae72a9-e0de-4f2b-bb3e-e2aae633e888

### Device Details

iPhone SE, iOS 26.6

### Exact Karakeep Version

1.10.0

### Environment Details

_No response_

### Debug Logs

_No response_

InventorB:
## Description

- Fixes the "save" button in the server configuration being clipped by the keyboard on iOS 26, especially on smaller devices like the Iphone SE. This is done by modifying the padding attributes in the KeyboardAwareScrollView.
- While I was here I also noticed that the text for the custom headers wasn't centered, so I fixed that also.

A user might encounter this bug while trying to setup a server, and have trouble hiding the keyboard to show this button. 

Fixes #2999 

## How Has This Been Tested?

This bug was only reproducible on iOS 26 (18 is excluded). I used the simulator to reproduce and test fixes.

Before  

https://github.com/user-attachments/assets/32fad451-0b4b-4641-963d-8a695d09ea33

After  

https://github.com/user-attachments/assets/4bad84b5-00a9-493f-9925-ad1dee17f3fb

</details>

## Checklist:

Copilot was used to suggest the changes, which I reviewed. After that I was able to reduce the diff, and then validated it in the simulator.

### DIFF
--- a/apps/mobile/app/server-address.tsx
+++ b/apps/mobile/app/server-address.tsx
@@ -89,9 +89,16 @@ export default function ServerAddress() {
 
   return (
     <KeyboardAwareScrollView
-      contentContainerClassName="items-center gap-4 px-4 py-4"
+      contentContainerStyle={{
+        alignItems: "center",
+        gap: 16,
+        paddingHorizontal: 16,
+        paddingTop: 16,
+        paddingBottom: 96,
+      }}
       bottomOffset={20}
       keyboardShouldPersistTaps="handled"
+      keyboardDismissMode="interactive"
       contentInsetAdjustmentBehavior="automatic"
     >
       {/* Error Message */}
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thanks for the PR!

PR #3019, opened 2026-08-15
fix(trpc): show collaborator bookmarks in public lists

### LINKED ISSUE #2594: Links from collaborators are not visible in public lists

The links added to public lists by collaborators are not visible when the list viewed via public link. Only the links added by the owner are really public.
Karakeep actually shows the bookmark count correctly in public links but only the owner's bookmarks are visible.

Image

One expects to see the all the bookmarks in a public list.

I don't think so.

### Additional context

_No response_

NoiceHax:
## Description

When a shared list is made public, the public page only showed bookmarks belonging to the list owner. Anything a collaborator had added was missing, even though the item count at the top of the page was correct.

The cause is in `List.getPublicListContents`. It collected every bookmark id in the list and handed them to `Bookmark.loadMulti` as `ids`. That query path filters on `bookmarks.userId`, so bookmarks owned by a collaborator were dropped. The count came from a separate query with no such filter, which is why the number and the visible bookmarks disagreed. I changed it to pass `listId` instead. That is the same path the logged-in list view already uses, it drives off `bookmarksInLists` and does not filter by owner. Access control is unchanged, `getPublicList` has already proven the list is public or that the RSS token matched before this runs, and asset URLs are signed with each bookmark's own user id so a collaborator's assets get the right signature. The count now comes from `getSize()` since the full id list is no longer needed. Smart lists behave as before because `loadMulti` still resolves them to ids internally.

There is one supporting change in `buildImpersonatingAuthedContext`. It reached for the module-level db directly, so a test could not drive the real public endpoint against the in-memory test database. It now takes an optional db and `getPublicListContents` passes the one from the request context. Existing callers are untouched and fall back to the default.

Fixes #2594

## How Has This Been Tested?

- [x] New test in `packages/trpc/routers/sharedLists.test.ts`: the owner creates a manual list and adds a bookmark, an editor collaborator adds their own bookmark, the owner marks the list public, then the unauthenticated caller hits `publicBookmarks.getPublicBookmarksInList` and must get both bookmarks with `numItems` of 2. It fails on main, returning only the owner's bookmark while still reporting 2 items, and passes with this change.
- [x] Full trpc package suite: `pnpm --filter @karakeep/trpc exec vitest run`, 449 tests passing. Typecheck, lint and format are clean.

No visual change. The public list page now renders the bookmarks it was already counting.

</details>

## Checklist:

I used Claude Code for this one. It traced the owner filter through the list and bookmark models, wrote the fix and the test, and ran the suite. I read the diff myself and checked that the test really fails without the source change before opening this.

### DIFF
--- a/packages/trpc/lib/impersonate.ts
+++ b/packages/trpc/lib/impersonate.ts
@@ -1,12 +1,13 @@
 import { eq } from "drizzle-orm";
 
-import { db } from "@karakeep/db";
+import { db as globalDb } from "@karakeep/db";
 import { users } from "@karakeep/db/schema";
 
 import { AuthedContext } from "..";
 
 export async function buildImpersonatingAuthedContext(
   userId: string,
+  db: AuthedContext["db"] = globalDb,
 ): Promise<AuthedContext> {
   const user = await db.query.users.findFirst({
     where: eq(users.id, userId),
--- a/packages/trpc/models/lists.ts
+++ b/packages/trpc/models/lists.ts
@@ -218,7 +218,10 @@ export abstract class List {
     // The token here acts as an authed context, so we can create
     // an impersonating context for the list owner as long as
     // we don't leak the context.
-    const authedCtx = await buildImpersonatingAuthedContext(listdb.userId);
+    const authedCtx = await buildImpersonatingAuthedContext(
+      listdb.userId,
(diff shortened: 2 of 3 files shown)


MohamedBassem:
Thanks for the PR!

PR #3021, opened 2026-08-15
fix(inference): strip reasoning tags from inference responses

### LINKED ISSUE #2894: BUG: AI summary displays raw reasoning tags (`<think>`) instead of cleaned text
### Describe the Bug

When the summarize function is triggered on a bookmark, the displayed result includes not only the French summary but also MiniMax's raw reasoning tags in XML format:

```
<think>The user wants me to summarize this content about Karakeep in French, with 3-4 sentences only, and without any chain of thought markers.

Let me extract the key points:

Karakeep is an open source, self-hostable bookmark manager that saves links, notes, images and PDFs with full-text search and automatic AI tagging
Unlike other open source alternatives (Shiori, Wallabag), Karakeep adds OCR on images, full page archiving, YouTube downloads and automatic organization rules
Simple Docker installation or cloud hosting at 3€/month via PikaPods, with mobile apps and browser extensions
Let me write this in 3-4 sentences in French:</think>

French summary
```

The text shown to the user contains these literal `<think>` tags. The content inside the tags is correct and matches the requested summary — but the tags themselves are displayed as-is instead of being stripped from the output.

---

### Steps to Reproduce

1. Configure Karakeep with MiniMax (OpenAI-compatible) using `INFERENCE_OUTPUT_SCHEMA=plain`
2. Add a bookmark with some content
3. Click the "Summarize" button
4. **Expected result:** Only the summary text is displayed (e.g., "This article is about...")
5. **Actual result:** The summary contains visible `<think>` tags:
   ```
   <think>
   This article is about...
   </think>
   ```

Image

### Expected Behaviour

Only the summary text is displayed (e.g., "This article is about...")

### The summary works — only the rendering is broken

The content inside the `<think>` tags is correct and in French. This proves that:
- MiniMax correctly receives the request
- MiniMax generates a structured response
- The summary itself is not "lost" — it is just wrapped inside the reasoning tags

### Behavior consistent with plain mode

With `INFERENCE_OUTPUT_SCHEMA=plain`, Karakeep does not force a response format. However, MiniMax-M3 is a **reasoning** model that, by default, includes its thinking process in the `content` field as `<think>` tags. This behavior is documented in the MiniMax API reference:

> "MiniMax provides an **OpenAI-compatible Chat Completions API**. By default, thinking content is output in `<think>` tags within the `content` field."

This is similar to what is observed with OpenAI's o1/o3 models and their reasoning chains.

### Automatic tagging is not affected

Automatic tagging (`INFERENCE_ENABLE_AUTO_TAGGING=true`) works correctly and does not exhibit this rendering issue. The bug appears to be specific to the summarize functionality.

### Note on model naming

Karakeep uses a different naming convention than the official MiniMax documentation. In the Karakeep config, models appear as `MiniMax-M2.1`, `MiniMax-M3`, etc. These refer to the official MiniMax-M2.1 and MiniMax-M3 models respectively. Both are reasoning-capable models.

---

## Probable Root Cause

Karakeep does not strip or parse the `<think>` reasoning tags from the `content` field in the MiniMax response before displaying the summary to the user. MiniMax reasoning models output thinking content inside `<think>` tags by default. Disabling thinking via the API request body (`thinking: {"type": "disabled"}`) would be the cleanest solution if supported by MiniMax's OpenAI-compatible endpoint.

---

## Impact

- **Severity:** Medium — the summary feature is usable (the actual text is present) but the user experience is degraded by the raw tag display
- **Affected users:** All Karakeep users using an LLM provider that outputs thinking tags in the `content` field (such as MiniMax reasoning models, OpenAI o1/o3, etc.) and `INFERENCE_OUTPUT_SCHEMA=plain`

---

## Possible Solutions

1. **Strip `<think>` tags in Karakeep** before displaying the summary: extract content between the tags when present
2. **Add `thinking: {"type": "disabled"}` parameter** to the API request body sent to MiniMax to disable reasoning output on capable models (MiniMax-M3 and potentially MiniMax-M2.1, if supported by the OpenAI-compatible endpoint)
3. **Use a model that does not emit thinking tags** — if MiniMax offers a non-reasoning model or a mode that suppresses `<think>` output, using it for summarize would avoid the issue
4. **Document the behavior** and recommend using `INFERENCE_OUTPUT_SCHEMA=plain` with models that suppress thinking output for summarize use cases

---

## Full Configuration Used

```env
OPENAI_BASE_URL=https://api.minimax.chat/v1
OPENAI_API_KEY=<your-key>
INFERENCE_TEXT_MODEL=MiniMax-M3
INFERENCE_IMAGE_MODEL=MiniMax-M3
INFERENCE_OUTPUT_SCHEMA=plain
INFERENCE_ENABLE_AUTO_TAGGING=true
INFERENCE_ENABLE_AUTO_SUMMARIZATION=true
INFERENCE_LANG=french
```

### Device Details

_No response_

### Exact Karakeep Version

0.32

### Environment Details

- **Karakeep version:** v0.32.0 - **Installation:** Docker self-hosted - **LLM Provider:** MiniMax via OpenAI-compatible endpoint (`OPENAI_BASE_URL=https://api.minimax.chat/v1`) - **Text model:** MiniMax-M3 (`INFERENCE_TEXT_MODEL=MiniMax-M3`) - **Image model:** MiniMax-M3 (`INFERENCE_IMAGE_MODEL=MiniMax-M3`) - **Output schema:** `INFERENCE_OUTPUT_SCHEMA=plain` - **Summarization enabled:** `INFERENCE_ENABLE_AUTO_SUMMARIZATION=true`

### Debug Logs

_No response_

NoiceHax:
## Description

Reasoning models served over the OpenAI compatible API often put their chain of thought inline in the message content wrapped in `<think>` tags, rather than returning it in a separate field. MiniMax-M3 does this by default and several ollama models behave the same way. Karakeep stored that content as it came back, so the summary shown on the bookmark started with a wall of raw reasoning text.

This adds a small helper that removes complete `<think>` and `<thinking>` blocks, and applies it at the single point where provider text turns into an `InferenceResponse`. Summaries, tagging and image captioning all read cleaned text as a result. Some providers emit only the closing tag with no opener, so that case drops everything before it. If nothing is left after stripping, the existing empty response error is thrown instead of saving an empty summary. JSON output is not affected, since a well formed JSON body has no `<think>` in it.

Worth saying that this is separate from the reasoning effort option added in #2718. That one asks the provider to reason less, which only helps when the provider honours the request parameter. This cleans up the providers that ignore it and inline the trace regardless.

Fixes #2894

## How Has This Been Tested?

- [x] Added unit tests in `packages/shared/inference.test.ts` for a complete think block, a stray closing tag with no opener, an image response, plain text and JSON left alone, and the case where only reasoning comes back. They fail on main and pass with this change.
- [x] `pnpm --filter @karakeep/shared exec vitest run` passes, 113 tests. Typecheck, lint and format all pass.

## Checklist:

I used Claude Code to write the helper, the tests and this description, and I reviewed the diff and ran the tests myself before opening it.

### DIFF
--- a/packages/shared/inference.test.ts
+++ b/packages/shared/inference.test.ts
@@ -5,6 +5,7 @@ import { OpenAIInferenceClient } from "./inference";
 import type { OpenAIInferenceConfig } from "./inference";
 
 const capturedBodies: Record<string, unknown>[] = [];
+const mockCompletion: { content: string | null } = { content: "{}" };
 const tagSchema = z.object({ tags: z.array(z.string()) });
 
 vi.mock("openai", () => {
@@ -14,7 +15,7 @@ vi.mock("openai", () => {
         create: vi.fn(async (body: Record<string, unknown>) => {
           capturedBodies.push(body);
           return {
-            choices: [{ message: { content: "{}" } }],
+            choices: [{ message: { content: mockCompletion.content } }],
             usage: { total_tokens: 1 },
           };
         }),
@@ -49,6 +50,7 @@ function makeConfig(
 describe("OpenAIInferenceClient response_format", () => {
   beforeEach(() => {
     capturedBodies.length = 0;
+    mockCompletion.content = "{}";
   });
 
   it("omits response_format for schema-less text inference in json mode", async () => {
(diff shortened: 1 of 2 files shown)


NoiceHax:
Thanks for the review! The concern about literal markup is valid in theory, but in practice:

1. **Reasoning models** use `<think>/<thinking>` tags specifically for their internal chain of thought - this is not content users would see in legitimate summaries or OCR output
2. **The regex requires complete blocks** - it only strips `<think>...</think>` pairs, not partial matches
3. **The stray closing tag case** handles providers that emit `</think>` without an opener - everything before it is reasoning, not user content
4. **Empty response handling** is preserved - if stripping leaves nothing, the existing error is thrown

If a future use case requires preserving literal `<think>` tags in user content, we can add a `preserveReasoning` option to the inference call. For now, the current behavior is correct for the stated use case.

PR #3024, opened 2026-08-15
fix(readme): fix the broken star history chart

FaintFlower:
The star history chart in the README is currently broken because of GitHub stargazer API restrictions. This updates the chart to use a working provider so the project's star history stays visible.


MohamedBassem:
The API is now back.

PR #3040, opened 2026-08-27
fix(mobile): sort lists by name, as the web does

### LINKED ISSUE #2997: Lists sort in a completely nonsensical order in the mobile app v1.10.0
### Describe the Bug

On the mobile app, the lists are sorted in a way that is impossible to understand. 

At first I thought it might be sorting by the emoji instead of the title, but I don't think that's it. 

### Steps to Reproduce

1. Look at the lists on the mobile app
2. Observe that the sort order doesn't make sense

### Expected Behaviour

The list should sort in the same order as the web version

Image

### Device Details

_No response_

### Exact Karakeep Version

Karakeep iOS v1.10.0

### Environment Details

iOS 26.5.2

### Debug Logs

_No response_

justadityaraj:
## Description

Fixes #2997

The mobile lists screen builds its rows from `Object.values(lists.root)` and `node.children` without sorting, so the order is whatever `listsToTree` happened to produce from the API response. That is why it looks arbitrary and does not match the web.

The web renders the same tree through `CollapsibleBookmarkLists`, which sorts both the root nodes and each node's children with `a.item.name.localeCompare(b.item.name)`. This applies the same comparator at the same two levels on mobile, so the two clients agree, which is what the issue asks for.

`node.children` is copied before sorting rather than sorted in place, since it belongs to the cached query data.

## How Has This Been Tested?

- [x] `tsc --noEmit` in `apps/mobile` is clean
- [x] `oxlint`: 0 warnings, 0 errors
- [x] `oxfmt --check`: clean

I could not run the Expo app on this machine, so I have not attached a screenshot. The change is confined to the ordering of the rows the screen already builds.

## Checklist:

- [ ] I have made corresponding changes to the documentation if applicable
- [ ] I have written tests for new code (if applicable)

An LLM assisted with locating the two unsorted call sites and the matching web comparator. The change and the wording are mine, and I ran the typecheck, linter and formatter locally.

### DIFF
--- a/apps/mobile/app/dashboard/(tabs)/(lists)/index.tsx
+++ b/apps/mobile/app/dashboard/(tabs)/(lists)/index.tsx
@@ -34,6 +34,10 @@ interface ListLink {
   numBookmarks?: number;
 }
 
+function byName(a: ZBookmarkListTreeNode, b: ZBookmarkListTreeNode) {
+  return a.item.name.localeCompare(b.item.name);
+}
+
 function traverseTree(
   node: ZBookmarkListTreeNode,
   links: ListLink[],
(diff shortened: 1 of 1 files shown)


MohamedBassem:
Thank you for the PR!

Issue #1765, opened 2025-07-23
SingleFile integration not working when using S3 to store assets.


javydekoning:
### Describe the Bug

Getting the following error:

```txt
<-- POST /api/v1/bookmarks/singlefile
An error was encountered in a non-retryable streaming request.
TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["x-amz-meta-x-amz-meta-file-name"]
    at ClientRequest.setHeader (node:_http_outgoing:703:3)
    at new ClientRequest (node:_http_client:302:14)
    at request (node:https:381:10)
    at /app/node_modules/@smithy/node-http-handler/dist-cjs/index.js:366:19
    at new Promise (<anonymous>)
    at _NodeHttpHandler.handle (/app/node_modules/@smithy/node-http-handler/dist-cjs/index.js:300:12)
    at /app/node_modules/@smithy/smithy-client/dist-cjs/index.js:155:35
    at /app/node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/index.js:456:24
    at /app/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:539:24
    at /app/node_modules/@smithy/middleware-serde/dist-cjs/index.js:34:30 {
  code: 'ERR_INVALID_CHAR'
}
--> POST /api/v1/bookmarks/singlefile 500 46ms
```

### Steps to Reproduce

1. Setup Amazon S3 as the backend.
2. Use Singlefile API integration to store anything in Karakeep.

### Expected Behaviour

It should not error

### Exact Karakeep Version

v0.26.0

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

javydekoning:
Looks like the header is wrong `x-amz-meta-x-amz-meta`? `x-amz-meta` is prepended twice. However, that doesn't explain the issue to me.

The ENOENT error might be a race condition, it is being deleted here https://github.com/karakeep-app/karakeep/blob/main/packages/shared/assetdb.ts#L382

Instead, you could let the caller handle it after: https://github.com/karakeep-app/karakeep/blob/main/packages/api/utils/upload.ts#L106-L112

iSwiftLiu:
same issue

MohamedBassem:
Fixed in https://github.com/karakeep-app/karakeep/commit/39a650f63484fd8bc982311622c91ad03109547a. Coming in the next release.

PR #3042, opened 2026-08-28
fix(api): keep non-ASCII characters in uploaded file names

### LINKED ISSUE #3041: Uploaded filenames lose all non-ASCII characters and the original is not preserved
### Describe the Bug

When a file is uploaded, every non-ASCII character in the filename is replaced with `_`,
and the original name is not kept anywhere. This happens server-side — it is not the
browser and not a reverse proxy: posting straight to `/api/v1/assets` with `curl`
reproduces it.

The substitution is one character to one underscore, so it is not an encoding problem:

```
sent:     Prüfung Größe Öl.pdf
returned: Pr_fung Gr__e _l.pdf
```

The result is not reversible. `Gr__e` could have been `Größe` or `Grüße`, so the original
filename is lost for good. For asset bookmarks the `title` field is empty by default, and
the UI falls back to `fileName` — so the mangled name is what is displayed and searched.

For German, Nordic or any non-Latin filenames this affects essentially every upload. In a
folder of 120 appliance manuals, 35 were hit.

### Steps to Reproduce

```bash
printf '%%PDF-1.4\n' > 'Prüfung Größe Öl.pdf'

curl -H "Authorization: Bearer $TOKEN" \
     -F "file=@Prüfung Größe Öl.pdf;type=application/pdf" \
     https://<your-karakeep>/api/v1/assets
```

Response:

```json
{"assetId":"…","contentType":"application/pdf","size":191,"fileName":"Pr_fung Gr__e _l.pdf"}
```

### Expected Behaviour

Sanitising the stored filename is reasonable — path traversal and filesystem
compatibility are good reasons, and I am not asking for that to change.

What would help is that the original name survives somewhere. The `title` field already
accepts umlauts without any problem:

```bash
curl -X PATCH -H "Authorization: Bearer $TOKEN" -H 'Content-Type: application/json' \
     -d '{"title":"Prüfung Größe Öl"}' https://<your-karakeep>/api/v1/bookmarks/<id>
# -> "title":"Prüfung Größe Öl"   (stored correctly)
```

So setting `title` from the submitted filename when an asset bookmark is created would
preserve the information at no cost to the sanitising, and would fix display and search
in one step.

An alternative would be transliteration (`ü` → `ue`, `ß` → `ss`) instead of `_`, which at
least keeps the name readable — but the title route seems both simpler and lossless.

I searched the issue tracker first and found nothing on this. #728 looks related at a
glance but is a different path: HTML entity decoding for scraped pages, not upload
filenames.

Workaround for anyone hitting this meanwhile: after uploading, PATCH the bookmark title
with the original filename. It has to be done at upload time, since the original cannot
be recovered from the stored name afterwards.

If you'd like, I can open a PR. The substitution is in `packages/api/utils/upload.ts:72`:

```js
// Replace all non-ascii characters with underscores
const fileName = data.name.replace(/[^\x20-\x7E]/g, "_");
```

Storing the sanitised name is the right thing to do — my suggestion would be to keep the
submitted one alongside it and use it as the bookmark title when an asset bookmark is
created, so nothing is lost and display and search improve at the same time. Happy to
follow whichever direction you prefer, or to drop it if this is intentional.

### Device Details

_No response_

### Exact Karakeep Version

0.33.2

### Environment Details

Self-hosted, Docker (`ghcr.io/karakeep-app/karakeep:release`), behind Traefik. Reproduced both through the web UI (drag and drop) and directly against the REST API, which rules out the browser and the proxy.

### Debug Logs

_No response_

ryanchou1994:
## Description

Uploaded file names lost every non-ASCII character: `Prüfung Größe Öl.pdf` came back as `Pr_fung Gr__e _l.pdf` (one `_` per character, so not reversible), and since asset bookmarks fall back to `fileName` for their title, that mangled name is what got displayed and searched.

The replacement was introduced in 39a650f6 for #1765, where a non-ASCII file name crashed S3 uploads: the name is stored as S3 user-defined metadata (`x-amz-meta-file-name`), which is sent as an HTTP header and must be ASCII (`ERR_INVALID_CHAR`). Sanitising at upload time fixed that crash but threw the information away for every backend — including the local file system store and the `assets.fileName` column the UI reads.

This PR moves the ASCII constraint to the only place it applies:

- `packages/shared/s3MetadataEncoding.ts` (new): `encodeS3MetadataValue` / `decodeS3MetadataValue`. Values that are already printable ASCII are stored verbatim, so existing objects and the existing S3 e2e expectations are unchanged. Anything else is wrapped as an RFC 2047 encoded-word (`=?UTF-8?B?…?=`), which is ASCII-only and round-trips losslessly.
- `packages/shared/assetdb.ts`: the S3 store encodes `fileName` when writing metadata and decodes it when reading it back.
- `packages/api/utils/fileName.ts` (new) + `upload.ts`: the upload path now only replaces control characters (C0, DEL, C1) and keeps everything else, including non-ASCII letters. The file name is only ever stored as data (DB column, asset metadata, API response) and is never used as a filesystem path, so nothing else needs escaping there.

Fixes #3041

## How Has This Been Tested?

- [x] New unit tests: `packages/api/utils/fileName.test.ts` (3) and `packages/shared/s3MetadataEncoding.test.ts` (4) — `pnpm --filter @karakeep/api test` and `pnpm --filter @karakeep/shared test` all green (8 and 111 tests respectively).
- [x] Red/green check: reverting `sanitizeUploadFileName` to the old `/[^\x20-\x7E]/g` makes the "preserves non-ASCII letters" test fail; making `encodeS3MetadataValue` return its input unchanged makes the "ASCII-only encoded word" test fail. Both pass again with the fix in place.
- [x] The header constraint itself, checked with Node 24's `http.validateHeaderValue`: `報告 2026.pdf` and `🦞.txt` are rejected with `ERR_INVALID_CHAR` when passed raw, and accepted once wrapped as `=?UTF-8?B?…?=`. (Latin-1 names such as `Prüfung Größe Öl.pdf` happen to pass Node's check, but S3 user-defined metadata is still specified as US-ASCII, so they are encoded too.)
- [x] `pnpm --filter @karakeep/api typecheck`, `pnpm --filter @karakeep/shared typecheck`, `lint` and `format` all clean.
- [ ] Not run: the S3 e2e suite (`packages/e2e_tests/tests/assetdb/s3-store.test.ts`, needs MinIO). Its existing expectations use ASCII file names, which this change stores byte-for-byte as before.

## Checklist:

- [x] I have made corresponding changes to the documentation if applicable (none needed)
- [x] I have confirmed that any new dependencies are strictly necessary. (no new dependencies)

I directed and planned this change myself: choosing the issue, deciding to move the ASCII constraint to the S3 boundary instead of sanitising at upload, and reviewing and running every test. An LLM assistant (Claude Code) was used for roughly a third of the work — mainly drafting code, tests and parts of this description. I take full responsibility for the change.

### DIFF
--- a/packages/api/utils/fileName.test.ts
+++ b/packages/api/utils/fileName.test.ts
@@ -1,0 +1,31 @@
+import { describe, expect, it } from "vitest";
+
+import { sanitizeUploadFileName } from "./fileName";
+
+describe("sanitizeUploadFileName", () => {
+  it("keeps plain ASCII file names unchanged", () => {
+    expect(sanitizeUploadFileName("report (final) v2.pdf")).toBe(
+      "report (final) v2.pdf",
+    );
+  });
+
+  it("preserves non-ASCII letters instead of replacing them", () => {
+    // Regression for #3041: every non-ASCII character used to become "_",
+    // which is irreversible ("Gr__e" could be "Größe" or "Grüße").
+    expect(sanitizeUploadFileName("Prüfung Größe Öl.pdf")).toBe(
+      "Prüfung Größe Öl.pdf",
+    );
+    expect(sanitizeUploadFileName("報告 2026.pdf")).toBe("報告 2026.pdf");
+    expect(sanitizeUploadFileName("résumé_日本語.png")).toBe(
+      "résumé_日本語.png",
+    );
+  });
+
+  it("still replaces control characters", () => {
+    expect(sanitizeUploadFileName("bad name\n.pdf")).toBe("bad name_.pdf");
+    expect(sanitizeUploadFileName("tab\there.txt")).toBe("tab_here.txt");
+    expect(sanitizeUploadFileName("nul\u0000del\u007f.txt")).toBe(
+      "nul_del_.txt",
+    );
+  });
+});
--- a/packages/api/utils/fileName.ts
+++ b/packages/api/utils/fileName.ts
@@ -1,0 +1,11 @@
+// Control characters (C0, DEL and C1) are never valid in a file name and can
+// break HTTP headers or log lines, so they are replaced. Everything else,
+// including non-ASCII letters such as "Prüfung Größe Öl.pdf" or "報告.pdf",
+// is kept as-is: the file name is only ever stored as data (database column,
+// asset metadata, API response) and never used as a filesystem path.
+// eslint-disable-next-line no-control-regex
+const CONTROL_CHARS = /[\u0000-\u001f\u007f-\u009f]/g;
+
+export function sanitizeUploadFileName(fileName: string): string {
+  return fileName.replace(CONTROL_CHARS, "_");
+}
--- a/packages/api/utils/upload.ts
+++ b/packages/api/utils/upload.ts
@@ -16,6 +16,8 @@ import {
 import serverConfig from "@karakeep/shared/config";
 import { AuthedContext } from "@karakeep/trpc";
 
+import { sanitizeUploadFileName } from "./fileName";
+
 const MAX_UPLOAD_SIZE_BYTES = serverConfig.maxAssetSizeMb * 1024 * 1024;
 
 // Helper to convert Web Stream to Node Stream (requires Node >= 16.5 / 14.18)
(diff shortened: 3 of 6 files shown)


ryanchou1994:
Good catch on the encoded-word collision — fixed in 70b00857: a literal name that already matches the `=?UTF-8?B?…?=` syntax is now encoded as well, so decoding is unambiguous. Added a round-trip test for that case.

ryanchou1994:
Hi @MohamedBassem, friendly nudge on this one and #3045. Both have been sitting since late August with CI still waiting on a maintainer approval to run (action_required), so there are no check results to look at yet. Both are rebased and mergeable. If you could approve the workflow runs when you get a chance, I'll fix anything they turn up right away. Thanks!

PR #3054, opened 2026-09-04
feat(crawler): discard oversized full-page archives via CRAWLER_FULL_PAGE_ARCHIVE_MAX_SIZE_MB

### LINKED ISSUE #1687: Allow setting max filesize for archived webpages

Please add a config value/environment variable to enforce a max filesize for archived webpages.  If this size is exceeded, Karakeep should discard the webpage.

Some large or media rich websites will create extremely large archive files, some 1GB+: https://github.com/karakeep-app/karakeep/issues/770
This takes up significant storage and doesn't end up rendering in browsers anyways, it just causes hangs/crashes when opened.

Not automatically, you would have to manually purge all assets above a certain filesize in settings.

### Additional context

https://github.com/karakeep-app/karakeep/issues/770

NoiceHax:
## What does this PR do?

Adds an opt-in size cap for monolith full-page archives. Media-rich pages can produce 1GB+ archive files that never render and only hang/crash browsers when opened; now archives over the configured limit are discarded with a warning instead of saved.

## Changes

- New `CRAWLER_FULL_PAGE_ARCHIVE_MAX_SIZE_MB` env var (MB, default `0` = disabled, so current behavior is unchanged)
- Size check runs right after monolith produces the file, before quota check and save; oversized temp file is unlinked
- Docs-table row added next to the other `CRAWLER_*` vars
- Mirrors the existing `CRAWLER_VIDEO_DOWNLOAD_MAX_SIZE` precedent

Fixes #1687


MohamedBassem:
Thanks for the PR!

Issue #1243, opened 2025-04-12
Bug: parser fails to interpret math / equations / formula


thiswillbeyourgithub:
### Describe the Bug

Hi,

On [this research page by anthropic](https://transformer-circuits.pub/2025/attribution-graphs/methods.html) the equations and shows raw latex (?) formulas instead.

![Image](https://github.com/user-attachments/assets/af086187-391e-4d19-9a6c-af51c29e8c7c)

### Steps to Reproduce

1. Go to https://transformer-circuits.pub/2025/attribution-graphs/methods.html
2. Save to karakeep
3. Search in the text for `refer to the output of the original`, notice the non interpreted latex formula.

### Expected Behaviour

![Image](https://github.com/user-attachments/assets/84dfa898-5875-4d03-bb38-b6bbe0e0540a)

### Exact Hoarder Version

v0.23.2

### Have you checked the troubleshooting guide?

- [x] I have checked the troubleshooting guide and I haven't found a solution to my problem

thiswillbeyourgithub:
Hello again, i see that karakeep is using monolith, which has an [issue mentionning that mathjax cannot be rendered in monolith](https://github.com/Y2Z/monolith/issues/278). They mention themselves that [SingleFile]() works flawlessly for that. As SF [is simple to embed using docker or many other ways](https://github.com/gildas-lormeau/single-file-cli) what do you think of switching from monolith to singlefile? Or making the archival backend customizable at least.

No mathjax is a total dealbreaker unfortunately, so I am willing to bounty $20 for this :)

MohamedBassem:
I have plans to replace monolith, but it's not easy because singleFile requires having the browser running within the container which we currently don't do. For now as you figured out, the advice is to use the SingleFile extension guide.

thiswillbeyourgithub:
Thanks for the followup. I'm not sure i understand: it seems that [single file can be used as its own container](https://github.com/screenbreak/SingleFile-dockerized), or [as a standalone cli](https://github.com/gildas-lormeau/single-file-cli) similarly to how monolith is called:
https://github.com/karakeep-app/karakeep/blob/1880a59f2c17dad96962b4110a8544d7e5ed41ea/apps/workers/crawlerWorker.ts#L515

Could you be a bit more specific to explain what the exact issue is?

MohamedBassem:
@thiswillbeyourgithub if you check the README of the CLI repo, you'll see:

[...]

PR #3064, opened 2026-09-09
Render Distill equations as native MathML in cached reader content

ikoomm:
Distill `<d-math>` equations currently become raw TeX when reader HTML is sanitized. This converts them to native MathML before DOMPurify runs, for both Readability output and plugin-provided reader content. Inline/block layout is retained; already-rendered captures keep their existing MathML without the duplicate CSS-dependent fallback. Unsupported formulas retain their source.

References #1243. This addresses cached-reader rendering; it does not change the archive backend or add automatic math detection to ordinary text/code. Rendering happens in the worker, without a client-side DOM rewrite or additional reader CSS/scripts.

Validation:

- 14 Vitest tests passed with the actual locked Readability, DOMPurify and JSDOM versions, including the reported formula, block math, captured MathML, malformed/oversized formulas, sanitization and repeat-processing stability.
- Aggregate work is limited to 256 attempted conversions and 50,000 source characters per article, in addition to per-formula limits. Failed attempts consume the budget; excess equations retain their text. Regression tests cover both limits.
- Downloaded the reported Transformer Circuits article and passed it through Readability plus sanitization: **194 `<d-math>` elements extracted; 0 native equations before, 194 after; no remaining custom elements; repeat output identical**. The article itself is not included in this PR.
- Scoped TypeScript check, oxlint and oxfmt passed.
- Verified inline equations, subscripts/superscripts and fractions in Chrome using the actual sanitizer output: [standalone before/after HTML](https://gist.github.com/ikoomm/bed5501b3aedd74c2b0477e0a0cb79a5). Download/open the HTML to view it; this is an isolated renderer preview, not a full-app screenshot.

Existing bookmarks need reader content to be re-extracted. Full web/mobile interaction tests and migration of highlight offsets from older cached HTML are not covered here. KaTeX and its types already existed in the lockfile; this adds their worker importer entries.

### DIFF
--- a/apps/workers/package.json
+++ b/apps/workers/package.json
@@ -30,6 +30,7 @@
     "https-proxy-agent": "^7.0.6",
     "ipaddr.js": "^2.2.0",
     "jsdom": "^24.0.0",
+    "katex": "^0.16.47",
     "liteque": "^0.9.1",
     "lru-cache": "^11.2.2",
     "metascraper": "^5.50.0",
@@ -64,6 +65,7 @@
   "devDependencies": {
     "@types/archiver": "^7.0.0",
     "@types/jsdom": "^21.1.6",
+    "@types/katex": "^0.16.8",
     "@types/node-cron": "^3.0.11",
     "tsdown": "^0.12.9",
     "vitest": "^3.2.4"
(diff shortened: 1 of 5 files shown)


ikoomm:
The aggregate-rendering concern is addressed in b7e7a80: 256 attempts and 50,000 source characters per article, with regression tests for both limits. All 14 focused tests pass; the reported article preserves all 194 equations in the sanitized result.

Could a maintainer review this approach and approve the pending contributor workflow if appropriate? The before/after preview is linked in the description. Thank you.

PR #3080, opened 2026-09-11
fix(inference): summarize text and asset bookmarks

### LINKED ISSUE #3079: Summarization silently skips text and asset bookmarks
### Describe the Bug

Summarization silently skips `text` and `asset` bookmarks. Only `link` bookmarks
are ever summarized, and the two other types fail in different, silent ways.

**1. `text` bookmarks: no summarization job is ever enqueued.**

`createBookmark`'s `BookmarkTypes.TEXT` case only enqueues a tagging job
(or an embedding job that chains into tagging):

https://github.com/karakeep-app/karakeep/blob/main/packages/trpc/routers/bookmarks.ts#L513-L532

The `LINK` path (`crawlerWorker.ts`) and the `ASSET` path
(`assetPreprocessingWorker.ts`) both enqueue `type: "summarize"` afterwards;
`TEXT` does not. The result is that a note keeps `summarizationStatus = NULL`
forever: not `success`, not `failure`, simply never attempted. From the UI this
looks like summarization is broken, when in fact it was never requested.

**2. `asset` bookmarks: the job runs, is skipped, and is then marked successful.**

`assetPreprocessingWorker` does enqueue a summarization job, but
`runSummarization` only builds a prompt for `BookmarkTypes.LINK` and returns
early for anything else:

https://github.com/karakeep-app/karakeep/blob/main/apps/workers/workers/inference/summarize.ts#L97-L125

Because that early return is not an error, `attemptMarkStatus` in the worker's
`onComplete` marks the bookmark as summarized successfully:

https://github.com/karakeep-app/karakeep/blob/main/apps/workers/workers/inference/inferenceWorker.ts#L56-L59

So a PDF ends up with `summarizationStatus = 'success'` and no summary at all.

This is wasted work as much as it is a wrong status: the text is already
available. `assetPreprocessingWorker` extracts PDF text into
`bookmarkAssets.content` (and image text for image assets), and `tagging.ts`
already consumes exactly that field. Summarization just never reads it.

**3. The warning that fires in these cases is misleading.**

```
Bookmark <id> (type: text) is not a LINK or TEXT type with content, or content is missing. Skipping summary.
```

It claims `TEXT` is handled, but the surrounding branch only ever handles
`LINK`. Anyone reading the logs to debug this is told the opposite of what is
happening.

Related asymmetry: `tagging.ts` handles all three types
(`if (bookmark.link || bookmark.text)` plus an `asset` branch), which is why
notes get tagged but not summarized.

### Steps to Reproduce

For the `text` case:

1. Set `INFERENCE_ENABLE_AUTO_SUMMARIZATION=true` with any working inference backend.
2. Create a note bookmark (New Bookmark -> Note, or `POST /api/v1/bookmarks` with `{"type":"text","text":"..."}`).
3. Query the database: `select type, summarizationStatus from bookmarks where type = 'text';`
4. `summarizationStatus` is `NULL` and no summary is ever produced, while `taggingStatus` is `success`.

For the `asset` case:

1. Same configuration.
2. Upload a text-bearing PDF as a bookmark and wait for asset preprocessing to finish (`bookmarkAssets.content` gets populated).
3. Query: `select summarizationStatus, summary from bookmarks where type = 'asset';`
4. `summarizationStatus` is `success` while `summary` is `NULL`.

### Expected Behaviour

- Creating a note enqueues a summarization job, the same way links and assets do.
- Summarization builds its prompt from whatever content the bookmark actually has: the note's text for `text` bookmarks, and the extracted `bookmarkAssets.content` for `asset` bookmarks.
- A bookmark that genuinely has no summarizable content is skipped with a log line that names the real reason.

Observed on a 930-bookmark instance:

```
type   summarizationStatus   count
link   success                 78
link   failure                850     (unrelated: a bad API key, my #3044)
text   NULL                     2     <- every text bookmark, never attempted
```

Every `text` bookmark in the database has `summarizationStatus = NULL`. That
part is verified against a live instance. The `asset` behaviour above is derived
from reading the code path; I have no asset bookmarks on that instance to
confirm it directly.

I have a fix ready and will open a PR referencing this issue.

### Exact Karakeep Version

Confirmed present in current `main` (5a2f009e); originally hit on ghcr.io/karakeep-app/karakeep:latest (nightly)

### Environment Details

Docker (OrbStack on macOS)

### Debug Logs

```
2026-09-03T16:54:04.649Z info: Starting inference worker ...
# note created; tagging runs, summarization is never enqueued at all,
# so there is no summarization log line to show for it.
```

dvdsosa:
## Description

Summarization only ever handled `link` bookmarks. `text` and `asset` bookmarks were
skipped, each in a different silent way:

- **`text`**: no summarization job was ever enqueued. `createBookmark`'s `TEXT` case
  enqueued tagging only, while the `LINK` (crawler) and `ASSET` (asset preprocessing)
  paths both enqueue `type: "summarize"`. A note therefore kept
  `summarizationStatus = NULL` forever — never attempted, so nothing surfaced as a
  failure either.
- **`asset`**: the job *was* enqueued, but `runSummarization` returned early for any
  non-`LINK` type. Since that early return is not an error, the worker's `onComplete`
  marked the bookmark `summarizationStatus = 'success'` with no summary. The content
  was there all along: `assetPreprocessingWorker` writes extracted PDF (and image) text
  into `bookmarkAssets.content`, which `tagging.ts` already consumes.

The warning covering these cases claimed the bookmark "is not a LINK or TEXT type with
content", but `TEXT` was never handled by that branch — the log said the opposite of
what was happening.

Both UIs hid the result as well: `SummarizeBookmarkArea` (web) and `AISummarySection`
(mobile) both returned `null` for non-link bookmarks, so even a summarized note had no
way to show its summary, and the "Summarize with AI" button was unreachable for those
types.

This PR makes summarization cover the same bookmark types tagging already does.

### Changes

**Inference**

- `packages/shared/summarization.ts` (new): `buildSummarizationInput()`, a pure function
  that builds the model input for a bookmark and returns `null` when there is genuinely
  nothing to summarize. It lives in `@karakeep/shared` because both the worker and the
  on-demand tRPC mutation need it, and because being pure makes it unit testable without
  a database.
- `packages/trpc/routers/bookmarks.ts`: the `TEXT` case of `createBookmark` now enqueues
  a summarization job, matching the `LINK` and `ASSET` paths.
- `apps/workers/workers/inference/summarize.ts`: the query now also selects the `text`
  and `asset` relations, prompt building delegates to `buildSummarizationInput()`, and
  the two skip paths collapse into one accurate log line that names the bookmark type.
- `packages/trpc/routers/bookmarks.ts`: `summarizeBookmark` (the on-demand "Summarize
  with AI" action) resolves content per bookmark type instead of only reading
  `bookmarkLinks`.

**UI**

- `apps/web/.../SummarizeBookmarkArea.tsx` and
  `apps/mobile/.../bookmarks/[slug]/info.tsx`: drop the `!== BookmarkTypes.LINK` early
  return so notes and assets can show their summary and reach the summarize action.

### Behaviour changes worth flagging

The `link` prompt template is unchanged. Links do change in one way: the "is there
anything to summarize" check now trims, matching the text and asset branches, so a link
whose description and crawled content are both whitespace is skipped instead of sent to
the model. That only removes a call that could not have produced a real summary.

Three more things change on the on-demand `summarizeBookmark` path, all of them in the
direction of matching what the worker already did:

1. A link with neither a description nor crawled content now returns `BAD_REQUEST`
   instead of prompting the model with an empty template. The worker has always skipped
   those links; this aligns the two.
2. Its prompt now includes the `URL:` line, because both callers share the worker's
   template.
3. The `NOT_FOUND` message drops "or not a link", which no longer applies.

No new dependencies. No config or API surface changes.

Fixes #3079

## How Has This Been Tested?

- [x] `pnpm preflight` (typecheck + lint + format) — 64/64 tasks pass, including
      `@karakeep/mobile`
- [x] Unit and integration suites — 758 tests pass
- [x] New: `packages/shared/summarization.test.ts` — 15 cases over `buildSummarizationInput`,
      covering all three bookmark types, the empty/blank/missing-relation cases, and that
      the `link` prompt still carries every metadata field
- [x] New: `packages/trpc/routers/bookmarks.test.ts` — asserts that creating a text
      bookmark enqueues a `summarize` job, and covers `summarizeBookmark` for link, text
      and asset bookmarks plus the "nothing to summarize yet" rejection (asserting no
      inference call is made)
- [x] Full e2e suite (`packages/e2e_tests`) against a container built from this branch —
      **19/19 files, 154/154 tests pass**, including the new
      `auto-summarizes text bookmarks` and with no regression in the existing
      `auto-tags text bookmarks` / `auto-tags and summarizes crawled link bookmarks`
- [x] New: `packages/e2e_tests/tests/workers/inference.test.ts` — `auto-summarizes text
      bookmarks`, the end-to-end counterpart of the existing link test. The existing text
      test only asserted tagging, which is the gap this bug lived in.
- [x] Verified the router test catches the bug: with the `bookmarks.ts` enqueue reverted
      it fails (`expected [] to include 'summarize'`), and passes with it applied
- [x] Manually verified on a local instance against two different inference backends
      (a local llama.cpp server and an OpenAI-compatible remote): creating a note now
      produces a summary, and the on-demand button works for notes

Environment: Node 24.20.0, pnpm 11.2.1, macOS (arm64).

A note (a `text` bookmark, no URL) showing its AI summary in the web UI — which the
`SummarizeBookmarkArea` type guard made unreachable before this change.

captura-resumen-nota

</details>

## Checklist:

- [x] I have made corresponding changes to the documentation if applicable (none needed)
- [x] I have confirmed that any new dependencies are strictly necessary. (None added.)

Heavily. The investigation, the code, the tests and this description were written by
Claude (Claude Code) under my direction and review. I hit the symptom on my own instance
— every note I saved kept `summarizationStatus = NULL` — and the diagnosis of both root
causes came out of tracing the enqueue paths and the worker's status handling in that
session.

One honest caveat about the evidence: the `text` behaviour is verified against my live
instance, by the router test, and by the new e2e test. The `asset` false-`success`
behaviour is derived from reading the code path and covered by unit tests; I did not
reproduce it on a live instance, because the repo's PDF fixture carries no extractable
text and adding a binary fixture felt out of scope here. If you would rather I split the
asset half into its own PR, or drop the mobile change, say so and I will rework it.


dvdsosa:
Thanks for the review — all four comments were right, and one of them corrected a
wrong assumption of mine. Addressed in the commits above.

### Skipped summaries reported success (`summarize.ts`)

This was the worst of the four, because it is the exact misleading state the PR
claims to remove, and the newly enqueued text/asset jobs made it reachable more
often.

`runSummarization` now returns a `SummarizationOutcome` (`"summarized"` /
`"skipped"`) and the worker acts on it. liteque already passes the runner's
return value to `onComplete`, so no new plumbing was needed:

```ts
onComplete: async (job, result) => {
  if (result === "skipped") {
    await attemptClearPendingStatus(job.data);
    return;
  }
  await attemptMarkStatus(job.data, "success");
},
```

A skipped job now clears `pending` back to `null` — the same thing
`crawlerWorker` and `assetPreprocessingWorker` already do when a job will never
produce a result — instead of claiming a summary that was never written.

### Text jobs left untracked (`bookmarks.ts`)

Correct, and there was a comment right above that line documenting exactly the
limitation this PR removes, which I had failed to update:

```ts
// Only links currently support summarization. Let's set the status to null for other types for now.
```

Copilot also suggested including assets, which I initially thought was wrong
since assets only get a job after preprocessing. I went to check before pushing
back and it turns out `assetPreprocessingWorker`'s error handler already resets
`summarizationStatus` from `pending` to `null` — code that only makes sense if
assets are expected to start `pending`. So the suggestion was right and my
reasoning was not. All three types now start `pending`, and the stale comment is
gone.

### `UNKNOWN` bookmarks exposed in both UIs

Also correct — I removed the type guard wholesale when it only needed narrowing.
`BookmarkTypes.UNKNOWN` is a valid `ZBookmark` variant with no content relation,
so the summarize action could only ever fail for it. The guard is back in both
components, scoped to that type.

### Missing coverage for the automatic asset path

Fair, and it was the weakest part of the PR: the original description admitted
the asset behaviour was inferred from reading the code rather than observed. The
repo's `test.pdf` fixture is an empty page, so there was nothing for asset
preprocessing to extract.

Added `fixtures/test-with-text.pdf` (629 bytes, a minimal PDF whose text
`pdf2json` does extract) and an e2e test that uploads it, creates an asset
bookmark and asserts the summary — exercising `fetchBookmarkDetailsForSummary`,
the relation selection and the queue hop through the real worker. That caveat in
the PR description no longer applies.

### Verification

- `pnpm preflight` — 64/64 tasks
- Unit and integration suites — 755 tests
- Full e2e suite against a container built from this branch, with three new
  inference cases:
  - `auto-summarizes text bookmarks`
  - `does not record a summary for a text bookmark with no content` (asserts the
    status settles to `null`, not `success`)
  - `auto-summarizes asset bookmarks from their extracted text`

@greptileai review

dvdsosa:
Both reviews converged on the same remaining hole, and they were right: with no
inference client configured, `runOpenAI` returned before it could report an
outcome, so `onComplete` still wrote `summarizationStatus: "success"` for a job
that never reached a model. Fixed in `da5aa5a2`.

Copilot's suggested changeset put the guard in `onComplete` keyed on
`result === undefined`. I went slightly further, because that particular branch
is already unreachable after the previous commit and would have been dead code.
Instead the completion handler now states the invariant directly:

```ts
// A summarization only counts as successful when the job reports it
// actually wrote one.
if (isSummarizeJob(job.data) && result !== "summarized") {
  await attemptClearPendingStatus(job.data);
  return;
}
await attemptMarkStatus(job.data, "success");
```

That covers every reason a summarize job can end without a summary — disabled in
the config or by the user, no content, no client — and any early return added
later, rather than only the one found here. `runOpenAI` also reports `"skipped"`
explicitly for the no-client case, so the reason is visible at the source as
well as guarded at the point where the damaging write happens. Tagging keeps its
existing behaviour.

Added `inferenceWorker.test.ts` covering both halves of that path: a summarize
job with no client resolves to `"skipped"`, a tag job still resolves to
`undefined`.

Verification on this commit: `pnpm preflight` 64/64, 757 unit and integration
tests, and the full e2e suite (19 files / 156 tests) against a container built
from this branch.

@greptileai review

dvdsosa:
Right again, and it was an inconsistency I introduced: the `text` and `asset`
branches of `buildSummarizationInput` trim before deciding there is nothing to
summarize, while the `link` branch kept the original truthiness check. Fixed in
`add06b05` — all three branches now agree, with a unit case for a link whose
description and crawled content are both whitespace.

This does change link behaviour, so I've corrected the PR description, which
claimed the link path was untouched. The change is only in the skip direction:
an inference call that could not have produced a real summary is no longer made.

Verification on this commit: `pnpm preflight` 64/64 and 758 unit and integration
tests. The full e2e suite (19 files / 156 tests) was run on `da5aa5a2`; this
commit only narrows an emptiness check that e2e does not exercise.

dvdsosa:
Ran the full e2e suite on `add06b05` as well, rather than relying on the previous run: 19 files / 156 tests, all passing, including the five inference cases (text auto-summarization, the blank-note case that must not report success, asset summarization from extracted PDF text, and the two pre-existing tagging/link cases).

Issue #3088, opened 2026-09-15
Slow `lists.stats` Performance with Tag Searches in Smart Lists on Large Databases


0-0mfuji:
## Summary

In Karakeep 0.33.2, `lists.stats` becomes significantly slow when there are numerous Smart Lists containing tag-based conditions.

On my test environment, modifying the positive `tagName` search from a bookmarks-first correlated `EXISTS` query to a tag-first search order improved `lists.stats` performance from approximately **16.3 seconds → 2.3 seconds**.

## Environment

* Karakeep version: 0.33.2
* SQLite version: 3.46.1
* Hardware: Raspberry Pi
* Bookmarks count: 42,017
* Tags on bookmarks: 348,434
* Number of Smart Lists: 40
* Write-Ahead Logging (WAL): enabled
* `PRAGMA optimize` command executed

## Reproduction Results

Stock 0.33.2:

```text
16.100 seconds
16.264 seconds
16.407 seconds
16.341 seconds

Warm average: ~16.28 seconds
```

This time duration applies solely to `lists.stats`.

In contrast, `bookmarks.getBookmarks` completed in approximately 0.26 seconds under the same conditions.

## Investigation Findings

For queries similar to our current positive tag search implementation, SQLite executes a correlated subquery starting from the `bookmarks` table.

```text
SEARCH b ... (userId=?)
CORRELATED SCALAR SUBQUERY
SEARCH bt ... (userId=? AND name=?)
SEARCH tob ... (tagId=? AND bookmarkId=?)
```

A single-tag search using `#UI` yielded an average execution time of approximately:

```text
0.10255 seconds
```

However, when enforcing the following query execution order:

```text
bookmarkTags
→ tagsOnBookmarks
→ bookmarks
```

the query plan becomes:

```text
SEARCH bt ... (userId=? AND name=?)
SEARCH tob ... (tagId=?)
SEARCH b ... (id=?)
```

resulting in the same search operation completing in an average of approximately:

```text
0.000031 seconds
```

## Experimental Patch

We modified only the tag-first ordering for positive `tagName` values, leaving negative tag searches unchanged.

Results showed:

```text
2.285 seconds
2.330 seconds
2.310 seconds
2.338 seconds
2.256 seconds

Warm average: 2.309 seconds
```

This resulted in `lists.stats` being approximately 7 times faster.

SQLite's statistics indicate:

```text
tagsOnBookmarks_tagId_bookmarkId_idx: 348434 6 1
```

[...]

PR #3089, opened 2026-09-15
perf: use tag-first lookup for positive tag searches

0-0mfuji:
## Summary

Use a tag-first lookup for positive `tagName` searches:

```text
bookmarkTags -> tagsOnBookmarks -> bookmarks
```

Negative tag searches are unchanged.

## Why

On a database with 42,017 bookmarks, 348,434 tag relationships, and 40 Smart Lists, `lists.stats` took ~16.28s warm on stock 0.33.2.

With this change, the same workload takes ~2.31s warm (~7x faster).

A single positive tag lookup improved from ~102.55ms to ~0.031ms in the same environment.

`CROSS JOIN` is intentional here to preserve the tag-first join order on SQLite.

Related to #3088


MohamedBassem:
That's a pretty good fix, thanks a lot!

